PR #26, opened 2017-05-20
[RFC] add `Resource.read` method for atomic reads of a resource's value

japaric:
this method lets you directly read the value of a resource without paying for
synchronization overhead (no need for `Threshold.raise`). The catch is that
resource value must have a size of 4-bytes or smaller to guarantee that the read
operation is atomic -- atomic as in it can't be interrupted.

---

### Unresolved questions

- Is this *really* memory safe?

It's easy to show that this is *not* memory safe for resources with a size in
memory greater than 4 bytes as the read operation would have to be split in two
LDR instructions and an interrupt may happen in between corrupting the value
being read. For example:

``` rust
#[derive(Clone, Copy)]
#[repr(u64)]
enum Silly {
    A = 0x0000_0000_FFFF_FFFF,
    B = 0xFFFF_FFFF_0000_0000,
}

static R1: Resource<Cell<Silly>> = Resource::new(Cell::new(Silly::A));

fn t1(..) {
    let r1 = R1.read();
    // LDR first word
    // ! interrupt occurs and R1 changes to variant B
    // LDR second word
    // you end with r1 = 0xFFFF_FFFF_FFFF_FFFF which matches no variant
    // thus you got yourself Undefined Behavior
}
```

ISTM that this `read` operation ought to be safe with 32-bit values as the
operation is performed in a single LDR operation. But can the processor
execution pipeline make `read` memory unsafe somehow? IDK.

What about reading a 8-bit or 16-bit value? Is the operation also atomic in
those cases?

- Should we check for alignment?

Are non-aligned memory loads non-atomic?

- Do we need to use `read_volatile`?

I think so. Consider this code:

``` rust
static R1: Resource<Cell<u32>, Cn> = Resource::new(Cell::new(0));

fn t1(..) {
    let a = R1.read();

    // ! interrupt occurs

    let b = R1.read();
}
```

`a` may not be the same as `b` because an interrupt in between may change the
value of `R1`. Using volatile ensures that the compiler won't merge the read
operations, which could change semantics.

- Should we worry about making it easier to do non-atomic updates?

Consider:

``` rust
static R1: Resource<Cell<i32>, C2> = Resource::new(Cell::new(0));

fn t1(_: Exti0Irq, ref prio: P1, ref thr: T1) {
    // Atomic update
    thr.raise(
        &R1, |thr| {
            let r1 = R1.access(prio, thr);
            r1.set(r1.get() + 1);
        }
    );

    // Non-atomic update
    let curr = thr.raise(&R1, |thr| R1.access(prio, thr).get());
    thr.raise(
        &R1, |thr| {
            let r1 = R1.access(prio, thr);
            r1.set(curr + 1);
        }
    );

    // Non-atomic update (now easier)
    let curr = R1.read();
    thr.raise(
        &R1, |thr| {
            let r1 = R1.access(prio, thr);
            r1.set(curr + 1);
        }
    );
}
```

- Should we also add `write`?

32-bit writes are also atomic; they map to a single STR instruction. So we could
add a `write` counterpart to allow atomic writes. My worry is that with `write`
non-atomic updates become even easier:

``` rust
static R1: Resource<Cell<i32>, C2> = Resource::new(Cell::new(0));

fn t1(_: Exti0Irq, ref prio: P1, ref thr: T1) {
    // Non-atomic update (even easier)
    let curr = R1.read();
    R1.write(curr + 1);
}
```

**NB** I intended `read` to work with `Resource<Cell<T>, Cn>` and with
`Resource<T, Cn> where T: Copy` but I got a "conflicting implementations" error
from the compiler. So I chose the `Cell` variant as that's what most people will
use in practice.

Thoughts? @perlindgren @whitequark

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -426,9 +426,9 @@ extern crate cortex_m;
 extern crate static_ref;
 extern crate typenum;
 
-use core::cell::UnsafeCell;
+use core::cell::{Cell, UnsafeCell};
 use core::marker::PhantomData;
-use core::ptr;
+use core::{mem, ptr};
 
 use cortex_m::ctxt::Context;
 use cortex_m::interrupt::Nr;
@@ -533,6 +533,29 @@ impl<T, RC> Resource<T, RC> {
         PT: GreaterThanOrEqual<RC>,
     {
         unsafe { Ref::new(&*self.data.get()) }
+    }
+}
+
+impl<T, RC> Resource<Cell<T>, RC> {
+    /// Performs an atomic read of the resource value
+    ///
+    /// This method has no synchronization overhead, and doesn't impose
+    /// additional task blocking
+    ///
+    /// # Panics
+    ///
+    /// - This method will panic if the size of `T` is greater than the size
+    ///   of a word (4 bytes)
+    pub fn read(&self) -> T {
+        assert!(mem::size_of::<T>() <= mem::size_of::<usize>());
+
+        // NOTE(volatile) this has to be volatile because unlike a read through
+        // `access` there's no guarantee that the resource value will *not*
+        // change between two consecutive reads (e.g. an interrupt may occur
+        // in between)
+        unsafe {
+            ptr::read_volatile(self.data.get() as *const _)
+        }
     }
 }
 


whitequark:
> It's easy to show that this is not memory safe for resources with a size in
> memory greater than 4 bytes as the read operation would have to be split in two
> LDR instructions

Not if you use LL/SC. You can load at most an entire cache line using LL/SC although for values much larger than 8 bytes this clearly loses against Threshold.raise.

japaric:
@whitequark this implementation isn't using LDREX/STREX but a vanilla LDR instruction. Also it seems that [LDREX/STREX is not interrupt safe](https://github.com/ARMmbed/mbed-drivers/issues/48) so it doesn't seem a good candidate for extending this to types bigger than 4 bytes.

I discussed this with @perlindgren and concluded that this kind of breaks the Resource abstraction because it lets any higher priority task read *intermediate* values of a resource that's being modified by a task that has locked the resource. Although, that's *not* memory unsafe by itself using intermediate values can lead to footguns (e.g. with enums representing states of a state machine) so it requires you to be careful with this `read` method.

We'd rather not add an API that requires you to be careful with it so we have decided to not add this.

PR #16, opened 2017-04-26
hide the unsafe Peripheral::new constructor

japaric:
`peripherals!` is the safe way to declare `Peripheral`s

PR #15, opened 2017-04-25
add a `peripherals` macro

japaric:

Issue #1, opened 2017-03-12
Cortex-M0(+) support


japaric:
There's no BASEPRI on this architecture.

We can start with just supporting cooperative tasks and then support `lock` by manually masking the other tasks using NVIC.

japaric:
Fixed in #34

PR #34, opened 2017-07-04
app! macro take 2

### LINKED ISSUE #13: Interoperation with cortex_m::peripheral::Peripheral.borrow is unsound
### Background

The `cortex-m` crate defines a `Peripheral` type which is a new type around the address block address of the peripheral. We use that `Peripheral` in our `Peripheral` constructor. svd2rust generates a bunch of `cortex_m::Peripheral` "instances" (as `pub const`s) which the user later uses to create a `rtfm::Peripheral` `static`.

### Problem

The problem is that `cortex_m::Peripheral` provides a `borrow` method that grants access to the peripheral within a `cortex_m::interrupt::free` critical section but that `borrow` doesn't know about the SRP or task priorities. Problematic code looks like this:

``` rust
use rtfm::{C1, P1, P2, Peripheral}
use stm32f30x::interrupt;

// RESOURCES
static GPIOA: Peripheral<stm32f30x::Gpioa, C1> = unsafe { Peripheral::new(stm32f30x::GPIOA) };

// TASKS
tasks!(stm32f30x, {
    t1: Task {
        interrupt: Exti0,
        priority: P1,
        enabled: true,
    },
    t2: Task {
        interrupt: Exti0,
        priority: P2,
        enabled: true,
    },
});

fn t1(_task: interrupt::Gpioa, prio: P1) {
    let ceil = prio.as_ceiling();

    // NOTE GPIOA has type `cortex_m_rtfm::Peripheral`
    let gpioa = GPIOA.access(&prio, ceil);

    // Immediately preempts
    rtfm::request(t2);
}

fn t2(_task: interrupt::Gpiob, _prio: P2) {
    cortex_m::interrupt::free(|cs| {
        // this GPIOA has type `cortex_m::peripheral::Peripheral`
        // this breaks SRP because GPIOA has ceiling C1 which is less than this task priority (P2)
        let gpioa = stm32f30x::GPIOA.borrow(cs);
    });
}
```

### Possible solutions

- Remove `cortex_m::interrupt::free` or have svd2rust generated code not create `cortex_m::peripheral::Peripheral` instances. Instead svd2rust should generate some other type for peripheral instances but that doesn't have a `borrow` method that works with `cortex_m::interrupt::free`

- Make `cortex_m::Peripheral::borrow` `unsafe` if and only if the `cortex-m-rtfm` crate is used with the `cortex-m` crate. I don't know if this is possible at all: Perhaps add a "rtfm" Cargo feature to `cortex-m` that makes `cortex_m::Peripheral::borrow` `unsafe` and have `cortex-m-rtfm` depend on `cortex-m` with that feature enabled. Because Cargo features are additive if one `cortex-m` instance in the dependency graph has the "rtfm" feature enabled then the `cortex-m` crate will always be compiled with that feature enabled. If this does work, I think it would probably be considered a misuse of Cargo features...

japaric:
### Update

More ports from RTFM v1 to this macro:

- [Two wheeled robot](https://github.com/japaric/2wd/pull/4)
- [WS2812B LED driver](https://github.com/japaric/ws2812b/pull/1)
- [Several examples from the blue-pill crate](https://github.com/japaric/blue-pill/pull/19)

---

[Full example](https://gist.github.com/japaric/e825cabe5a95d8e75e08917a4bf4d195)

^ This is semantically the same example presented in #31, but with this new
macro 33% less code is required.

Next iteration of the rtfm! macro presented in #31. The most notable change wrt
the previous version is that the macro has become a ~~compiler plugin~~ procedural macro.

# Pros

- Automatic ceiling derivation. The ceiling will be computed from the
  information about which tasks use which resources.

- Simpler task and resource specification. There's no split between interrupts
  used as tasks and exceptions used as tasks; there's only tasks now. Likewise
  there's no distinction between peripherals used as resources and data
  resources; there's only resources now.

- More flexible syntax. The fields can be shuffled around without breaking the
  parser.

- Simpler signatures as there's no `Peripherals` / `Resources` split.

- Adaptive signatures. The signature of idle will be `fn() -> !` if it makes no
  uses of resources and has no local data, but it will change to
  `fn(idle::Local)` if local data is used, etc.

- Less claims. Full access to resources is given to tasks whose priority matches
  the resource ceiling.

- More control over error messages related to the parsing and expansion of
  `rtfm!`.

# Cons

`rtfm!` is now a compiler plugin. This makes the chance of it breaking across
nightly updates greater than zero. However, I've implemented the compiler plugin
on top of token streams so we should be safe from compiler changes related to
the language syntax and parser. Once function style macros becomes available via
the more stable proc-macros feature it should be straightforward to move to it.
And yes, I tried [proc-macro-hack] which allows function style proc macros on
top of custom derive proc macros but it doesn't work for no-std / cross
compilation because the hack requires a proc-macro crate compiled for the target
(which is not possible for thumb because proc-macro contains `String`s).

[proc-macro-hack]: https://github.com/dtolnay/proc-macro-hack

# Other changes

- I realized that the change to scope based safety fixes #13 so I've added a
  `borrow` / `borrow_mut` method to resources to get direct access to them
  within a global critical section.

- Given that we pretty much generate the implementation of the claim methods in
  the application crate I have changed things such that (NVIC) PRIORITY_BITS is
  no longer in the cortex-m-rtfm, but rather it's expected to be provided by the
  device crate. This way we don't need a bunch of P{2,3,..} Cargo features in
  the cortex-m-rtfm crate. OTOH, device generated crates now must provide a
  `NVIC_PRIO_BITS` constant that provides this information. SVD files contain
  this information in the form of the optional [`<nvicPrioBits>`] field.
  svd2rust will be taught to generate the `NVIC_PRIO_BITS` constant.

[`<nvicPrioBits>`]: https://www.keil.com/pack/doc/CMSIS/SVD/html/elem_cpu.html

- The declaration of task local data and the task function path has been moved
  into a `task!` macro. This macro doesn't have to be in the root of the crate.

cc @whitequark


macros/src/syntax/parse.rs
```diff
@@ -0,0 +1,496 @@
...
+        let mut pieces = vec![];
+        loop {
+            if let Some(tt) = tts.next() {
+                if tt == TokenTree::Token(Token::Eq) {
+                    break;
+                } else {
+                    pieces.push(tt);
+                }
+            } else {
+                panic!("expected type, found EOM");
```
whitequark: Nit: maybe expand `EOM` to `end of macro`? It took me a few moments to recognize.

macros/src/trans.rs
```diff
@@ -0,0 +1,486 @@
...
+                    interrupts.push(quote! {
+                        let nvic = #device::NVIC.borrow(cs);
+                    });
+                }
+
+                let priority = task.priority;
+                interrupts.push(quote! {
+                    let prio_bits = #device::NVIC_PRIO_BITS;
+                    let hw = ((1 << prio_bits) - #priority) << (8 - prio_bits);
+                    nvic.set_priority(#device::Interrupt::#name, hw);
```
whitequark: Nit: could this be extracted as another function on `NVIC`? Seems useful.

whitequark:
Outstanding work @japaric! You've addressed nearly every concern of mine, except:
  * Regarding explicit specification of resources for the `init` function, I still think it's somewhat pointless, but I concede that it's not worth the added implementation complexity. One could argue that there's the benefit of seeing an overview of every resource initialized.
  * The naming of `Local`. It's not really important.

macros/src/syntax/parse.rs
japaric: will do

macros/src/trans.rs
japaric: you mean adding a `set_logical_priority` to NVIC or adding a `logical2hw` function to map a logical priority to a "hardware" value that can be directly written to an NVIC register

japaric:
@whitequark Thanks

> Regarding explicit specification of resources for the init function, I still think it's somewhat pointless, but I concede that it's not worth the added implementation complexity.

I was thinking about this and it would be easy to change the implementation such that init has access to any all the resources declared in other tasks. The issue I see with this approach is that init is likely going to require more resources, specially peripherals (e.g. RCC), than what the other tasks use so you would still need to list more resources in `init.resources`.

The problem with giving init access to all the peripherals is that there's no such list that the `rtfm!` macro can use. At least not right now. I was thinking of adding a `Peripherals` struct to device crates which contains a reference to each and every peripheral / register block and make its constructor unsafe. With that change the init signature could be changed to `fn($device::Peripherals, Resources)` to make it have access to all the device peripheral and all the application resources. I'll give this a try.

macros/src/trans.rs
whitequark: former

japaric:
the last commit removes init.resources from rtfm! and makes idle.local optional. The change depends on japaric/svd2rust#123

Diff of the example shown below:

``` diff
--- a/0-example.rs	2017-07-06 17:54:13.908406632 -0500
+++ b/0-example.rs	2017-07-06 17:55:23.428268225 -0500
@@ -42,25 +41,23 @@

     init: {
         path: init,
-        resources: [AFIO, GPIOA, GPIOC, TIM2, RCC, USART1],
     },

     idle: {
-        local: {},
         path: idle,
         resources: [DWT, SLEEP_CYCLES],
     },
 }

-fn init(r: init::Resources) {
-    let timer = Timer(r.TIM2);
-    let serial = Serial(r.USART1);
+fn init(p: init::Peripherals, _: init::Resources) {
+    let timer = Timer(p.TIM2);
+    let serial = Serial(p.USART1);

-    led::init(r.GPIOC, r.RCC);
+    led::init(p.GPIOC, p.RCC);

-    serial.init(BAUD_RATE.invert(), r.AFIO, None, r.GPIOA, r.RCC);
+    serial.init(BAUD_RATE.invert(), p.AFIO, None, p.GPIOA, p.RCC);

-    timer.init(FREQUENCY.invert(), r.RCC);
+    timer.init(FREQUENCY.invert(), p.RCC);
     timer.resume();
 }

@@ -122,4 +119,4 @@

     let byte = serial.read().unwrap();
     serial.write(byte).unwrap();
-}
\ No newline at end of file
+}
```

japaric:
I have [ported] two non trivial [applications] to the latest rtfm! version from cortex-m-rtfm v0.1.1. I'm quite happy with the result. The Cells and RefCells are gone; the task code is shorter, mainly because no `access` calls are needed with the new rtfm!; I even saw a drop in CPU usage of 0.5% (15% -> 14.5%) in my most demanding app just from eliminating the RefCells. (I also saw a fixed drop in binary size of about 480 bytes but I think this is mainly due to changes in cortex-m-rt)

[ported]: https://github.com/japaric/2wd/commits/v2
[applications]: https://github.com/japaric/ws2812b/commits/v2

whitequark:
nice!

japaric:
I have now ported all the examples in the blue-pill crate to this macro (japaric/blue-pill#19). The result in a reduction in LoC of about 30% in average.

In another news this has already broke due to a change in rustc:

```
error[E0425]: cannot find function `set_parse_sess` in module `proc_macro::__internal`
```

whitequark:
@japaric Why are you using proc macro internals?

japaric:
@whitequark I thought that functionlike! proc macros weren't implemented given the existence of proc-macro-hack and the lack of documentation about macros 2.0 so I was using a compiler plugin. But I figured out how to use `#![proc_macro]` so I'll switch to that.

japaric:
Some changes:

- the `rtfm!` macro has been renamed to `app!`
- the macro parser has been refactored into its [own crate](https://github.com/japaric/rtfm-syntax). The main motivation for this is to share code between this crate and the port of the tasks and resources model to [the MSP430 architecture](https://github.com/japaric/msp430-rtfm). Yup, you can now use this very same macro to write concurrent applications on MSP430 :tada:.
- I have made the fields `init.path` and `idle.path` optional. If omitted these fields default to the paths `init` and `idle` respectively, which is what most people will use, I expect.
- I have replaced all the panic! / unwraps in the proc macro implementation with proper error handling (error-chain :heart:). You now get nice backtraces when there's an error in the contens of the `app!` macro. For example:

``` rust
app! {
    device: blue_pill::stm32f103xx,

    tasks: {
        SYS_TICK: {
            priority: 1,
            resources: { SYST }, // <- wrong delimiter
        },
    },
}
```

```
error: proc macro panicked
  --> examples/blinky.rs:16:1
   |
16 | / app! {
17 | |     device: blue_pill::stm32f103xx,
18 | |
19 | |     tasks: {
...  |
24 | |     },
25 | | }
   | |_^
   |
   = help: message: Error: parsing
           Caused by: parsing `tasks`
           Caused by: parsing task `SYS_TICK`
           Caused by: parsing `resources`
           Caused by: expected Bracket, found Brace
```

I believe this is good as the error messages can get without proper span information.

whitequark:
This is amazing work.

japaric:
I tested the effect of removing the Threshold token; that is instead of tracking the value of the preemption threshold in the source the threshold value is tracked at runtime by reading the BASEPRI register. You can see the required changes in the [no-threshold](https://github.com/japaric/cortex-m-rtfm/tree/no-threshold) branch. The result was that the `claim` closure ends up being triplicated in the output machine code.

For example, this:

``` rust
fn idle(mut r: idle::Resources) -> ! {
    r.STATE.claim_mut(|_| {
        asm::nop();
        asm::nop();
        asm::nop();
        asm::nop();
    });

    loop {}
}
```

produces this:

``` armasm
08000186 <main>:
(..)
 80001ae:       f3ef 8010       mrs     r0, PRIMASK
 80001b2:       f010 0f01       tst.w   r0, #1
 80001b6:       d104            bne.n   80001c2 <main+0x3c>
 80001b8:       bf00            nop
 80001ba:       bf00            nop
 80001bc:       bf00            nop
 80001be:       bf00            nop
 80001c0:       e014            b.n     80001ec <main+0x66>
 80001c2:       f3ef 8011       mrs     r0, BASEPRI
 80001c6:       f1c0 0110       rsb     r1, r0, #16
 80001ca:       b2c9            uxtb    r1, r1
 80001cc:       290f            cmp     r1, #15
 80001ce:       d809            bhi.n   80001e4 <main+0x5e>
 80001d0:       21f0            movs    r1, #240        ; 0xf0
 80001d2:       f381 8811       msr     BASEPRI, r1
 80001d6:       bf00            nop
 80001d8:       bf00            nop
 80001da:       bf00            nop
 80001dc:       bf00            nop
 80001de:       f380 8811       msr     BASEPRI, r0
 80001e2:       e003            b.n     80001ec <main+0x66>
 80001e4:       bf00            nop
 80001e6:       bf00            nop
 80001e8:       bf00            nop
 80001ea:       bf00            nop
 80001ec:       e7fe            b.n     80001ec <main+0x66>
```

which seems like a non starter.

For comparison this is what the `Threshold` (this PR) version generates:

``` armasm
08000178 <main>:
(..)
 80001a0:       21f0            movs    r1, #240        ; 0xf0
 80001a2:       f3ef 8011       mrs     r0, BASEPRI
 80001a6:       f381 8812       msr     BASEPRI_MAX, r1
 80001aa:       bf00            nop
 80001ac:       bf00            nop
 80001ae:       bf00            nop
 80001b0:       bf00            nop
 80001b2:       f380 8811       msr     BASEPRI, r0
 80001b6:       e7fe            b.n     80001b6 <main+0x3e>
```

I think we can revisit this after v0.2.0. Maybe rewriting claim in assembly can reduce the bloat. The closure will *always* be executed, and what changes is the "guards" (interrupt::disable or basepri::write or nothing) around the closure execution. I think the "memory" clobber in interrupt::disable / basepri::write may be deoptimizing the code too much.

EDIT: Compiling with opt-level = z reduces the binary size by turning the closure into a separate routine. But the closure needs to be larger than some size for this optimization to occur:

``` armasm
080001a4 <main>:
(..)
 80001c0:       b510            push    {r4, lr}
 80001c2:       f3ef 8010       mrs     r0, PRIMASK
 80001c6:       f010 0f01       tst.w   r0, #1
 80001ca:       d102            bne.n   80001d2 <main+0x2e>
 80001cc:       f7ff ffd3       bl      8000176 <app::idle::{{closure}}>
 80001d0:       e012            b.n     80001f8 <main+0x54>
 80001d2:       f3ef 8411       mrs     r4, BASEPRI
 80001d6:       f1c4 0010       rsb     r0, r4, #16
 80001da:       b2c0            uxtb    r0, r0
 80001dc:       280f            cmp     r0, #15
 80001de:       d807            bhi.n   80001f0 <main+0x4c>
 80001e0:       20f0            movs    r0, #240        ; 0xf0
 80001e2:       f380 8811       msr     BASEPRI, r0
 80001e6:       f7ff ffc6       bl      8000176 <app::idle::{{closure}}>
 80001ea:       f384 8811       msr     BASEPRI, r4
 80001ee:       e003            b.n     80001f8 <main+0x54>
 80001f0:       f7ff ffc1       bl      8000176 <app::idle::{{closure}}>
 80001f4:       e8bd 4010       ldmia.w sp!, {r4, lr}
 80001f8:       e7fe            b.n     80001f8 <main+0x54>
 80001fa:       bf00            nop
 80001fc:       e000e406        .word   0xe000e406
 8000200:       e000e100        .word   0xe000e100

08000176 <app::idle::{{closure}}>:
 8000176:       bf00            nop
 8000178:       bf00            nop
 800017a:       bf00            nop
 800017c:       bf00            nop
 800017e:       bf00            nop
 8000180:       bf00            nop
 8000182:       bf00            nop
 8000184:       bf00            nop
 8000186:       bf00            nop
 8000188:       bf00            nop
 800018a:       bf00            nop
 800018c:       bf00            nop
 800018e:       bf00            nop
 8000190:       bf00            nop
 8000192:       4770            bx      lr
 ```

japaric:
more changes: documentation, examples and a `Resource` trait which you can use to write generic code that can operate on a set of resources regardless of the task where it's called from (see example) and that optimizes very well.

japaric:
@homunkulus try

homunkulus:
:hourglass: Trying commit dd539d1 with merge dd539d1...

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/cortex-m-rtfm/builds/258286490?utm_source=github_status&utm_medium=notification)

japaric:
@homunkulus try

homunkulus:
:hourglass: Trying commit 7d0c07c with merge 7d0c07c...

homunkulus:
:broken_heart: Test failed - [status-travis](https://travis-ci.org/japaric/cortex-m-rtfm/builds/258289992?utm_source=github_status&utm_medium=notification)

japaric:
@homunkulus try

japaric:
@homunkulus r+

homunkulus:
:pushpin: Commit 3d2be6d has been approved by `japaric`

japaric:
@homunkulus r+

homunkulus:
:pushpin: Commit 697c2d6 has been approved by `japaric`

japaric:
@homunkulus r+

homunkulus:
:pushpin: Commit 2d80f36 has been approved by `japaric`

homunkulus:
:hourglass: Testing commit 2d80f36 with merge 2d80f36...

homunkulus:
:sunny: Test successful - [status-travis](https://travis-ci.org/japaric/cortex-m-rtfm/builds/258803918?utm_source=github_status&utm_medium=notification)
Approved by: japaric
Pushing 2d80f3631bc3bf382ae24c25dbf8ff33af2ad430 to master...

PR #44, opened 2017-09-06
[WIP] travis: disable cache

jonas-schievink:
Travis is currently working with an old yanked version of the
stm32f103xx crate (0.7.2). This is bad.

CI will probably fail on this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,10 +51,6 @@ script:
 
 after_script: set +e
 
-cache: cargo
-before_cache:
-  - chmod -R a+r $HOME/.cargo
-
 branches:
   only:
     - auto
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -80,7 +80,6 @@
 extern crate cortex_m;
 extern crate cortex_m_rtfm_macros;
 extern crate rtfm_core;
-extern crate static_ref;
 
 use core::u8;
 


jonas-schievink:
Alternatively it might be possible to run `cargo update` as the first step

jonas-schievink:
```
error: <inline asm>:2:17: error: unknown directive
                .thumb_func
                ^


error: <inline asm>:4:21: error: invalid instruction mnemonic 'b'
                    b DEFAULT_HANDLER
                    ^


error: aborting due to 2 previous errors

error: Could not compile `stm32f103xx`.
```

Same error I'm getting locally (which is preventing me from running the cfail tests I need for https://github.com/japaric/cortex-m-rtfm/pull/43), caused by svd2rust generating ARM inline assembly. What now?

japaric:
> Travis is currently working with an old yanked version of the
stm32f103xx crate (0.7.2).

I think I can clear the Travis cache once to fix issues like instead of outright disabling caching.

> Same error I'm getting locally (which is preventing me from running the cfail tests I need for #43), caused by svd2rust generating ARM inline assembly.

This is caused by the `global_asm!` block and only when the "rt" features is enabled. I have a svd2rust patch in the works to fix this.

japaric:
> I have a svd2rust patch in the works to fix this.

See japaric/svd2rust#140

jonas-schievink:
> I think I can clear the Travis cache once to fix issues like instead of outright disabling caching.

Yeah, this PR was only a test. I'm worried about this happening again in the future, though. Running `cargo update` as part of CI would prevent that, but it would only ever test against the most up-to-date versions of all dependencies. But maybe that's a good thing?

japaric:
> Running cargo update as part of CI would prevent that

Yes but, at worst, it would also invalidate the build cache on every build / PR.

> but it would only ever test against the most up-to-date versions of all dependencies.

I think that if an old version of a dependency is causing problems then we should bump the version requirement in Cargo.toml instead of just testing a new version in Travis. Ideally, we should also test against the oldest versions of the dependencies since we are promising, in the Cargo.toml, that they work but I don't think there's a command to test such configuration. At least, using the cache will exercise old versions.

I'm going to close this PR since it's not really applicable as it is.

Issue #53, opened 2017-11-24
work around Cortex-M7 BASEPRI erratum


japaric:
See erratum 837070 in "ARM Processor Cortex-M7 (AT610) and Cortex-M7 with FPU (AT611), Product revision r0, Sofware Developers Errata Notice".

Basically the problem is that `msr BASEPRI` won't immediately take effect and the instruction following it won't have the new / desired dynamic priority and thus it can be preempted as if BASEPRI hadn't been modified.

The suggested workaround is to temporarily disable interrupts when `msr BASEPRI` is issued, so basically something like this:

``` armasm
cpsid i
msr BASEPRI
cpsie i
```

We can directly apply this fix in claim since claim only modifies BASEPRI when interrupts are enabled. (If interrupts are disabled, claim is a no-operation) 

However, It would be ideal if we only applied this fix when targeting Cortex-M7 devices. And I think we can do that but we'll need some help from svd2rust. SVD files contains these `<cpu>` and `<revision>` fields that would let us identify chips with the problem. Device crates could expose `CPU` and `REVISION` constants that claim could check to decide whether to apply the workaround or not.

Thanks @pftbest for reporting this.
cc @perlindgren

perlindgren:
I haven't looked closely into this, but the gut feeling is that it should work with a NOP following the BASEPRI related instruction. The assembly instructions should be executed in "program order" (not being allowed for the compiler to re-order).

PR #60, opened 2017-12-23
work around Cortex-M7 BASEPRI erratum

japaric:
This PR works around the issue in RTFM and depends on the SVD file properly declaring the CPU name.

The other solution to the problem is to patch the issue in the cortex-m crate: i.e. to patch the
`register::basepri::write` method. In that context there's no CPU information so we'll have to apply
the fix to the `thumbv7m-none-eabi`, `thumbv7em-none-eabi` and `thumbv7em-none-eabihf` targets --
which would negatively affect several applications.

It should be noted that this PR would leave `cortex_m::basepri::write` broken for Cortex-M7 targets.

cc @perlindgren
see #53 for background information
depends on japaric/svd2rust#163


japaric:
Closing in favor of japaric/cortex-m#72 and #61

PR #78, opened 2018-05-24
Unused resource is not declared

bergus:
Declare statics for all resources not just those with an ownership (that are used in a task or idle).
Used resources without definition are already found in `tasks()`.

I (more or less accidentally) had a resource, late-initialised by `init`, that was not used in any tasks. The compiler was rightfully complaining that `main` was trying to assign to a variable that didn't exist in scope.

### DIFF
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -11,10 +11,10 @@ version = "0.3.1"
 
 [dependencies]
 failure = "0.1.1"
-proc-macro2 = "0.3.6"
-quote = "0.5.1"
-rtfm-syntax = "0.3.0"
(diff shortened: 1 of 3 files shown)


japaric:
A lot has change since this PR and the v0.4.0 release. If I'm reading this correctly the issue was declaring a late resource but not assigning it to a task. That works correctly (it compiles) with v0.4.0:

``` rust
#[app(device = lm3s6965)]
const APP: () = {
    static mut FOO: u32 = ();

    #[init]
    fn init() {
        FOO = 0;
    }
};
``` 

If you encounter any issue with the latest version please open a new issue.

PR #79, opened 2018-05-30
Modified Resource to not always require Send

thequux:
It is possible that some tasks will need a resource that is
initialized at runtime, but is not send (for example, the motivating
example is a smoltcp interface). This patch makes it so that a
resource that is only ever used in one task does not need to be Send.

Compatibility issues:
 * Anything that depends on `<T as Resource>::Data : Send` should
   instead assert that `T: SharedResource`

### DIFF
--- a/macros/src/trans.rs
+++ b/macros/src/trans.rs
@@ -4,7 +4,7 @@ use proc_macro2::TokenStream;
 use syn::Ident;
 use syntax::check::App;
 
-use analyze::Context;
+use analyze::{Context, Ceiling};
 
 pub fn app(ctxt: &Context, app: &App) -> TokenStream {
     let mut root = vec![];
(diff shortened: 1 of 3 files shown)


japaric:
Thanks for the PR but I believe this change is not sound. We have a compile-fail test against [such scenario].

[such scenario]: https://github.com/japaric/cortex-m-rtfm/blob/v0.4.0-beta.1/tests/cfail/late-not-send.rs

`init` has an effective static priority of 0. When doing runtime initialization you could be creating a value at that priority and then moving the value into a task with higher priority. That operation requires the value to be `Send` because it's a transfer between execution contexts.

An example of where this goes wrong is [`unsend::Pool`], which is a fixed-size memory pool whose allocations (`unsend::Box`) are automatically deallocated when `drop`-ed. If we didn't enforce a `Send` bound then it would be possible to allocate `unsend::Box`es in `init` and send them to different tasks, that could run at different priorities.

`unsend::Pool` is safe because it, nor its allocations, implement the `Send` trait. This means that the pool and the boxes allocated on it are constrained to the same execution context (or the same priority level). If that were not the case then it would be possible to run `Box.drop` at different priority levels and that would lead to data races.

[`unsend::Pool`]: https://docs.rs/alloc-singleton/0.1.0/alloc_singleton/nightly/pool/unsend/struct.Pool.html

However, it could be the case our use of `Send` in RTFM is not correct and that we should be using a different marker trait. If you can open an issue with details about your use case I can take a closer look.

PR #87, opened 2018-09-01
Update cortex-m, cortex-m-rt, stm32f103xx

ykomatsu:
`cortex-m-rtfm 0.3.4` is not compatible with latest `stm32f103xx-hal` (>= japaric/stm32f103xx-hal@5487385a8d05c5b8ea28e59b18a627d1acaefacf)

Usage:
1. Add ~~`cortex-m-rt-macros` and~~ `panic-` *something* crate to `[dependencies]` table in your `Cargo.toml`
2. Add `#![no_main]` attribute to your code
3. Add ~~`extern crate cortex-m-rt-macros;` and~~ `extern crate panic_` *something* `;` to your code

### DIFF
--- a/macros/src/trans.rs
+++ b/macros/src/trans.rs
@@ -11,7 +11,7 @@ fn krate() -> Ident {
 pub fn app(app: &App, ownerships: &Ownerships) -> TokenStream {
     let krate = krate();
     let mut root = vec![];
-    let mut main = vec![quote!(#![allow(path_statements)])];
+    let mut main = vec![];
 
     ::trans::tasks(app, ownerships, &mut root, &mut main);
     ::trans::init(app, &mut main, &mut root);
@@ -22,12 +22,8 @@ pub fn app(app: &App, ownerships: &Ownerships) -> TokenStream {
         use #krate::entry;
 
         #[entry]
+        #[allow(path_statements, unsafe_code)]
         fn main() -> ! {
-            main_impl();
-        }
-
-        #[allow(unsafe_code)]
-        fn main_impl() -> ! {
             #(#main)*
         }
     });


examples/zero-tasks.rs
```diff
@@ -2,8 +2,11 @@
 #![deny(unsafe_code)]
 #![deny(warnings)]
 #![no_std]
+#![no_main]
 
+extern crate cortex_m_rt_macros;
```
TeXitoi: That would be better to depend on cortex_m_rt as you already need it on your Cargo.toml to use the rt feature.

TeXitoi: I think the better solution would be to reexport the needed macro directly in cortex-m-rtfm to not depend on extern crate by the user.

ykomatsu: I think you're right.

Could you check 0054ef7c387c911c168f103ba59ae3d4751a2201, please?

TeXitoi: Great! https://github.com/TeXitoi/rusty-clock/commit/1f10a38bc173f3661f6545afba71741dfd2b62c6

macros/src/trans.rs
```diff
@@ -18,8 +19,15 @@ pub fn app(app: &App, ownerships: &Ownerships) -> TokenStream {
...
+        use #krate::entry;
+
+        #[entry]
+        fn main() -> ! {
+            main_impl();
+        }
+
         #[allow(unsafe_code)]
-        fn main() {
+        fn main_impl() -> ! {
```
TeXitoi: Can't you just add `#[entry]` on main without creating `main_impl`?

ykomatsu: Well, let's see, is this all right? 493f71943a0d983d636f89135a88dd141b5e80d5

ReeceStevens:
I would like to chime in for this PR; the current cortex-m-rtfm crate is actually not binding exception handlers at the moment due to the `SYS_TICK` -> `SysTick` change and it has broken my application. Happy to jump in if there's any extra help needed here-- @TeXitoi or @ykomatsu , do either of you know the state of this?

ykomatsu:
@ReeceStevens Thank you for checking my branch!

@japaric If there is no problem with this pull request, could you merge it, please?

ReeceStevens:
That is great news! Can’t wait to try it out. Thanks for the hard work getting this framework working with the beta tool chain :)

PR #96, opened 2018-11-02
adapt to changes in cortex-m-rt

burjui:


burjui:
Tested changes on STM32FDISCOVERY board with SysTick interrupt. I had to add #![no_main] to my crate, since the `app' procedural macro doesn't allow inner attributes

TeXitoi:
#87 doesn't fix your issue?

burjui:
I had no idea #87 existed. It does work, though.

PR #132, opened 2019-01-16
Make builds reproducible

hugwijst:
This is done by using `BTreeMap`s and `BTreeSet`s to get deterministic ordering.

Also updated the CI job to check reproducibility of all examples.

### DIFF
--- a/macros/src/codegen.rs
+++ b/macros/src/codegen.rs
@@ -3,8 +3,6 @@
 use proc_macro::TokenStream;
 use std::{
     collections::{BTreeMap, HashMap},
-    sync::atomic::{AtomicUsize, Ordering},
-    time::{SystemTime, UNIX_EPOCH},
 };
 
 use proc_macro2::Span;
@@ -44,6 +42,8 @@ struct Context {
     tasks: BTreeMap<Ident, Task>,
     // Alias (`struct` / `static mut`)
     timer_queue: Ident,
+    // Generator of Ident names or suffixes
+    ident_gen: IdentGenerator,
 }
 
 struct Dispatcher {
@@ -63,19 +63,22 @@ struct Task {
 
 impl Default for Context {
     fn default() -> Self {
+        let mut ident_gen = IdentGenerator::new();
+
         Context {
             #[cfg(feature = "timer-queue")]
-            baseline: mk_ident(None),
+            baseline: ident_gen.mk_ident(None),
             dispatchers: BTreeMap::new(),
-            idle: mk_ident(Some("idle")),
-            init: mk_ident(Some("init")),
-            priority: mk_ident(None),
+            idle: ident_gen.mk_ident(Some("idle")),
+            init: ident_gen.mk_ident(Some("init")),
+            priority: ident_gen.mk_ident(None),
             statics: Aliases::new(),
             resources: HashMap::new(),
-            schedule_enum: mk_ident(None),
+            schedule_enum: ident_gen.mk_ident(None),
             schedule_fn: Aliases::new(),
             tasks: BTreeMap::new(),
-            timer_queue: mk_ident(None),
+            timer_queue: ident_gen.mk_ident(None),
+            ident_gen,
         }
     }
 }
@@ -140,13 +143,13 @@ pub fn app(app: &App, analysis: &Analysis) -> TokenStream {
         () => quote!(),
     };
 
-    let timer_queue = timer_queue(&ctxt, app, analysis);
+    let timer_queue = timer_queue(&mut ctxt, app, analysis);
 
     let pre_init = pre_init(&ctxt, &app, analysis);
 
     let assertions = assertions(app, analysis);
 
-    let main = mk_ident(None);
+    let main = ctxt.ident_gen.mk_ident(None);
     let init = &ctxt.init;
     quote!(
         #resources
@@ -213,7 +216,7 @@ fn resources(ctxt: &mut Context, app: &App, analysis: &Analysis) -> proc_macro2:
                 pub static #mut_ #name: #ty = #expr;
             ));
 
-            let alias = mk_ident(None);
+            let alias = ctxt.ident_gen.mk_ident(None);
             if let Some(Ownership::Shared { ceiling }) = analysis.ownerships.get(name) {
                 items.push(mk_resource(
                     ctxt,
@@ -229,7 +232,7 @@ fn resources(ctxt: &mut Context, app: &App, analysis: &Analysis) -> proc_macro2:
 
             ctxt.statics.insert(name.clone(), alias);
         } else {
-            let alias = mk_ident(None);
+            let alias = ctxt.ident_gen.mk_ident(None);
             let symbol = format!("{}::{}", name, alias);
 
             items.push(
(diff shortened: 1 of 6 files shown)


hugwijst:
The last bit of build jitter seems to be caused by the randomly generated suffixes. Is there any parallelism in the code? Otherwise only using the number of times `mk_ident` is called should result in deterministic suffixes.

hugwijst:
> The last bit of build jitter seems to be caused by the randomly generated suffixes. Is there any parallelism in the code? Otherwise only using the number of times `mk_ident` is called should result in deterministic suffixes.

Some more digging shows that this would also need to be fixed in https://github.com/japaric/owned-singleton/blob/master/macros/src/lib.rs#L180.

macros/src/codegen.rs
```diff
@@ -1913,52 +1916,45 @@ fn mk_typenum_capacity(capacity: u8, power_of_two: bool) -> proc_macro2::TokenSt
...
+        let n;
+        let mut s = if let Some(name) = name {
+            n = 4;
+            format!("{}_", name)
         } else {
-            s.push(('0' as u8 + rng.gen::<u8>() % 10) as char)
+            let crate_name = env!("CARGO_PKG_NAME").replace("-", "_").to_lowercase();
+            n = 16;
+            format!("{}__internal__", crate_name)
+        };
```
hugwijst: If prepending the package name is good enough namespacing, we could just as well append sequentially increasing digits instead of randomly generating a suffix.

hugwijst:
Any feedback? @japaric?

japaric:
Thanks for the PR, @hugwijst. Unfortunately, this change lets people trigger Undefined Behavior (UB) in safe code and that's a no-go.

Here's how:

- Grab any of the examples and expand it using [`cargo-expand`](https://crates.io/crates/cargo-expand), e.g. `cargo expand --example task`.
- Copy down the identifier of one of the tasks, e.g. `baz_aa7w`
- Now modify the code to call the task (i.e. `baz_aa7w()`) from any other task
- UB in safe code

All tasks are potentially non-reentrant because they can have task local data (the `static mut` declared at the beginning of the task handler) so calling them directly can break Rust aliasing rules and produce UB. That's not the only scenario where that can happen: #135 describes several other scenarios where calling tasks as functions produces UB.

The reason we are currently generating random identifiers for the tasks *on each run* is exactly to prevent this issue. Even if you do `cargo expand` and take note of the task name that identifier won't be valid the next time you build the crate. This makes it *impossible* to invoke tasks as functions.

---

Ideally, I would use *hygienic* identifiers instead of random identifiers. `macro_rules!` (AKA macros 1.0) has hygienic identifiers. For example the code below doesn't compile:

``` rust
macro_rules! create_hygienic_identifier {
    () => {
        let x = 42;
    }
}

fn main() {
    create_hygienic_identifier!();
    println!("The answer is {}", x);
    //                           ~ error: not found in this scope
}
```

``` text
$ # expanded code for reference
$ cargo expand
fn main() {
    let x = 42;
    {
        ::std::io::_print(::std::fmt::Arguments::new_v1(
            &["The answer is ", "\n"],
            &match (&x,) {
                (arg0,) => [::std::fmt::ArgumentV1::new(arg0, ::std::fmt::Display::fmt)],
            },
        ));
    };
}
```

because macros 1.0 are hygienic.

However, procedural macros (AKA macros 1.2), which is what RTFM uses, completely lack hygiene. I don't even know if hygienic identifiers are in their roadmap.

---

There are other ways to get reproducible builds.

One way is to use `export_name` on all the internal `static` and `fn` items. That way we can give them stable symbol names like `task::foo` while keeping the random identifier (so calling `foo()` from source code will still not compile). In fact I was doing this at the beginning but it caused problems with [`cargo-call-stack`](https://github.com/japaric/cargo-call-stack): it seemed that `-Z emit-stack-sizes` (which `cargo-call-stack` re-compiles the crate with) caused unused symbols to *not* be GC-ed by the linker and that resulted in the call graph having a bunch of empty nodes so I removed all the uses of `export_name` in #107.

Another way is to not use random identifiers for the tasks and `static mut`s while also making all tasks expand into `unsafe fn` (unsafe functions). That would make tasks `unsafe` to call, and `unsafe` code triggering UB is acceptable. This alternative I rejected early on while implementing v0.4.0 because I have found that (beginner) Embedded Rust developers tend to show little to no hesitation in using `unsafe` code (which makes sad and worried because `unsafe` is usually hard to get right). I decided it was best to not give them the chance to use `unsafe`.

---

@hugwijst I think the way forward here would be to partially undo #107, i.e. add back the `export_name` attribute but with stable names (e.g. `task::foo`, `resource::BAR`). You may also need the `BTree*` changes in this PR, not sure. And I would prefer if the emission of the `export_name` attribute only happened if a certain Cargo feature is enabled, at least until I figure out if I can fix the issues I saw with `cargo-call-stack`.

Let me know if you need any help with that!

TeXitoi:
Can't you generate the code as private in a private module to avoid the user calling directly the generated methods instead of this random trick? Or declare these functions unsafe?

TeXitoi:
Sorry, I've only read the notification, thus the unsafe solution is not a good suggestion.

hugwijst:
Thanks for the explanation @japaric.

Part of the problem with making the builds reproducible is that I don't fully understand what source code changes result in assembly (layout) changes.

My main assumption has been that the actual values of identifiers don't matter, as long as they are declared in the same order. This is why I changed the `Hash*` data structures to `BTree*`.

Today I tried to generate identifiers with the following format: `{name}_{call_count}_{random_suffix}`, where `call_count` is an incrementing integer. This would result in a consistent ordering of identifiers between runs, while still having an unpredictable part.

Sadly this still results in unreproducible builds: the compiler appears to use the source code (or maybe just the identier names?) as a source of randomness for its hash functions. These hashes are used to give symbols a unique suffix. As some methods seem to be sorted by there symbol name in the final assembly, this can result in different a different method ordering and thus different binaries.

I'll investigate if partly reverting #107 helps. If it does, I'll put it behind a feature gate and update this PR.

hugwijst:
The solution described above does seem to work on release builds. I've pushed a commit to ensure that that is true across all tested platforms.

From the jitter I saw in debug builds the problem is with some methods in the `heapless` crate. Looking at https://github.com/rust-lang/rust/blob/master/src/librustc_codegen_utils/symbol_names.rs, it could be due to type parameters of heapless, as these are taken into account when generating the symbol hash. This doesn't show up in release builds because the methods get inlined.

Using `export_name` doesn't solve this problem, as the hashes appended to the symbol names can not be controlled.

The current PR works for me for release builds, but will fail if/when the compiler decides not to inline one of these functions.

Would a solution be to add a `cfg` that generates deterministic names, potentially prefixing them with `__do_not_call_directly__internal__` or something? That combined with marking the functions `unsafe` might be good enough?

japaric:
> From the jitter I saw in debug builds

I wonder if the problem is that dev builds use multiple codegen units while release builds use a single codegen unit (at least in this repository). I'm not sure if parallel llvm codegen is deterministic as it produces a bunch of separate object files; IDK if the linker always sorts symbols alphabetically.

@hugwijst I would suggest setting the codegen units for the dev profile [to 1](https://github.com/japaric/cortex-m-rtfm/blob/aadd00c06896c49ad5dc672c2e7e15bb10b61833/Cargo.toml#L64) and see if that helps.

These changes look reasonable to me and I'd be happy to merge this PR after a rebase. (Sorry for the merge conflicts but I had to land #140 ahead of this one to fix a soundness issue).

PR #133, opened 2019-01-17
Fix build on recent nightlies.

hugwijst:


japaric:
@hugwijst Thanks!

bors r+

PR #140, opened 2019-02-12
fix soundness issue: forbid early returns in init

japaric:
TL;DR

1. v0.4.1 will be published once this PR lands

2. v0.4.0 will be yanked once v0.4.1 is out

3. v0.4.1 will reject code that contains early returns in `init` *and* contains
   late resources. Yes, this is a breaking change but such code is unsound /
   has undefined behavior.

4. as of v0.4.1 users are encouraged to use `fn init() -> init::LateResources`
   instead of `fn init()` when they make use of late resources.

---

This PR fixes a soundness issue reported by @RalfJung. Basically, early
returning from `init` leaves *late resources* (runtime initialized statics)
uninitialized, and this produces undefined behavior as tasks rely on those
statics being initialized. The example below showcases a program that runs into
this soundness issue.

``` rust
const APP: () = {
    // this is actually `static mut UNINITIALIZED: MaybeUninit<bool> = ..`
    static mut UNINITIALIZED: bool = ();

    #[init]
    fn init() {
        // early return
        return;

        // this is translated into `UNINITIALIZED.set(true)`
        UNINITIALIZED = true; // the DSL forces you to write this at the end
    }

    #[interrupt(resources = [UNINITIALIZED])]
    fn UART0() {
        // `resources.UNINITIALIZED` is basically `UNINITIALIZED.get_mut()`

        if resources.UNINITIALIZED {
            // undefined behavior
        }
    }
};
```

The fix consists of two parts. The first part is producing a compiler error
whenever the `app` procedural macro finds a `return` expression in `init`. This
covers most cases, except for macros (e.g. `ret!()` expands into `return`) which
cannot be instrospected by procedural macros. This fix is technically a
breaking change (though unlikely to affect real code out there) but as per our
SemVer policy (which follows rust-lang/rust's) we are allowed to make breaking
changes to fix soundness bugs.

The second part of the fix consists of extending the `init` syntax to let the
user return the initial values of late resources in a struct. Namely, `fn() ->
init::LateResources` will become a valid signature for `init` (we allowed this
signature back in v0.3.x). Thus the problematic code shown above can be
rewritten as:

``` rust
const APP: () = {
    static mut UNINITIALIZED: bool = ();

    #[init]
    fn init() -> init::LateResources {
        // rejected by the compiler
        // return; //~ ERROR expected `init::LateResources`, found `()`

        // initialize late resources
        init::LateResources {
            UNINITIALIZED: true,
        }
    }

    #[interrupt(resources = [UNINITIALIZED])]
    fn UART0() {
        if resources.UNINITIALIZED {
            // OK
        }
    }
};
```

Attempting to early return without giving the initial values for late resources
will produce a compiler error.

~~Additionally, we'll emit warnings if the `init: fn()` signature is used to
encourage users to switch to the alternative `init: fn() -> init::LateResources`
signature.~~ Turns out we can't do this on stable. Bummer.

The book and examples have been updated to make use of `init::LateResources`.

In the next minor version release we'll reject `fn init()` if late resources
are declared. `fn init() -> init::LateResources` will become the only way to
initialize late resources.

This PR also prepares release v0.4.1. Once that version is published the unsound
version v0.4.0 will be yanked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Use an absolute link to the book so it works when landing from crates.io
   documentation page
 
+- The initialization function can now be written as `fn init() ->
+  init::LateResources` when late resources are used. This is preferred over the
+  old `fn init()` form.
+
 ### Fixed
 
 - `#[interrupt]` and `#[exception]` no longer produce warnings on recent nightlies.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,31 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [v0.4.1] - 2019-02-12
+
+### Added
+
+- The RTFM book has been translated to Russian. You can find the translation
+  online at https://japaric.github.io/cortex-m-rtfm/book/ru/
+
+- `Duration` now implements the `Default` trait.
+
+### Changed
+
+- [breaking-change][] [soundness-fix] `init` can not contain any early return as
+  that would result in late resources not being initialized and thus undefined
+  behavior.
+
 - Use an absolute link to the book so it works when landing from crates.io
   documentation page
 
-## [v0.4.0] - 2018-11-03
+### Fixed
+
+- `#[interrupt]` and `#[exception]` no longer produce warnings on recent nightlies.
+
+## [v0.4.0] - 2018-11-03 - YANKED
+
+Yanked due to a soundness issue in `init`; the issue has been mostly fixed in v0.4.1.
 
 ### Changed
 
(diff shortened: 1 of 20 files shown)


TeXitoi:
I was thinking if this would not be a more simple interface: the user declare a struct with all the resources, and this struct must be returned by the init function. In this case, all the resources must be late:

```rust
const APP: () = {
    #[resources]
    struct Resources {
        uninitialized: bool,
    }

    #[init]
    fn init() -> Resources {
        // initialize resources
        Resources {
            unitinialized: true,
        }
    }

    #[interrupt(resources = [uninitialized])]
    fn UART0() {
        if resources.uninitialized {
            // OK
        }
    }
};
```

japaric:
@TeXitoi that looks reasonable to me and could a candidate for v0.5.0 (\*) but would need an RFC.

(\*) it's possible to add such alternative syntax to v0.4.x but I'd rather not have 3 different ways to do the same thing.

TeXitoi:
Of course for 0.5. I thought of this interface while using 0.3, and this PR remind me this idea.

Do you want I write such an RFC? If yes, I'll try to find the time to at least write a first draft within a week.

japaric:
> Do you want I write such an RFC? 

Yes, please.

TeXitoi:
RFC proposed: https://github.com/japaric/cortex-m-rtfm/issues/143

Shnatsel:
Thank you for yanking the crate!

It's also a good idea to create a security advisory at https://github.com/RustSec/advisory-db so that even people with a committed Cargo.lock could check if they're running an unsound version and upgrade.

PR #151, opened 2019-02-15
make builds reproducible

japaric:
This is a rebased and augmented version of #132. With this PR both dev and release builds that do not use the owned-singleton stuff become reproducible. (I haven't really bothered to make owned-singleton reproducible since [lifo] is way more ergonomic than [alloc-singleton] and will eventually make its way into heapless).

[lifo]: https://github.com/japaric/lifo
[alloc-singleton]: https://crates.io/crates/alloc-singleton

Thanks @hugwijst for doing the bulk of the work!

closes #132

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /book/*/book
 /target
 Cargo.lock
+*.hex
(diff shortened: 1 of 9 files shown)


hugwijst:
This is awesome! Thanks for getting this one over the line @japaric :smile:

PR #152, opened 2019-02-16
added cycle_count to Instant to get the cycle count

### LINKED ISSUE #150: `rtfm::Instant` cannot be formatted with the default formatter
I guess I'm doing something wrong here but I'm out of ideas.
I try to print the value of an Instant without using `{:?}`

The most interesting line is:

> --> src/main.rs:202:53
       |
202 |             let s_count = format!("{},{:>10.2}\r\n",now,*resources.val+5.5);
       |                                                                        ^^^ `rtfm::Instant` cannot be formatted with the default formatter
       |
    = help: the trait `core::fmt::Display` is not implemented for `rtfm::Instant`
    = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
    = note: required by `core::fmt::Display::fmt`

`now` is the rtfm::Instant.

I know I could use `{:?}` but then I would get something like `Instant(1163512974)`, I do not want the Instant and  (). Cutting the string does not seem to be the right solution.

oni303:
fixes #150

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -146,6 +146,10 @@ impl Instant {
         assert!(diff >= 0, "second instant is later than self");
         Duration(diff as u32)
     }
+    /// Returns the cycle count of this Instant
+    pub fn cycle_count(&self) -> i32{
+        self.0
+    } 
 }
 
 #[cfg(feature = "timer-queue")]


japaric:
Thanks for the PR. `Instant` is an opaque type and we don't plan to expose any details about its implementation. See https://github.com/japaric/cortex-m-rtfm/issues/150#issuecomment-464453119

PR #158, opened 2019-02-23
implement RFC #128: #[interrupt(binds = ..)]

### LINKED ISSUE #128: [RFC] #[interrupt(binds = "..")]
## Summary

Extend the syntax to let the user name interrupt and exception tasks differently
from the interrupt / exception they bind to.

## Background

This code doesn't compile today:

``` rust
//~ ERROR: the name `USART1` is defined multiple times

#![no_main]
#![no_std]

extern crate panic_halt;

use stm32f103xx::USART1;

#[rtfm::app(device = stm32f103xx)]
const APP: () = {
    static mut USART: Option<USART1> = None;

    #[init]
    fn init() {}

    #[interrupt]
    fn USART1() {}
};

```

One workaround is to remove the `USART1` import and use `stm32f103xx::USART1`
instead of `USART1` in the definition of the `USART` resource.

The underlying issue is that the `#[interrupt]` attribute will expand into a
module named `USART1`. This module contains types like `crate::USART1::Spawn`
and `crate::USART1::Schedule` (note the `USART1` prefix) and it's required to
differentiate the `Spawn` struct of one task from the others'.

## Design

This RFC proposes extending the syntax of both the `#[exception]` and
`#[interrupt]` attributes to take a `binds` argument.

The value of `binds` must be a valid exception / interrupt name for the target
device. When omitted `binds` defaults to the task (function) name. The module
in the expansion of `#[exception]` / `#[interrupt]` will match the task and not
the `binds` value.

An example to illustrate the changes:

``` rust
// No import collision this time
use stm32f103xx::USART1;

#[rtfm::app(device = stm32f103xx)]
const APP: () = {
    static USART: USART1 = ();

    #[init]
    fn init() {
        // ..
    }

    #[interrupt(binds = USART1, resources = [USART], spawn = [foo])]
    fn serial() {
        bar(spawn);

        // ..
    }

    #[task]
    fn foo() {
        // ..
    }

    extern "C" {
        fn EXTI0();
    }
};

// NOTE: `serial::Spawn` and not `USART1::Spawn`
fn bar(_: serial::Spawn) {}
```

## Unresolved questions

- Bikeshed alternative names for `binds`.

japaric:
closes #128 

r? @korken89 or @TeXitoi 

suggestions for tests are welcome! (2 of the 3 tests I added hit bugs in my implementation)

### DIFF
--- a/macros/src/check.rs
+++ b/macros/src/check.rs
@@ -106,12 +106,8 @@ pub fn app(app: &App) -> parse::Result<()> {
     }
 
     // Check that free interrupts are not being used
-    for (name, interrupt) in &app.interrupts {
-        let name = if let Some(ref binds) = interrupt.args.binds {
-            binds
-        } else {
-            name
-        };
+    for (handler, interrupt) in &app.interrupts {
+        let name = interrupt.args.binds(handler);
 
         if app.free_interrupts.contains_key(name) {
             return Err(parse::Error::new(
--- a/macros/src/check.rs
+++ b/macros/src/check.rs
@@ -106,10 +106,16 @@ pub fn app(app: &App) -> parse::Result<()> {
     }
 
     // Check that free interrupts are not being used
-    for int in app.interrupts.keys() {
-        if app.free_interrupts.contains_key(int) {
+    for (name, interrupt) in &app.interrupts {
+        let name = if let Some(ref binds) = interrupt.args.binds {
+            binds
+        } else {
+            name
+        };
+
+        if app.free_interrupts.contains_key(name) {
             return Err(parse::Error::new(
-                int.span(),
+                name.span(),
                 "free interrupts (`extern { .. }`) can't be used as interrupt handlers",
             ));
         }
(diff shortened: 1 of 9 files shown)


macros/src/check.rs
```diff
@@ -106,10 +106,16 @@ pub fn app(app: &App) -> parse::Result<()> {
...
 
     // Check that free interrupts are not being used
-    for int in app.interrupts.keys() {
-        if app.free_interrupts.contains_key(int) {
+    for (name, interrupt) in &app.interrupts {
+        let name = if let Some(ref binds) = interrupt.args.binds {
+            binds
+        } else {
+            name
+        };
```
TeXitoi: You wrote that (or on the form `.as_ref().unwrap_or()`) 5 times. Maybe it would be better to abstract over this? Maybe there is somewhere that such a pattern is forgotten?

korken89:
This LGTM, however if you want to abstract as @TeXitoi says you can add this and we merge.

macros/src/check.rs
japaric: Done. I turned the `binds` into a private field and made an accessor method that requires passing the exception / interrupt name.

japaric:
bors r=korken89

PR #159, opened 2019-02-23
reject duplicate arguments in #[interrupt] and #[exception]

japaric:
This program was being accepted:

``` rust
#[task(
    capacity = 1,
    capacity = 2,
    priority = 1,
    priority = 2,
)]
fn foo() {}
```

now it will trigger a compiler error

r? @korken89 || @TeXitoi


japaric:
bors r=TeXitoi

PR #169, opened 2019-04-16
book: note that late resources must be Send

### LINKED ISSUE #130: Cannot use non-`Send` resource.
In my project I have a resource `struct` with a `*mut`-member, so it's non-`Send`. I want a periodic task to use this resource. As of now it is the only task that needs access to it.

Compilation fails because of `rtfm::export::assert_send` although the [RTFM book] says that 
> you do not need the `Send` trait on types that are transferred between tasks that run at the same priority

Am I missing something or is this a bug?

[RTFM book]: https://japaric.github.io/cortex-m-rtfm/book/by-example/types-send-sync.html#send

japaric:
closes #130


japaric:
Travis is green; merging manually to avoid running CI again (bors)

PR #170, opened 2019-04-16
check task priority at compile time

japaric:
before we were checking the priority at runtime. The compile time error message
when the priority is too high is kind of awful though.


japaric:
bors r=TeXitoi

PR #173, opened 2019-04-19
add sandbox example

etrombly:
Added a real world example for my sandbox project.


etrombly:
I'll add one once I build the frame. Now that the software side is done it shouldn't be too much longer.

AfoHT (approved):
bors merge

PR #175, opened 2019-04-21
document internals

japaric:
note that this assumes that RFC #155 has been implemented

[Rendered text](https://japaric.github.io/rtfm5/book/en/internals.html)

Do not merge this before PR #176


book/en/src/internals/tasks.md
```diff
@@ -1,3 +1,397 @@
...
+                        lock(self.priority(), RQ1_CEILING, || {
+                            // put the taks in the ready queue
+                            RQ1.split().1.enqueue_unchecked(Ready {
+                                task: T1::bar,
+                                // ..
+                            })
+                        });
+
+                        // pend the interrupt that runs the task dispatcher
+                        rtfm::pend(Interrupt::UART0);
```
japaric: I'm currently wondering if this needs a compiler fence to prevent compiler
optimizations from changing the intended semantics.

`rtfm::pend` is a volatile write but `unqueue_unchecked` is a non-volatile
operation so the compiler is free (to some extend) to reorder these two
operations. If we zoom in into these two statements:

``` rust
lock(self.priority(), RQ1_CEILING, || {
    // put the taks in the ready queue
    RQ1.split().1.enqueue_unchecked(Ready {
        task: T1::bar,
        // ..
    })
});

// pend the interrupt that runs the task dispatcher
rtfm::pend(Interrupt::UART0);
```

Assuming that the dynamic priority is equal to the priority ceiling of `RQ1` we
can omit the critical section thus we end up with:

``` rust
RQ1.split().1.enqueue_unchecked(Ready {
    task: T1::bar,
    // ..
})

rtfm::pend(Interrupt::UART0);
```

If we then inline all operations to the maximum we get:

``` rust
// `RQ1` has type `Queue<Ready, const N>` which is equal to
// `struct { tail: AtomicU8, head: AtomicU8, buffer: [Ready; N] }`

/* `RQ1.split().1.enqueue_unchecked(..)` */
let tail = RQ1.tail.load(Ordering::Relaxed);
let ready = Ready { task: T1::bar, /* .. */ };

RQ1.buffer.get_unchecked_mut((tail as usize) % N).write(ready);

atomic::compile_fence(Ordering::Release); // ▲

RQ1.tail.store(tail + 1, Ordering::Relaxed);

/* `rtfm::pend(Interrupt::UART0)` */
ptr::write_volatile(0xE000E200, 1 << 4);
```

The `Release` fence means "preceding reads and writes cannot be moved past
subsequent writes". This is required for memory safety: the `buffer.write`
should *not* be reordered to after the `tail.store`.

However, I think the compiler is allowed to swap the `tail.store` and
`write_volatile` operations. The consequence of that would be that `UART0`
handler could fire and run the task dispatcher which would then see the *old*
`tail` value and miss the just enqueued entry. This doesn't break memory safety
but it would cause the `spawn`-ed task to be missed and not run until some other
task is `spawn`-ed.

I think to be on the safe side we should add another `Release` fence right before
the `write_volatile` (`rtfm::pend`). That is:

``` rust
// ..

RQ1.buffer.get_unchecked_mut((tail as usize) % N).write(ready);

atomic::compile_fence(Ordering::Release); // ▲

RQ1.tail.store(tail + 1, Ordering::Relaxed);

// NEW!
atomic::compile_fence(Ordering::Release); // ▲

/* `rtfm::pend(Interrupt::UART0)` */
ptr::write_volatile(0xE000E200, 1 << 4);
```

Now the compiler can't swap `tail.store` and `write_volatile`. In the generated
code this would be reflected as:

``` rust
lock(self.priority(), RQ1_CEILING, || {
    // put the taks in the ready queue
    RQ1.split().1.enqueue_unchecked(Ready {
        task: T1::bar,
        // ..
    })
});

// NEW! complete the `enqueue` operation before pending the interrupt
atomic::compile_fence(Ordering::Release); // ▲

// pend the interrupt that runs the task dispatcher
rtfm::pend(Interrupt::UART0);
```

I think this should not negatively impact optimizations as there's already a
compiler fence in `enqueue_unchecked`.

perlindgren: Taking a step back:

So what you are touching upon here boils down to what model of computation / semantics we want.

Resources as critical sections.
From a modelling perspective, one can see Resources as specifications of named critical sections, i.e., a way to ensure ordering and mutual exclusion.

From a pure modelling perspective pend is an "async" in RTFM. "async" semantics is that they will execute independently to the sender, inheriting the permissible execution window (base and deadline). "asyncs" can have payload in RTFM, the semantics follow that of sequential ordering data dependencies (i.e., parameters should be computed first). Side effects of synchronous operations before the "async" should be observable to the "pended" task.

Under that model even if the actual BASEPRI does not need to be manipulated, we should have barriers around the lock (to reflect the model).

This have not been formalized but builds on the Timber ideas. 

For this case a single barrier might be sufficient to uphold the model above. (I recall from the original paper RTFM Step 1, that we enforced  barriers at both entry and exit to be on the safe side).

Side note:
In Rust-RTFM we don't yet have a formalized semantics. We can define it the way we want (the only real requirement is that it does not break SRP, and does not break the Rust ownership semantics). What we deal with here is outside of these requirements, but very important to lay the foundation for reasoning about programs in Rust-RTFM, and about giving the user fine grained control.

I think the idea/notion of Resources as named critical sections is a very nice abstraction, allowing you to define logical resources (by maker typed static muts), and use those to provide fine grained ordering and concurrency control. (Of course it also covers the standard use of resources to safely share mutable data).

/Per

book/en/src/internals/tasks.md
```diff
@@ -1,3 +1,397 @@
...
+mod foo {
+    // ..
+
+    pub struct Context<'a> {
+        pub spawn: Spawn<'a>,
+        // ..
+    }
+
+    pub struct Spawn<'a> {
+        // tracks the dyanmic priority of the task
```
burrbull: dynamic

japaric:
@korken89 I have document the overflow behavior.

PR #176, opened 2019-04-21
 implement RFCs 147 and 155, fix #141, etc.

### LINKED ISSUE #141: reject `fn init()` syntax when late resources are used
This syntax allows unsound code as described in PR #140

### LINKED ISSUE #147: [RFC] All functions must be safe
## Summary

Functions marked with any of these attributes: `exception`, `idle`, `init`,
`interrupt` and `task` must be safe `fn`s and not `unsafe fn`s.

## Motivation

Put us in line with rust-lang/rfcs#2585. All the functions inside the `app`
block can only be called by the hardware and are safe to call by the hardware
(the DSL ensures that's the case) -- this should be reflected in the signature.

If a user wants to use unsafe code inside any of these functions they should
explicitly use an `unsafe` block instead of using `unsafe fn` as a short-hand
for `unsafe fn() { unsafe { .. } }`.

## Design

The DSL will reject `unsafe fn`s. For example:

``` rust
#[rtfm::app(..)]
const APP: () = {
    #[init]
    fn init() { .. } // OK

    #[idle]
    unsafe fn idle() -> ! { .. } //~ ERROR: `idle` must have signature `fn() -> !`
};
```

## Drawbacks

Unsafe code will suffer from (even more) rightward drift.

``` rust
#[rtfm::app(..)]
const APP: () = {
    #[init]
    fn init() {
        unsafe {
            // unsafe stuff
        }

        // safe stuff
    }
};
```

### LINKED ISSUE #155: [RFC] explicit `Context` parameter
## Summary

Have all handlers take a `Context` argument as first parameter.

## Motivation

Be less magical. Currently values like `resources`, `spawn` and `schedule` are
directly injected into the value namespace of each handler. This RFC proposes
packing those values in a `Context` struct which appears in the signature of
each handler. See below:

``` rust
// today
#[task(resources = [..], spawn = [..], schedule = [..])]
fn foo() {
    let foo = resources.FOO;
    spawn.bar().ok();
    schedule.baz().ok();
}

// with this RFC
#[task(resources = [..], spawn = [..], schedule = [..])]
fn foo(c: foo::Context) {
    let foo = c.resources.FOO;
    c.spawn.bar().ok();
    c.schedule.baz().ok();

    // spawn.bar().ok(); //~ ERROR value `spawn` is not in scope
}
```

This also makes writing trampolines much easier:

``` rust
#[task(resources = [..], spawn = [..], schedule = [..])]
fn foo(c: foo::Context) {
    some::module::function(c)
}
```

Of course, the layout of `Context` is still auto-generated so which fields are
available in *this* particular `Context` remains more or less a mystery -- this
RFC doesn't help with that.

## Design

All handlers now require that their first parameter is `$handler_name::Context`.
Tasks that take inputs will have their inputs appear as second, third, etc.
parameters.

``` rust
#[app]
const APP: () = {
    // ..

    #[init(resources = [..], spawn = [..], scheduled = [..])]
    fn init(c: init::Context) -> init::LateResources {
        // ..
    }

    #[idle(resources = [..], spawn = [..], scheduled = [..])]
    fn idle(c: idle::Context) -> ! {
        // ..
    }

    // task with no input
    #[task(resources = [..], spawn = [..], schedule = [..])]
    fn foo(c: foo::Context) {
        // ..
    }

    // task with one input
    #[task(resources = [..], spawn = [..], schedule = [..])]
    fn bar(c: bar::Context, input: i32) {
        // ..
    }

    // task with two inputs
    #[task(resources = [..], spawn = [..], schedule = [..])]
    fn baz(c: baz::Context, first: i32, second: i64) {
        // ..
    }

    // task with empty context
    #[task]
    fn quux(c: quux::Context) {
         // ^ parameter is still required in this case
    }
};
```

## Alternatives

In addition to this change we can also flatten resources into
`$handler::Context` , effectively removing the need for the `resources.` prefix.

``` rust
// the main proposal
#[task(resources = [..])]
fn foo(c: foo::Context) {
    c.resources.FOO.lock(|foo| /* .. */);
}

// with this extension
#[task(resources = [..])]
fn foo(c: foo::Context) {
    c.FOO.lock(|foo| /* .. */);
}
```

Note that this extension does *not* run into the drawbacks that plague RFC #146.

## Drawbacks

Application authors would have to write more code: the signature of handlers are
longer and accessing resources and spawning / scheduling tasks requires a
`$context.` prefix per access / invocation (though all those can be replaced
with some de-structuring at the beginning of the handler).

japaric:
This PR:

- Implements RFC 147: "all functions must be safe"

- Implements RFC 155: "explicit Context parameter"

- Implements the pending breaking change #141: reject assign syntax in `init`
  (which was used to initialize late resources)

- Refactors code generation to make it more readable -- there are no more random
  identifiers in the output -- and align it with the book description of RTFM
  internals (see PR #175).

- Makes the framework hard depend on `core::mem::MaybeUninit` and thus will
  require nightly until that API is stabilized.

- Fixes a ceiling analysis bug where the priority of the system timer was not
  considered in the analysis (TODO backport this into the v0.4.x branch).

- Shrinks the size of all the internal queues by turning `AtomicUsize` indices
  into `AtomicU8`s.

- Removes the integration with `owned_singleton`.

closes #141
closes #147
closes #155

Additionally:

- This changes CI to push v0.5.x docs to
  https://japaric.github.io/rtfm5/book/en/ -- we need to do this because our
  official docs are hosted on https://japaric.github.io/cortex-m-rtfm and we
  need to keep them on v0.4.x until we release v0.5.0

- I propose that we use the master branch to develop the upcoming v0.5.0.

- I have created a branch v0.4.x for backports; new v0.4.x releases will come
  from that branch.

r? @korken89 @texitoi, sorry for doing all the impl work in a single commit --
I know that makes things harder to review for you.

Suggestions for compile-pass and compile-fail tests are welcome


examples/pool.rs
```diff
@@ -0,0 +1,67 @@
...
+        rtfm::pend(Interrupt::I2C0);
+    }
+
+    #[interrupt(priority = 2, spawn = [foo, bar])]
+    fn I2C0(c: I2C0::Context) {
+        // claim a memory block, leave it uninitialized and ..
+        let x = P::alloc().unwrap().freeze();
+
+        // .. send it to the `foo` task
+        c.spawn.foo(x).ok().unwrap();
```
TeXitoi: Why `.ok().unwrap()` and not just `.unwrap()`?

korken89:
And great work!

examples/pool.rs
japaric: See https://github.com/japaric/cortex-m-rtfm/pull/179#discussion_r279186524

PR #179, opened 2019-04-22
[backport] book: indirection for faster message passing

### LINKED ISSUE #126: [RFC] Message<T>: avoid copying message payloads
## Summary

Add an opt-in mechanism to avoid copies when receiving messages.

## Background

Message passing always involves a copy on the receiver side. Internally, the
message payload is stored in a `static mut` variable. Right before the receiver
task starts this payload is copied (usually onto the stack) using the
`ptr::read` function to make the payload available *by value*.

This copy can be unnecessary when the task only needs a reference to the
payload. For example:

``` rust
#[rtfm::app(device = ..)]
const APP: () = {
    // ..

    // `packet` will be copied from a `static mut` variable onto `foo`'s stack
    #[task]
    fn foo(packet: [u8; 128]) {
        process(&packet);
    }
};

fn process(packet: &[u8]) {
    // ..
}
```

## Design

This RFC proposes extending the syntax of `#[task]` to provide an opt-in
mechanism to avoid these copies.

When the user specifies `x: rtfm::Message<T>` as the input of the `#[task]` the
codegen pass will avoid copying the payload; instead, the task will receive a
proxy to the payload: `rtfm::Message<T>`. This proxy only allows access *by
reference* to the payload.

The API of `rtfm::Message` is shown below:

``` rust
// crate: rtfm

pub struct Message<T> { /* .. */}

impl<T> ops::Deref for Message<T> {
    type Target = T;

    // ..
}

impl<T> ops::DerefMut for Message<T> {
    // ..
}

impl<T> Message<T> {
    /// Copies the message payload to the stack, freeing the message buffer
    pub fn read(self) -> T {
        // ..
    }
}

impl<T> Drop for Message<T> { /* .. */ }

// Sending `Message<T>` to a higher priority task would break the `Drop` impl
impl<T> !Send for Message<T> { /* ... */ }
```

Here's an updated version of our opening example. It avoids copying `packet`
onto `foo`'s stack.

``` rust
#[rtfm::app(device = ..)]
const APP: () = {
    // ..

    #[task]
    fn foo(packet: Message<[u8; 128]>) {
        process(&packet);
    }
};

fn process(packet: &[u8]) {
    // ..
}
```

It's important to note that using `rtfm::Message` will delay freeing the message
buffer -- remember that RTFM message queues have fixed capacity. This affects
the success of the `spawn` and `schedule` operations.

When `fn task(x: T)` is used the message buffer is freed *before* the task is
started, but when `fn task(x: rtfm::Message<T>)` is used the message buffer is
freed when `x` is dropped / destroyed. The code below illustrates the
difference.

``` rust
#[rtfm::app(device = ..)]
const APP: () = {
    #[init(spawn = [foo, bar, baz])]
    fn init() {
       // these never panic
       spawn.foo(0).unwrap();
       spawn.bar(1).unwrap();
       spawn.baz(2).unwrap();
    }

    #[task(capacity = 1, spawn = [foo])]
    fn foo(x: u32) {
        spawn.foo(x + 1).unwrap(); // this *never* panics
    }

    #[task(capacity = 1, spawn = [bar])]
    fn bar(x: Message<u32>) {
        spawn.bar(x + 1).unwrap(); // this *always* panics

        drop(x); // message buffer is freed here
    }

    #[task(capacity = 1, spawn = [baz])]
    fn baz(x: Message<u32>) {
        drop(x); // message buffer is freed here

        spawn.baz(x + 1).unwrap(); // this *never* panics
    }
};
```

### Why `rtfm::Message<T>` and not `Message<T>`?

Using the latter syntax would break code like this:

``` rust
// User defined struct
struct Message<T> { /* .. */ }

#[rtfm::app(device = ..)]
const APP: () = {
    // ..

    #[task]
    fn foo(x: Message<u32>) {
        bar(x);
    }
};

fn bar(x: Message<u32>) {
    // ..
}
```

The prefix is needed to avoid breaking existing code.

## Benchmark

Consider the following program:

``` rust
#![deny(warnings)]
#![no_main]
#![no_std]

extern crate panic_halt;

use core::ptr;

use cortex_m::asm;

#[rtfm::app(device = stm32f103xx)]
const APP: () = {
    #[init(spawn = [foo])]
    fn init() {
        core.DWT.enable_cycle_counter();

        spawn.foo([0; 128]).ok().unwrap();
    }

    #[task]
    fn foo(x: [u8; 128]) {
        let sum: u32 = x.iter().cloned().map(u32::from).sum();
        unsafe {
            ptr::read_volatile(&sum);
        }
        drop(x);
        asm::bkpt();
    }

    extern "C" {
        fn EXTI0();
    }
};
```

It takes 778 cycles to get to `asm::bkpt` from `EXTI0`. If the signature of
`foo` is changed to `fn(rtfm::Message<[u8; 128]>)` to avoid copying the array
then the execution time reduces to 421 cycles (~46% reduction).

## Alternatives

We could require that the input of `#[task]`s is *always* `Message<T>`. This
is a breaking change because the signature of all existing `#[task]`s would
have to change from `fn(a: A, b: B, ..)` to `(msg: Message<(A, B, ..)>)`.
Also, it's not always beneficial to "avoid the copy". In the case of small
values like `u32` the payload will likely be loaded into a register instead of
being copied onto the stack. The main proposal is more flexible than this
alternative.

## Unresolved questions

- Can this approach be used on the sender side to avoid copies from the stack to
the `static mut` message buffer? For example, an operation like `spawn.foo([0;
128])` will allocate an array on the stack, zero it (memclr) and then copy
(memcpy) it into the message buffer (`static mut` variable).

We could use a newtype over `MaybeUninit<T>` (`Envelope<T>`?) that `DerefMut`s
to `MaybeUninit<T>`. (A.N. This seems like it would deserve its own RFC /
discussion thread).

- Bikeshed alternative names for `rtfm::Message`.

---

There's a reference implementation in #127, but note that, at the time of
writing, the implementation uses `Message<T>` instead of `rtfm::Message<T>` --
this needs to be fixed to fully adhere to the RFC.

japaric:
closes #126 

documents how to use indirection, specifically `heapless::Pool`, to reduce the overhead of message passing.


examples/pool.rs
```diff
@@ -0,0 +1,67 @@
...
+        rtfm::pend(Interrupt::I2C0);
+    }
+
+    #[interrupt(priority = 2, spawn = [foo, bar])]
+    fn I2C0() {
+        // claim a memory block, leave it uninitialized and ..
+        let x = P::alloc().unwrap().freeze();
+
+        // .. send it to the `foo` task
+        spawn.foo(x).ok().unwrap();
```
TeXitoi: Same remark, why `.ok()`

japaric: TL; DR because `[T; 128]` doesn't implement the `fmt::Debug` trait.

`unwrap` requires that the error type implements `Debug` but it doesn't in this case. The error type is the value the message that couldn't be sent; in this case the value has type `Box<A>` and that only implements when `A::Type: Debug`, which in this case it's `[u8; 128]` but `[T; N]` implement Debug only up to N = 32.

japaric:
bors try

japaric:
bors try

japaric:
bors try

japaric:
bors try

japaric:
Finally ... Merging manually

PR #180, opened 2019-04-26
 Update cortex-m version requirement to allow new 0.6.0 version

david-sawatzke:


korken89:
I made another pull which fixes the issue in this one. Thanks for PRing!

PR #181, opened 2019-04-28
Fix to make all work with cortex-m 0.6.0

korken89:


korken89:
With this it should be compatible with the new `cortex-m` version. r? @TeXitoi @japaric

japaric:
This is no longer necessary as cortex-m is using the semver trick to let v0.5.x and v0.6.x interoperate.

I would prefer to keep the version requirement as it is (i.e. specifying the minimum minor version rather than a minor version range).

TeXitoi:
@korken89 can we close this?

korken89:
Sure, I think we are ok

burrbull:
`cargo tree` doesn't work with cortex-m 0.5.8
```
error: no matching package named `cortex_m_0_6` found
location searched: registry `https://github.com/rust-lang/crates.io-index`
required by package `cortex-m v0.5.10`
    ... which is depended on by `cortex-m-rtfm v0.5.0-alpha.1 (https://github.com/japaric/cortex-m-rtfm#fafeeb27)`
```

PR #185, opened 2019-05-01
unlock the DWT, if locked, when using the timer queue

### LINKED ISSUE #184: timer-queue: need to unlock DWT
> RTFM schedule doesn't work on STM32F7xx. After some test at Oxydize with Jorge, we found that the DWT timer is never started. It needs some more configuration.
By adding the line unsafe { (*DWT::ptr()).lar.write(0xC5ACCE55) }; it works.
It should be added to the cortex-m DWT init for STM32F7xx

Originally reported by @musitdev in rust-embedded/cortex-m#144

japaric:
closes #184


japaric:
this is no longer applicable to the v0.5.x releases because configuration is now to be done by the application author. However, it would be good to backport this PR into the v0.4.x release cc @korken89 .

PR #186, opened 2019-05-01
[backport] unlock the DWT, if locked, when using the timer queue

japaric:
**HEADS UP** this PR targets the branch v0.4.x

### DIFF
--- a/macros/src/codegen.rs
+++ b/macros/src/codegen.rs
@@ -2024,6 +2024,15 @@ fn pre_init(ctxt: &Context, app: &App, analysis: &Analysis) -> proc_macro2::Toke
 
     // Set the cycle count to 0 and disable it while `init` executes
     if cfg!(feature = "timer-queue") {
+        exprs.push(quote!(
+            // unlock the DWT, if software locked
+            // See ARM CoreSight Architecture Specification v3.0, section B2.3.10
+            // 1 << 0 = Software lock mechanism is implemented.
+            // 1 << 1 = Writing to the other registers in the component is blocked
+            if p.DWT.lsr.read() & 0b11 == 0b11 {
+                p.DWT.lar.write(0xC5ACCE55);
+            }
+        ));
         exprs.push(quote!(p.DWT.ctrl.modify(|r| r & !1);));
         exprs.push(quote!(p.DWT.cyccnt.write(0);));
     }


japaric:
bors try

musitdev:
I made some test and I get some strange result that why It took me some time. I've test on two board, one discovery F746 and another board (st-link deveice) with the same processor. I use Ubuntu 1604, openocd 0.10.0 and arm gdb 8-2018-q4-major, rust 1.36.0. First test with RTFM version "0.4.3". What I see with a simple one task scheduled that blink: 
 * no modification of DWT: don't blink
 * modification of DWT blink.
 * after the last test test with no modification of DWT: blink.
 * reset no modification of DWT: blink.
 * power of / power on, no modification of DWT: don't blink
 * modification of DWT: blink.

It seem that to blink the modification of DWT is needed but once it has been done, it's done until the power of.
After I've tested with the pull request version and I didn't manage to make it blink event if I modifiy the DWT or make it blink and after start the pull request version.

japaric:
@musitdev debug settings tend to be "sticky"; they stick until you power cycle (turn it off and then turn it on) the device.

If I'm reading your comment correctly this PR didn't work? If it didn't work could you report the value of DWT_LSR after power cycling the device?

musitdev:
Could you give me the command to get DWT_LSR values and when to apply then?

japaric:
@musitdev You can do it in Rust with a program like this:

``` rust
#[entry] 
fn main() -> ! {
    let p = cortex_m::Peripherals::take().unwrap();
    println!("{}", p.DWT.lsr.read());
}
```

Or you can do it with just OpenOCD:

``` console
$ # run openocd -f (..) in another terminal
$ telnet 4444
> mdw 0xE0001FB4
```

In either case you should power cycle the device first. I think I read somewhere that the contents of that register returns different values when read by the device vs when read by the debugger so it's best to try both ways.

With the first approach you should remove anything not essential from openocd.gdb (like `monitor tpiu (..)`) and it would be best if you read the register without relying on the debugger (print it using a serial interface rather than semihosting).

With the second approach you should flash an empty program (`loop {}` / no-rtfm) before power cycling the device.

---

Alternatively, you could try a variation of this PR that always writes to the LAR register (i.e. drop the `if LSR == 0b11`) and see if that works after power cycling the device. If that works we can go with that, it shouldn't cause problems to devices that don't have a "software lock", I think.

korken89:
Is there something holding this PR back?

musitdev:
I have to make some test. Sorry I didn't have must time. I've prepared the test but didn't made them. I think i'll do them this week.

korken89:
Thanks for the update

musitdev:
Finally I manage to finish the test. On a Nucleo F746ZG, arm-none-eabi-gd 7.10 , OpenOCD 10.
OpenOCD command mdw 0xE0001FB4 return 00000001
I made a prg that turn on a led when val = p.DWT.lsr.read(); is 1. 
With openocd started the led is on even after a reset.
I flash the prg.
If I power off/on the led is off, make some reset off. I start OpenOCD the led turn on.
I stop openOCD. led on. I do a reset, led on. Power on/off, led off.
To have DWT on, I must have openocd started once.

korken89:
bors retry

PR #188, opened 2019-05-02
bump heapless version to v0.4.4

### LINKED ISSUE #182: When using min-const-fn, SPSC can fail to be created
We noticed at the Oxidize Workshop that there is a way to get the SPSC queue to fail to be made, see the example below.
I checked the `heapless` implementation, and I was not able to get this to happen without RTFM, so maybe the issue is in here.

Any ideas? @japaric @TeXitoi 
Original code: https://github.com/korken89/rtfm_workshop

```
#![no_main]
#![no_std]

// panic handler
extern crate dwm1001;
extern crate panic_semihosting;

use cortex_m_semihosting::hprintln;
use dwm1001::nrf52832_hal as hal;
use hal::{DMAPool, TXQSize};

use heapless::{
    pool::singleton::Box,
    spsc::{Consumer, Producer, Queue},
};

use rtfm::app;

#[app(device = crate::hal::target)]
const APP: () = {
    static mut PRODUCER: Producer<'static, Box<DMAPool>, TXQSize> = ();
    static mut CONSUMER: Consumer<'static, Box<DMAPool>, TXQSize> = ();

    #[init(spawn = [])]
    fn init() -> init::LateResources {
        // for the producer/consumer of TX
        static mut TX_RB: Option<Queue<Box<DMAPool>, TXQSize>> = None;

        hprintln!("init").unwrap();

        *TX_RB = Some(Queue::new());
        let m = TX_RB.as_mut().take();

        // Should always be some
        match &m {
            Some(_) => hprintln!("Some ok").unwrap(),
            None => hprintln!("Danger danger").unwrap(), // <-- This happens!!!! D:
        }

        let (txp, txc) = m.unwrap().split();

        init::LateResources {
            PRODUCER: txp,
            CONSUMER: txc,
        }
    }
};
```

japaric:
appears to fix #182

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,7 +46,7 @@ cortex-m-rtfm-macros = { path = "macros", version = "0.5.0-alpha.1" }
 
 [dependencies.heapless]
 features = ["smaller-atomics", "min-const-fn"]
-version = "0.4.3"
+version = "0.4.4"
 
 [dev-dependencies]
 cortex-m-semihosting = "0.3.2"


japaric:
This seems to not have fixed #182 so there's no much point in bumping the version of this dependency. Closing.

PR #190, opened 2019-05-08
book: document scope and limitations of the schedule, Instant and Duration APIs

### LINKED ISSUE #121: Instant::elapsed() may give incorrect Duration values
As `Instant`s are created using solely the DWT, they are limited to a range of 32 bits. For example, the nRF52832 has a fixed system clock of 64MHz. This means that the cycle count will roll over every 67 seconds or so.

If you do the following:

```rust
let start = Instant::now(); // Lets say this happens at cycle 0xFFFF_FFF0
// ... some time passes
let elapsed = start.elapsed(); // Lets say this happens just after the rollover, at the next 0x0000_000F

// `elapsed` will now be (0x0000_000F - 0xFFFF_FFF0)
// which in debug mode *should* panic, and in release mode will == 31, which is correct

// Now we wait 670 seconds or so, when 10 more rollovers have happened
// timer is now 0x0000_000F after all of those rollovers
let elapsed_2 = start.elapsed();

// `elapsed` will now be (0x0000_000F - 0xFFFF_FFF0)
// which is still `31`, and incorrect!
```

Many devices have a much faster core clock than 64MHz, encountering this more often.

The larger issue is a case of aliasing. Let's say you are writing a debounce algorithm, which fires on every rising OR falling edge of a button press.

```rust
match state {
    RisingEdge => {
        *TRIGGER_TIME = Instant::now();
    }
    FallingEdge => {
        if *TRIGGER_TIME.elapsed() >= 16_000_000.cycles() { // 250ms at 64MHz
             do_something();
        } else {
             // do nothing, bouncy trigger
        }
    }
}
```

Here's what we would get, depending on the `cycle_count` time, if the RisingEdge occurred at `0x8000_0000`:

| Time (cycles) at Falling Edge | Duration (cycles) since Rising Edge | Behavior | Correct? |
| :---                                       | :---                                            | :---         | :---         |
| 0x80f4_23ff | 15_999_999 | do nothing | YES |
| 0x80f4_2401 | 16_000_001 | do_something() | YES |
| 0xffff_ffff | 2_147_483_647 | do_something() | YES |
| 0x0000_0000 (+1 rollover) | 2_147_483_648 | do_something() | YES |
| 0x7f0b_dc00 (+1 rollover) | 4_278_967_296 | do_something() | YES |
| 0x8000_0001 (+1 rollover | 1 | do nothing | NO |

japaric:
closes #121

### DIFF
--- a/book/en/src/by-example/timer-queue.md
+++ b/book/en/src/by-example/timer-queue.md
@@ -95,3 +95,32 @@ Running the program on real hardware produces the following output in the consol
 ``` text
 {{#include ../../../../ci/expected/baseline.run}}
 ```
+
+## Caveats
+
+The `Instant` and `Duration` APIs are meant to be exclusively used with the
+`schedule` API to schedule tasks *with a precision of a single core clock
+cycle*. These APIs are *not*, for example, meant to be used to time external
+events like a user pressing a button.
+
+The timer queue feature internally uses the system timer, `SysTick`. This timer
+is a 24-bit counter and it's clocked at the core clock frequency so tasks
+scheduled more than `(1 << 24).cycles()` in the future will incur in additional
+overhead, proportional to the size of their `Duration`, compared to task
+scheduled with `Duration`s below that threshold.
+
+If you need periodic tasks with periods greater than `(1 << 24).cycles()` you
+likely don't need a timer with a resolution of one core clock cycle so we advise
+you instead use a device timer with an appropriate prescaler.
+
+We can't stop you from using `Instant` to measure external events so please be
+aware that `Instant.sub` / `Instant.elapsed` will never return a `Duration`
+equal or greater than `(1 << 31).cycles()` so you won't be able to measure
+events that last more than `1 << 31` core clock cycles (e.g. seconds).
+
+Adding a `Duration` equal or greater than `(1 << 31).cycles()` to an `Instant`
+will effectively overflow it so it's not possible to schedule a task more than
+`(1 << 31).cycles()` in the future. There are some debug assertions in place to
+catch this kind of user error but it's not possible to prevent it with 100%
+success rate because one can always write `(instant + duration) + duration` and
+bypass the runtime checks.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -126,6 +126,16 @@ pub struct Peripherals<'a> {
 /// A measurement of a monotonically nondecreasing clock. Opaque and useful only with `Duration`
 ///
 /// This data type is only available when the `timer-queue` feature is enabled
+///
+/// **NOTE:** Both `Instant` and `Duration` are only meant to be used with the `schedule` API.
+///
+/// We can't stop you from using `Instant.sub` to measure things so please be aware that you won't
+/// be able to measure events that span longer than `(1 << 31) - 1` core clock cycles.
+///
+/// Also note that adding a duration equal or greater than `(1 << 31).cycles()` to an `Instant` will
+/// effectively overflow it. There's a debug assertion in place to prevent this user error but it's
+/// not possible to catch this error with 100% success rate because one can write `(instant +
+/// duration) + duration` to bypass runtime checks.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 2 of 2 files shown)


jamesmunns:
Would it make sense to call these something other than a ~~hammer~~ Instant or Duration to prevent confusion?

:+1: from me re: the improved docs either way!

perlindgren:
In the RTFM task/resource model, we had just a baseline for each event/message (which was an absolute point time). For each event/message there was also a relative deadline, and minamal inter-arrival time. The two latter is useful to reason about timing properties. 
Asynchronous messages could be given a relative time offset and optionally o new relative deadline (else inherited from the sender). All designed to be simple to understand and reason about.

IMHO the cortex-m-rtfm implementation should strive to be as simple as to understand use (and reason about) as possible. I'm a bit concerned that we introduce too many types/primitives, that blurs the picture, and potentially makes the system less amenable to static analysis.

korken89:
One thought here, why not make the Instant be a 64 bit number? I played with that idea here, and it seems like it might be relevant: https://github.com/korken89/trustflight_firmware/blob/198e21f5065e2803c4fca3f5bf0e28b2ef675b9b/board/src/lib.rs#L183
This codes needs 43 clocks to get a time when I measured it, which can act as reference. 

Then there is no confusion and the time is valid and monotonic "forever".

japaric:
@jamesmunns I think `Instant` and `Duration` are the right names for these abstractions; at least I can't think of better names. Also they are prefixed by `rtfm::`; they are not  [`core::time::Duration`](https://doc.rust-lang.org/core/time/struct.Duration.html) or some `hal::Instant`, which is how I would name what @korken89 posted.

@perlindgren stripping down the `Instant` and `Duration` APIs or even removing them is certainly an option. I think they are still useful to provide.

@korken89 I worry about the overhead of this requirement:

``` rust
// Note: This function assumes it will be called at least once per overflow (about once
// every 59.7 seconds if the MCU is running at 72 MHz)
```

There's no interrupt associated to the DWT; the only timer interrupt RTFM can rely on is the SysTick which is a 24-bit timer. For correctness, we would have to fire the SysTick every `(1 << 24).cycles()` (~200ms @ 72MHz) to make sure we don't miss a DWT overflow -- we can't rely on the user periodically `schedule`-ing any task even if they declare them in their `#[task]`s.

Requiring a "disable all interrupts"-style critical section in `Instant::now` also seems unfortunate. We could replace that critical section with a *resource* but then tasks would need to declare their intention to use the `Instant` API and `Instant::now` would have to become a method on `$task::Context`.

perlindgren:
@japaric
Requiring a "disable all interrupts"-style critical section in Instant::now also seems unfortunate. We could replace that critical section with a resource but then tasks would need to declare their intention to use the Instant API and Instant::now would have to become a method on $task::Context.

Isn't 'Instant::now' is just a read of the DWT, and in such case atomic (and non destructive). Or are you referring to a 64 bit Instant, that would need more trickery?

It might be me missing something here...
/Per

perlindgren:
.. at some point we should have a look at a trait for the timer driving the scheduler, so that we could use any HW timer. 

Regarding systick. The systick can (at least on some platforms if recall correctly) be divided by 8. This makes sense, there is little or no point in having timing accuracy at cycle clock level, as we have a dispatch OH that is larger, and it would reduce the OH of handling systicks. We don't need to run the systick at all if there are nothing in the timer queue (I did not check your impl, but in the prototype impl of the paper regarding this topic, we disabled the interrupt in such cases). That wont work if we want a 64 bit timer though (as we need to know when the DWT has wrapped).

I think a general purpose 32 bit timer would solve this issue for us (as proposed in the paper). 

Best
  Per

korken89:
In the C++ version of RTFM I used the 64-bit version and called the time function in each systick overflow, but if it is the view that "time" should not be provided in RTFM here I think there is no need to continue that discussion and leave it for some HAL-ish implementation.

japaric:
as it is this is no longer applicable to the v0.5.x releases because the monotonic is now user defined. However, it would be good to backport this PR into the v0.4.x release cc @korken89 .

Issue #204, opened 2019-06-14
[RFC] multi-core support


japaric:
EDIT(2019-06-18): revised with both homogeneous and heterogeneous multi-core support and better documented the contract with the `{device}` crate.

## Summary

Add *experimental* heterogeneous and homogeneous multi-core support behind
Cargo features.

### Semantics

In multi-core applications each core will basically run its own independent RTFM
sub-application. Each core will have its own tasks and resources (`static mut`
variables) and the cores will be able to communicate using message passing:
cross-core `spawn` & `schedule` API.

It's not possible to share resources (`static mut` variables) between cores.
However, it is possible for one core to initialize another core's resources and
it's also possible to share `static` variables between cores.

### Feature gating

The heterogeneous multi-core support will live behind a `heterogeneous` Cargo
feature considered *experimental*. Likewise, homogeneous multi-core support will
be gated by a `homogeneous` Cargo feature.

### Semver exception

We'll reserve the right to do breaking changes to both multi-core APIs in patch
releases and will recommend end users to pin to a exact version of
`cortex-m-rtfm` when using either API (i.e. `=0.5.0` in `Cargo.toml`).

It's unlikely that that these breaking changes will include changes in the
end-user API or the syntax of the `#[rtfm::app]` macro. The breaking changes we
foresee are at the level of the binary contract with the `{device}` crate.
There's no multi-core equivalent to the `cortex-m-rt` crate or
`svd2rust`-generated crates; we eventually will move to whatever the ecosystem
settles on -- this is a breaking change we expect to occur but want to avoid it
requiring a minor version bump for users of the single-core API.

### Syntax

The `#[app]` attribute will gain a `cores` argument that indicates the number of
cores the application will use. The argument expects an integer value equal
or greater than 2. Omitting this argument indicates that the application is a
single-core application.

``` rust
// dual-core application
#[rtfm::app(cores = 2)]
const APP: () = {
   // ..
};
```

In multi-core mode all tasks, `#[init]` and `#[idle]` functions need to indicate
in which core they'll run using the `core` argument.

``` rust
// dual-core application
#[rtfm::app(cores = 2)]
const APP: () = {
    // core #0 initialization function
    #[init(core = 0)]
    fn init(c: init::Context) {
        // cross-core message passing
        c.spawn.ping(0).ok();
    }

    #[task(core = 0)]
    fn pong(c: pong::Context, x: u32) {
        c.spawn.ping(x + 1).ok();
    }

[...]

Issue #203, opened 2019-06-14
[RFC] supporting other architectures


japaric:
## Current state

`cortex-m-rtfm` is the main port of the RTFM framework. There are two outdated
ports of RTFM for MSP430 and RISCV and two experimental Cortex-R and Linux
ports. None of these ports shares code with one another so they have diverged or
stalled since their inception.

The `cortex-m-rtfm` port is under @japaric's GitHub account and currently has 3
people maintaining it.

## Proposal

This RFC proposes refactoring out the RTFM syntax (meta-language) and analysis
pass that currently live in `cortex-m-rtfm` into a crate that can be used to
easily implement ports for different architectures.

## Rationale

We want to make RTFM available on as many architectures as possible with minimal
maintenance effort and with as similar APIs / syntax as possible.

## Detailed design

This RFC proposes the following crate hierarchy:

### `rtfm-core`

This crate will only contain the `Mutex` trait and the `Exclusive` newtype and
it's meant to be used to implement generic library code.

``` rust
pub trait Mutex {
    type Data;

    fn lock<T>(&mut self, f: impl FnOnce(&mut Self::Data) -> T) -> T;
}

pub struct Exclusive<T>(pub &'a mut T);

// trivial implementation (no-op)
impl<T> Mutex for Exclusive<T> { .. }
```

### `rtfm-syntax`

This crate will contain the parser of the `#[app]` and the analysis pass (e.g.
priority ceiling analysis) of the RTFM specification.

The main function in this crate is the `parse` function which parse the input of
the `#[app]` attribute.

``` rust
use syn::Parse;

use crate::{ast::App, analyze::Analysis};

pub fn parse(
    args: TokenStream,
    input: TokenStream,
    settings: Settings,
) -> parse::Result<(App, Analysis) {
    ..
}
```

To accommodate the different capabilities of the many RTFM ports this function
takes a `Settings` struct that indicates which *extensions* the port implements.

``` rust
pub struct Settings {
    // multi-core support: `#[app(cores = ..)]`, `#[task(core = ..)]`, etc.
    pub parse_cores: bool,

    // `#[exception]`
    pub parse_exception: bool,

    // `#[interrupt]`
    pub parse_interrupt: bool,

    // `schedule` API
    pub parse_schedule: bool,

    // `extern { fn interrupt(); }`
    pub parse_extern_interrupt: bool,

    // ..
}
```

### `$ARCH-rtfm`

[...]

Issue #201, opened 2019-06-14
[RFC] opt-in `Peripherals` steal


japaric:
## Current behavior

The `init::Context` struct includes a `device` field whose type is
`#device::Peripherals`, a singleton that's a collection of all the device
peripherals.

## Proposal

Don't include this `device` field by default. Add a `peripheral` argument to the
`#[rtfm::app]` attribute that lets the user opt into today's behavior.

## Rationale

`#device::Peripherals` is a singleton; in multi-core mode only *one* core can
take *all* the peripherals so it's necessary to specify which core will take the
singleton.

## Detailed design

The `#[rtfm::app]` attribute will gain a `peripherals` argument whose semantics
varies between single-core and multi-core mode.

### Single-core mode

In single core mode the argument takes a boolean that indicates whether the
device peripherals will be made available during the `init` function. When the
`peripherals` argument is omitted the peripherals will not be made available.

``` rust
// NOTE omitting `peripherals` also means `peripherals = false`
#[rtfm::app(device = pac, peripherals = false)]
const APP: () = {
    #[init]
    fn init(c: init::Context) {
        // let device: pac::Peripherals  = c.device;
        //~^ error: no field named `device` in this struct
    }
};
```

``` rust
#[rtfm::app(device = pac, peripherals = true)]
const APP: () = {
    #[init]
    fn init(c: init::Context) {
        // OK
        let device: pac::Peripherals  = c.device;
    }
};
```

### Multi-core mode

In multi-core mode the `peripherals` argument takes a integer that indicates
which core will take the device peripherals during `init`. If the argument is
omitted then the peripherals will not be taken by the framework.

``` rust
#[rtfm::app(cores = 2, device = pac, peripherals = 1)]
const APP: () = {
    #[init(core = 0)]
    fn init(c: init::Context) {
        // let device: pac::Peripherals  = c.device;
        //~^ error: no field named `device` in this struct
    }

    #[init(core = 1)]
    fn init(c: init::Context) {
        // OK -- this core takes all peripherals
        let device: pac::Peripherals  = c.device;
    }
};
```

## Alternatives

In single-core mode, should the default be to have RTFM take the peripherals?
That is `peripherals` would default to `true` matching today's behavior and one
would have to opt-out by writing `peripherals = false`.

One could design a syntax to split the peripherals between the many `init`
functions in multi-core mode but this can't be implemented today's without extra
help from `svd2rust` so *this* proposal doesn't delve into that question.

[...]

Issue #200, opened 2019-06-14
[RFC] user defined `monotonic` timer


japaric:
## Current behavior

The `schedule` API internally makes uses of two timers: the `DWT_CYCCNT` (the
cycle counter) as a monotonic timer / counter and the `SysTick` (system timer)
to generate timeouts. Currently, these can't be changed and the result is that
the `schedule` API can't be used on ARMv6-M where `DWT_CYCCNT` doesn't exist.

## Proposal

Require that the user specifies the monotonic timer that will be used to
implement the `schedule` API. The user will still be able to use `DWT_CYCCNT` as
the monotonic timer but this will *not* be the default; the user must specify
this timer, or some other timer, in their application.

This RFC does *not* propose a mechanism to sub the `SysTick` timer with a
different timer.

## Rationale

The `DWT_CYCCNT` is not an appropriate monotonic timer for multi-core
applications: each Cortex-M core has its own `DWT` peripheral and its own cycle
counter so it's not possible to synchronize these two counters (there's no
register to do this), plus the counters may be operating at different frequencies
resulting in one core's `Instant` having a different meaning that other cores'
`Instant`s. Also, as was previously mentioned the `DWT_CYCCNT` is not available
on ARMv6-M cores; this limits on which devices one can use the `schedule` API.

In multi-core applications it's better to use a device specific, constant
frequency timer visible to all cores as the monotonic timer. And in single-core
applications one may want to use a 64-bit timer or a prescaled 32-bit timer as
the monotonic timer; this way one can `schedule` tasks with long periods, e.g.
in the order of seconds (cc @jamesmunns).

Lastly, using a device-specific timer as the monotonic timer lets programmers
use the `schedule` API on single-core ARMv6-M devices.

### The `Monotonic` trait

The following trait will be added to the `cortex-m-rtfm` crate.

``` rust
struct Fraction { num: u32, den: u32 }

/// A monotonic clock / counter
pub trait Monotonic {
    /// A measurement of this clock
    type Instant: Copy + Ord + Sub;

    /// The ratio between the SysTick (system timer) frequency and the frequency of this clock
    fn ratio() -> Fraction;

    /// Returns the current time
    ///
    /// NOTE this must be an atomic operation
    fn now() -> Self::Instant;

    /// Resets the counter to *zero*
    ///
    /// NOTE the runtime will execute this method with interrupts disabled
    unsafe fn reset();

    /// A `Self::Instant` that represents a count of *zero*
[...]

PR #205, opened 2019-06-14
rtfm-syntax refactor + heterogeneous multi-core support

### LINKED ISSUE #163: Add Cargo feature to disable handling of `Peripherals`
RTFM currently acquires a `Peripherals` instance and hands it to the user. This can be problematic, as it doesn't compose with libraries that want to do the same. While it can be a good idea to change the affected library, there may be valid reasons for wanting to control the creation of `Peripherals`.

One such example would be HAL libraries that want to provide strong static guarantees without introducing runtime overhead. I believe it can be argued that a HAL library, which is intended to be the primary means to access a given device, would be within its rights to require full control over `Peripherals`, even though that doesn't compose with other libraries.

@japaric and I talked about this on IRC yesterday, and he said that RTFM's handling of `Peripherals` is just a convenience feature (my understanding is that RTFM only actually requires `CorePeripherals`), and that he wouldn't be opposed adding a Cargo feature that disables this. This sounds like a good solution to me.

### LINKED ISSUE #178: [RFC] default all task capacities to 1
## Summary

If the user didn't specify the capacity of a task default to a capacity of `1`.

## Motivation

Today when the capacity of a task is not specified the framework tries to pick
a capacity that reduces the chances of `spawn` / `schedule` failing at runtime.
To do so the framework assumes that (a) each context that may spawn / schedule
a particular task `X` will do so at most *once* per invocation and (b) that such
context runs not so often that the task will be dispatched before the context
runs again.

This works reasonably well sometimes and sometimes it can lead to more memory
usage than necessary. More importantly, though, it makes it hard to estimate the
capacity of a task from just its attribute. Quick, what are the capacities of
tasks `foo` and `bar` in the following program?

``` rust
#[app(device = /* .. */)]
const APP: () = {
    #[task(spawn = [bar])]
    fn foo(c: foo::Context) {
        // ..
    }

    #[task(spawn = [bar])]
    fn bar(c: bar::Context) {
        // ..
    }

    #[init(spawn = [bar])]
    fn init(c: init::Context) {
        // ..
    }

    extern "C" {
        fn EXTI1();
    }
};
```

The answer is `0` (yes, *zero*) and `3`, respectively. `foo` can never be
`spawn`-ed so its capacity is `0`; `bar` appears in three `spawn` lists so
its capacity is `3`. Modifying the program can make these numbers fluctuate,
making things even harder to predict.

Therefore I propose that the capacity of a task defaults to the smallest
(sensible) value of `1` when not specified. This behavior makes capacities fully
predictable.

## Design

Drop the logic used to guess an appropriate task capacity and simply default to
a value of `1` if the capacity is not specified by the user.

## Alternatives

We could ask the user to specify the capacity of all tasks, that is we could
have no default. However, this would burden the end user with deciding on a
capacity as soon as they add a new task. This could break the development flow
and hurt productivity; choosing the best / right capacity requires looking at
the program as a whole and it's likely best done in later stages of development.

japaric:
this PR implements RFCs #178, #198, #199, #200, #201, #203 (only the refactor
part), #204, #207, #211 and #212.

most cfail tests have been removed because the test suite of `rtfm-syntax`
already tests what was being tested here. The `rtfm-syntax` crate also has tests
for the analysis pass which we didn't have here -- that test suite contains a
regression test for #183.

the remaining cfail tests have been upgraded into UI test so we can more
thoroughly check / test the error message presented to the end user.

the cpass tests have been converted into plain examples

EDIT: I forgot, there are some examples of the multi-core API for the LPC541xx in [this repository](https://github.com/japaric/lpcxpresso54114)

people that would like to try out this API but have no hardware can try out the
x86_64 [Linux port] which also has multi-core support.

[Linux port]: https://github.com/japaric/linux-rtfm

closes #178 #198 #199 #200 #201 #203 #204 #207 #211 #212 
closes #163 
cc #209 (documents how to deal with errors)


japaric:
All the RFCs that have been implemented in this PR (list in PR description) have been accepted, the single-core API docs have been updated and CI is green so this PR is ready to land.

Once it lands we can do a pre-release (beta release) under the `rtfm` moniker to let users try out the new syntax / API.

Before the proper release we need to make a decision regarding the crate name (#208) and document the multi-core API (#227)

r? @korken89 || @TeXitoi

japaric:
:tada: 

bors r+

japaric:
https://travis-ci.org/japaric/cortex-m-rtfm/builds/585246392 CI passed but bors didn't merge it automatically so I'll merge this PR manually.

PR #221, opened 2019-07-26
Enable the DWT unit before poking at its registers. (v0.4.x)

### LINKED ISSUE #196: Schedule does not work properly in init after connecting power.
This one is a bit weird. 
I set up a periodic task following the example from the book. The task is suppose to run once every 5ms. After connecting power, it looks like the task only starts running 10 seconds after init, and from there it runs properly (5ms period). But if I press the reset button just after connecting power everything works as it should.

I just encounter the problem if I completely disconnect power and reconnect and don't press the reset button. I changed the schedule in init to spawn and maintained the schedule inside the software task and now it works as it should even just after connecting power.

I really don't know why pressing reset would be that different from connecting power for the first time.

I am testing this on a stm32f103c8 (bluepill).

jmgao:
On cold boot, the DWT unit is off, which means our attempts to clear and
disable the cycle counter in pre-init don't work. This seems to result
in access to CYCCNT always returning 1073741824, which delays tasks
scheduled in init by that many cycles (~15s at 72 MHz).

Fixes #196.

### DIFF
--- a/macros/src/codegen.rs
+++ b/macros/src/codegen.rs
@@ -507,7 +507,6 @@ fn post_init(ctxt: &Context, app: &App, analysis: &Analysis) -> proc_macro2::Tok
 
     // Enable cycle counter
     if cfg!(feature = "timer-queue") {
-        exprs.push(quote!(p.DCB.enable_trace()));
         exprs.push(quote!(p.DWT.enable_cycle_counter()));
     }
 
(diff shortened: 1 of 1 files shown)


korken89:
Hi, can you provide a minimal example to test this with?
I am having issues replicating this on an nRF52832.

jmgao:
I've uploaded a minimal light blinking repro for the F103 bluepill at https://github.com/jmgao/rtfm_repro_196.
If you start it with the stlink, it'll start blinking immediately, if you disconnect the power and reconnect it without the debugger, it won't start blinking until ~15 seconds post-boot.

Maybe it's specific to the the Cortex-M3? I couldn't find anything in the reference manual specifying behavior when you read the CYCCNT register when DWT is off.

jmgao:
Hm, doesn't seem like it, I see the same behavior on an F303, which is a cortex-M4: https://github.com/jmgao/rtfm_repro_196/tree/f303

korken89:
Thank you for the example!

I am still unable to confirm this with the board I have at hand (nRF52832). 
However, the fix you propose does make sense and still works on the board I have. 
I will approve it, thanks for finding this corner case! Good job!

korken89:
@japaric Will you merge?

japaric:
@jmgao thanks!

PR #222, opened 2019-07-26
Enable the DWT unit before poking at its registers.

### LINKED ISSUE #196: Schedule does not work properly in init after connecting power.

jmgao:
On cold boot, the DWT unit is off, which means our attempts to clear and
disable the cycle counter in pre-init don't work. This seems to result
in access to CYCCNT always returning 1073741824, which delays tasks
scheduled in init by that many cycles (~15s at 72 MHz).

Fixes #196.

### DIFF
--- a/macros/src/codegen.rs
+++ b/macros/src/codegen.rs
@@ -1204,6 +1204,8 @@ fn pre_init(app: &App, analysis: &Analysis) -> Vec<proc_macro2::TokenStream> {
 
     // Set the cycle count to 0 and disable it while `init` executes
     if cfg!(feature = "timer-queue") {
+        // We need to explicitly enable the trace block to set CYCCNT.
+        stmts.push(quote!(core.DCB.enable_trace();));
         stmts.push(quote!(core.DWT.ctrl.modify(|r| r & !1);));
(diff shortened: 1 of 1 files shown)


korken89:
@japaric Will you merge?

japaric:
As per the approved RFC #200 initialization of the monotonic timer is the job of the end user; the framework no longer touches the DWT peripheral in the general case (it does call `Monotonic::reset` which may modify CYCCNT if rtfm::CYCCNT is being used).

Closing this PR as it would no longer apply once PR #205 lands.

PR #225, opened 2019-08-09
A fix that stops panics from the old MaybeUninit impl

korken89:
Sometimes when running in debug mode the current implementation of `MaybeUninit` would panic, an example can be found here: https://github.com/korken89/nrf52-hal/blob/2bb89427fa05a7b04a563be0c84f40969b3bb4df/examples/rtfm-uarte-interrupts/src/main.rs#L106
On debug this line will always panic, on release all is fine.

The issue seemed to be the manual `MaybeUninit` implementation, so I have merged the two implementations together.
Please have a look and see if my fix is unsound (it does however work, but not sure on UB).

cc @japaric @TeXitoi


TeXitoi:
Can't we just remove the wrapper and avoid using `MaybeUnitit::write`?

korken89:
We can, I do not quite understand the proc macro monster well enough to find all that need changing. I wrote to japaric to see if he has time to sit down and help me a little the first time, but I am still waiting for an answer.

TeXitoi:
The wrapper interface seems to be in sync with the real implementation, thus only `use` declaration should be changed to make it works. `write` is basically an affectation followed by a `get_mut`.

Sorry, but I don't have the time in the short term to work on that.

src/export.rs
```diff
@@ -86,61 +84,28 @@ impl<T> MaybeUninit<T> {
...
-    pub fn write(&mut self, val: T) {
-        // NOTE(volatile) we have observed UB when this uses a plain `ptr::write`
-        unsafe { ptr::write_volatile(&mut self.value, Some(val)) }
+    #[cfg(not(feature = "nightly"))]
+    pub fn write(&mut self, value: T) -> &mut T {
+        let ptr = self.inner.as_mut_ptr();
+        unsafe {
+            ptr::write_volatile(ptr, value);
+            &mut *ptr
+        }
```
TeXitoi: I'd rewrite this more like that:
```rust
pub fn write(&mut self, value: T) -> &mut T {
    self.inner = core::mem::MaybeUninit::new(value);
    unsafe { self.get_mut() }
}
```

src/export.rs
```diff
@@ -86,61 +84,28 @@ impl<T> MaybeUninit<T> {
...
     }
 
+    #[cfg(not(feature = "nightly"))]
     pub unsafe fn get_mut(&mut self) -> &mut T {
-        if let Some(x) = self.value.as_mut() {
-            x
-        } else {
-            unreachable!(MSG)
-        }
+        &mut *self.inner.as_mut_ptr()
```
TeXitoi: As this is the straightforward implementation, maybe we can remove the cfg and implementing it even in nightly?

src/export.rs
korken89: Thanks for the input! I will update this

src/export.rs
korken89: This we can do, I will remove the !nightly flag

korken89:
No problem that you don't have time, I need to work on stuff like this to learn the code base as well :)

src/export.rs
```diff
@@ -64,13 +62,11 @@ impl Priority {
     }
 }
 
-#[cfg(feature = "nightly")]
```
japaric: This change bumps the MSRV from 1.31.0 to 1.35.0 (or 36) so it cannot be done in the v0.4.x line -- it's a breaking change.

korken89: Agreed, but the current implementation is UB - then the only choice is to yank all of 0.4.x IMO :/

japaric: There's another choice: add an opt-in `core-maybeuninit` Cargo feature that switches the uses of MaybeUninit to the `core` version but also bumps the MSRV to 1.35 (or 36, I forget). The panic message can then be changed to point to that feature, which works on a recent stable release, instead of `nightly`. This option is not a breaking change.

korken89: Hmm, this we probably can do.
It seems strange though to allow this to exist for stable as we are clearly hitting UB - all of `0.4.x` should be yanked the second we release `0.5.0` or we can do another release of `0.4.x` and remove the stable support so people can continue to run `0.4.x` safely.

I think these are our only 2 options. Thoughts @japaric ?

korken89:
We will most likely create a fix for this soon, stay tuned

korken89:
Accepted on the meeting 2019-10-23

PR #230, opened 2019-08-31
Bump cortex-m and cortex-m-rt

nickray:
I'm not sure if there's a reason to stick with `cortex-m` 0.5?
It would be nice to depend on `0.x` versions to not have to bump this crate whenever improvements are made in `y` of `0.x.y`.
My use case is to enable LPC55S6x.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,8 +40,8 @@ name = "types"
 required-features = ["timer-queue"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


nickray:
I'm not sure what this build error is. Would work to bump at tip of `heterogeneous` too :)

korken89:
Hi, this is available in the latest master, reopen if you feel something is missing :)

PR #237, opened 2019-09-15
fix index.html redirect

japaric:


japaric:
It works without this.

PR #239, opened 2019-09-15
fix redirects and CNAME

japaric:


japaric:
skipping CI because this is just the docs deployment

PR #240, opened 2019-09-15
faster doc builds and fix redirect on 0.4

japaric:


japaric:
skipping CI because docs deployment

PR #267, opened 2019-11-06
fix CI: use an older nightly on the x86_64 build

japaric:


jonas-schievink:
(for a more permanent solution that doesn't require nightly, you could also check out https://github.com/dtolnay/trybuild)

PR #268, opened 2019-11-07
CI: replace compiletest-rs with trybuild

japaric:
We use compiletest to run compile-fail tests but compiletest depends on compiler
internals so it breaks every now and then and requires nightly. With trybuild we
can also run compile-fail tests but it works on stable and it already has
reached version 1.0

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -37,8 +37,11 @@ main() {
     mkdir -p ci/builds
 
     if [ $T = x86_64-unknown-linux-gnu ]; then
-        # compile-fail tests
-        cargo test --test single --target $T
+        if [ $TRAVIS_RUST_VERSION == 1.*.* ]; then
+            # test on a fixed version (MSRV) to avoid problems with changes in rustc diagnostics
+            # compile-fail tests
+            cargo test --test single --target $T
+        fi
 
         if [ $TRAVIS_RUST_VERSION = nightly ]; then
             # multi-core compile-pass tests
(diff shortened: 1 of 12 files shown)


ui/single/task-priority-too-high.stderr
```diff
@@ -1,9 +1,13 @@
+warning: unused import: `rtfm::app`
+ --> $DIR/task-priority-too-high.rs:3:5
+  |
+3 | use rtfm::app;
```
jonas-schievink: Uh, spurious warning? Is this a known rustc bug?

ci/script.sh
```diff
@@ -37,10 +37,10 @@ main() {
     mkdir -p ci/builds
 
     if [ $T = x86_64-unknown-linux-gnu ]; then
-        if [ $TRAVIS_RUST_VERSION = nightly ]; then
-            # compile-fail tests
-            cargo test --test single --target $T
+        # compile-fail tests
+        cargo test --test single --target $T
```
jonas-schievink: The compiler output might change between versions, so running this on all channels might cause issues later on

japaric: good point. I have changed the condition to test only when `rustc --version == $MSRV`. $MSRV remains constant for a whole major version of rtfm (e.g. 1.36.0 for v0.5.x)

ui/single/task-priority-too-high.stderr
japaric: The warning it's also present on 1.36.0. Perhaps this has been there for a while but compiletest ignores warnings by default?

jonas-schievink: Yeah, that's probably what's happening

PR #277, opened 2019-11-18
TimerQueue.dequeue: don't set SYST reload to 0

mpasternacki:
ARM Architecture Reference Manual says: "Setting SYST_RVR to zero has the effect of disabling the SysTick counter independently of the counter enable bit."

If Monotonic's ratio is less than one, the timeout calculations
can compute zero if next task is scheduled after current instant, but
before next timer tick. This results in disabling SYST and freezing the
timer queue.

The division by ratio's denominator rounds downward and the dequeue
condition is `if instant < now`. If ratio is small enough, this results
in unnecessary interrupts:

Let's say `instant - now` is 99 and ratio is 1/25. Then, `dur` will
equal 3 and the next tick will happen at `now + 75`. In the next
interrupt, `instant > now` and additional tick needs to be scheduled
(which doesn't happen, because now `instant - now` is less than 25, so
reload will be set to 0 and timer queue will stop). Adding one to
computed duration will prevent both freezing and additional interrupts.

When ratio is 1 or close, timer queue code overhead will prevent this
from happening. I am working with a chip where CPU is clocked at 600MHz
and SysTick is 100kHz and the freeze happens quite often.

### DIFF
--- a/src/tq.rs
+++ b/src/tq.rs
@@ -68,6 +68,13 @@ where
                             .map(|x| x / ratio.denominator)
                     }) {
                         None => MAX,
+
+                        // ARM Architecture Reference Manual says:
+                        // "Setting SYST_RVR to zero has the effect of
+                        // disabling the SysTick counter independently
+                        // of the counter enable bit."
+                        Some(0) => 1,
+
                         Some(x) => cmp::min(MAX, x),
                     };
                     mem::transmute::<_, SYST>(()).set_reload(dur);


mpasternacki:
By the way, is there any legitimate scenario where ratio is not fixed and known in build time? If instead of a method ratio was expressed as an associated typenum, it would let the compiler generate tighter code for the critical path and/or allow simpler implementations for common situations (e.g. ratio=1, numerator=1, denominator=1, large/small ratio).

src/tq.rs
```diff
@@ -68,7 +68,7 @@ where
                             .map(|x| x / ratio.denominator)
                     }) {
                         None => MAX,
-                        Some(x) => cmp::min(MAX, x),
+                        Some(x) => cmp::min(MAX, x + 1),
```
japaric: Could you add a comment in code explaining why we don't want to use the value `0`? Citing the reference manual.

Also I think it would be better to min-max clamp this instead of modifying the computed value, as in use `cmp::min(MAX, cmp::max(x, 1))`.

korken89:
```suggestion
                        Some(x) => cmp::min(MAX, cmp::max(x, 1)),
```

japaric:
> By the way, is there any legitimate scenario where ratio is not fixed and known in build time?

No, runtime variable `ratio` is not supported as in it's error prone to use because it will cause previously scheduled tasks to run at different times (not the ones specified by the user).

> it would let the compiler generate tighter code for the critical path and/or allow simpler implementations for common situations (e.g. ratio=1, numerator=1, denominator=1, large/small ratio).

What optimizations do you have in mind? I expect LLVM const propagator to optimize away any unnecessary operation with the current code. It would also be fine to add conditionals on the `ratio` here; LLVM will optimize away the unused branches so only one path is taken at runtime and no condition is evaluated (`e.g. ratio == 1`).

korken89:
After this lands I propose that we yank 0.5.0 and release 0.5.1, anyone against @rtfm-rs/devs ?

src/tq.rs
mpasternacki: Clamping with `min` was my initial idea, but I realized that this way the interrupt is scheduled one tick too early, see example with `instant - now == 99` and ratio 1/25 in the fourth paragraph.

Unless I misunderstand something, in this situation the reload value would be set to 3 and the handler would run 75 `Instant` ticks later, still 24 ticks too early. In the next call, `instant > now`, `(instant - now) / ratio.denominator` is 0 and we reschedule again.

mpasternacki: I just noticed an off-by-one in my own calculations above: after the `reset_current()`, there's one more tick from 0 to reload value that I didn't include.

Confirmed that you're right by adding a global counter increased from `dequeue` and printed after ~10 seconds. The count with `max(x, 1)` is a _little_ bit smaller than counter with `x+1`, but the difference is too small to be caused by the effect I was suggesting.

korken89:
bors try

PR #278, opened 2019-11-18
Cyccnt

perlindgren:
The subtractions in `elapsed` and `duration` may cause an overflow panic in debug mode. This is solved by using wrapping arithmetics.


japaric:
bors r=texitoi

PR #294, opened 2020-01-21
use the safe DWT::unlock API

japaric:
instead of a unsafe write_volatile call


japaric:
For some reasons, no PRs show up in the bors queue. Let's see if closing and re-opening the PR helps.

japaric:
Now, this PR shows up in the bors queue but it needs to be approved and then `bors r+`-ed. cc @korken89

PR #302, opened 2020-03-07
Print module name and priority

AfoHT:


korken89:
Thanks for the PR! Could you provide a description of what it does and why it was needed?

perlindgren:
This is just an experimental branch, so we can merge directly.

PR #304, opened 2020-04-07
Moved to new rtfm-core with mutex trait and common parts

korken89:


korken89:
When https://github.com/rtfm-rs/rtfm-core/pull/10 is merged and released we can merge here.

PR #306, opened 2020-04-14
Retain cfg-attributes on resources

### LINKED ISSUE #301: #[cfg]'ing away resources
It seems its currently not possible to `#[cfg]` away resources, though with some investigation it seems its actually pretty close to working.

Minimal example:
```rust
const APP: () = {
    struct Resources {
        // A resource
        #[init(0)]
        shared: u32,
        
        #[cfg(feature = "feature_x")]
        x: u32,

        dummy: (),
    }

    #[init]
    fn init(_: init::Context) -> init::LateResources {
        rtfm::pend(Interrupt::UART0);
        rtfm::pend(Interrupt::UART1);

        init::LateResources {
            #[cfg(feature = "feature_x")]
            x: 0,
            dummy: () // dummy such that we have at least one late resource
        }
    }

    // `shared` can be accessed from this context
    #[task(binds = UART0, resources = [#[cfg(feature = "feature_x")] x, shared])]
    fn uart0(cx: uart0::Context) {
        let shared: &mut u32 = cx.resources.shared;
        *shared += 1;

        hprintln!("UART0: shared = {}", shared).unwrap();

        #[cfg(feature = "feature_x")]
        {
            hprintln!("UART0: X = {}", shared).unwrap();
        }
    }
};
```

this fails to compile with:

```rust
error[E0425]: cannot find value `x` in this scope
  --> src/main.rs:22:9
   |
22 |         x: u32,
   |         ^ not found in this scope
error: aborting due to previous error
For more information about this error, try `rustc --explain E0425`.
error: could not compile `rtfm-cfg`.
To learn more, run the command again with --verbose.
```

Looking at the expanded output, there are a few key points:

1. The `x` field is correctly `cfg`'d away _within_ the `init` function:
```rust
fn init(_: init::Context) -> init::LateResources {
    rtfm::pend(Interrupt::UART0);
    rtfm::pend(Interrupt::UART1);
    init::LateResources { dummy: () }
}
```
I'm guessing this is expected, as that cfg comes from user code, not from the proc_macro

2. The `LateResources` struct does **not** keep/apply the cfg:
```rust
/// Resources initialized at runtime
#[allow(non_snake_case)]
pub struct initLateResources {
    pub dummy: (),
    pub x: u32,
}
```

3. Doesn't seem to be a way to cfg away resources in tasks declarations:
```rust
#[task(binds = UART0, resources = [#[cfg(feature = "feature_x")] x, shared])] 
fn uart0() {
// .. the cfg attr is ignored
```
```rust
#[cfg_attr(feature = "shared", task(binds = UART0, resources = [shared]))]
#[cfg_attr(not(feature = "shared", task(binds = UART0)))]
fn uart0() {
// .. complains that uart0 must be declared outside the macro
```

The full project example can be found here: https://github.com/MabezDev/rtfm-cfg-experiments

I've also done a little digging inside the `rtfm-syntax` crate, it _seems_ to me the parse function does correctly keep the cfg's (printing `app` inside `parse2`) so I'm unsure where that information is lost in the chain of crates/functions. I think it probably needs someone with more proc_macro experience than me to look at it.

AfoHT:
When rust 1.43 lands as stable this will resolve #301 and allow for the kind of conditional compilation exemplified in the issue.

Tested on beta and nightly.

### DIFF
--- a/examples/t-cfg-resources.rs
+++ b/examples/t-cfg-resources.rs
@@ -1,0 +1,40 @@
+//! [compile-pass] check that `#[cfg]` attributes applied on resources work
+//!
+#![no_main]
+#![no_std]
+
+use panic_halt as _;
+
+#[cfg(rustc_is_nightly)]
+mod example {
+
+    #[rtfm::app(device = lm3s6965)]
+    const APP: () = {
+        struct Resources {
+            // A resource
+            #[init(0)]
+            shared: u32,
+
+            // A conditionally compiled resource behind feature_x
+            #[cfg(feature = "feature_x")]
+            x: u32,
+
+            dummy: (),
+        }
+
+        #[init]
+        fn init(_: init::Context) -> init::LateResources {
+            init::LateResources {
+                // The feature needs to be applied everywhere x is defined or used
+                #[cfg(feature = "feature_x")]
+                x: 0,
+                dummy: () // dummy such that we have at least one late resource
+            }
+        }
+
+        #[idle]
+        fn idle(_cx: idle::Context) -> ! {
+            loop {}
+        }
+    };
+}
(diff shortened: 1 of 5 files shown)


korken89:
Lets wait for #307 to merge, then we can merge here.

AfoHT:
Before merging I'd like to rebase and append the example to `ci/script.sh`

PR #307, opened 2020-04-20
Use build.rs for conditional compilation

AfoHT:
Extend the current test suite to allow for running tests on newer rustc-versions than current MSRV.

Required by #306 to add special tests for future MSRV.

To exclude an example from the regular non-nightly testing:

```
#![no_main]
#![no_std]

#[cfg(rustc_is_nightly)]
mod example {

    use panic_halt as _;

    const APP: () = {
    <more code>
    }
}
```


korken89:
bors retry

PR #312, opened 2020-05-21
Use embedded-time for Clock/Instant/Duration implementation

### LINKED ISSUE #309: Monotonic Implementation Difficulties
I am working on a unified time-type library for embedded applications [embedded-time]( https://www.github.com/PTaylor-FluenTech/embedded-time). There is one considerable pain point trying to make it compatible with the Monotonic trait for use with RTFM. That is the requirement that the Duration must implement a special TryInto<u32> to convert into the System Timer ticks. This use of TryInto<u32> feels inappropriate.

What I would like to suggest is instead have the System Timer implement some trait (let's call it `Clock`). This `Clock` trait would require a Fraction/Ratio indicating the precision of the clock. In that way a TryInto<impl Clock> could be implemented to convert a duration to a different precision.

This would be roughly in line with how time is handled in the std::chrono library in C++.

I'm going to look into creating an example implementation of this structure, but any suggestions/comments would be much appreciated.

PTaylor-us:
This PR is an example implementation using the [`embedded-time`](https://crates.io/crates/embedded-time) crate. It will allow usage of _human_ time units.

For example, it replaces the `cycles` with _real_ time units:
```rust
    #[task(resources = [led1], schedule = [turn_off_led1])]
    fn turn_on_led1(cx: turn_on_led1::Context) {
        let led1 = cx.resources.led1;
        led1.set_low().unwrap();

        cx.schedule
            .turn_off_led1(cx.scheduled + 250.milliseconds())
            .unwrap();
    }
```

# embedded-time
`embedded-time` provides a comprehensive library for implementing abstractions over
hardware and work with _instants_ and _durations_ in an intuitive way.
 
- `Clock` trait allowing abstraction of hardware timers/counters for timekeeping.
- Work with time using _milliseconds_, _seconds_, etc. rather than _cycles_ or _ticks_.
- Includes examples for the nRF52_DK development kit board.

Fixes #309 #219 

# Todo
- [x] Add SysTimer/SysTick clock speed parameter
- [x] Make `sys_timer_freq` optional (only required when using scheduling)


korken89:
Hi, thanks for the work!

One quick thought, I'd change `sys_timer_freq` to `monotonic_freq` to hightlight the connection to monotonic.
I will have a deeper look into this and take it up in the next meeting as well.

PTaylor-FluenTech:
> One quick thought, I'd change `sys_timer_freq` to `monotonic_freq` to hightlight the connection to monotonic.

The sys_timer_freq parameter is to declare the frequency of the SysTick timer. The selected monotonic clock/timer frequency is declared independently.

PTaylor-FluenTech:
I know on the nRF52, the SYST_CALIB register is not used (reads `0`). I'm not sure how often the use of that register is implemented by chip manufacturers, but it would make a perfect alternative to needing to define `sys_timer_freq`. That way, that manual definition would only be required if using scheduling AND the SYST_CALIB is `0`.

korken89:
Hi, thanks for the PR!

We just did the rename to RTIC - could you rebase from master?
I think this is very interesting and will add it to the agenda of next meeting.

PTaylor-FluenTech:
I've rebased to master as well as did some much-needed cleanup to my PR branch. I look forward to any feedback.

korken89:
Hi, we talked about this PR at yesterdays meeting and @perlindgren and I will do a review.
The overall view of the PR is positive and would lay foundations for further developments.

Could you look into the CI errors and upgrade to the latest `embedded-time`?

PTaylor-FluenTech:
> Could you look into the CI errors and upgrade to the latest `embedded-time`?
@korken89 Absolutely. I believe many of the CI errors are due to the current embedded-time crate requiring nightly. I'm currently using `associated_type_bounds` and `const_fn` feature flags. I believe I can drop the `const_fn` pretty easily, but the other may be difficult. If you know of a crate that would replace the need for that flag, please let me know.

PTaylor-FluenTech:
I have updated the `embedded-time` version to the latest. It is now buildable with `stable`. I believe the two currently-failing tests are not actually failures.

PTaylor-us:
I'm investigating some important changes based on received feedback.

korken89:
Thanks  for an update!

Sorry for our delay, we have not yet had time for a proper review yet.

PTaylor-us:
One of the recent changes to `embedded-time` has been the change of the `Clock` trait methods to take &self, allowing impls to be stateful. I am unsure what the best way might be to incorporate this change into RTIC as the current framework only supports stateless / static (I may be missing the right terminology here) monotonic clocks. Perhaps using a stateless trait impl (perhaps the `Monotonic` trait could be used for this) that reads from a `static` `Option<impl embedded_time::Clock>`?

korken89:
Hi, sorry for letting this go so slow - I'm currently swamped.

Could we look into in this again the coming weeks?

mattico:
> One of the recent changes to embedded-time has been the change of the Clock trait methods to take &self, allowing impls to be stateful. I am unsure what the best way might be to incorporate this change into RTIC as the current framework only supports stateless / static (I may be missing the right terminology here) monotonic clocks. Perhaps using a stateless trait impl (perhaps the Monotonic trait could be used for this) that reads from a static Option<impl embedded_time::Clock>?

Brainstorming a few solutions:

### 1. Keep requiring the `Monotonic` impl to be stateless. 

Don't change the `Monotonic` trait. If you have an `impl Clock` you can stash it in a static then impl `Monotonic` on a unit struct that accesses that static. If you're implementing it yourself you can ignore `Clock` altogether. Users can call e.g. `CycCnt::now()` to easily access the current time, since we no longer have the rtic `Instant` type with its `now()` method.

### 2. Allow the `Monotonic` impl to be stateful, and put it in a static.

The user initializes the `Monotonic` impl in `init()` and passes it to RTIC somehow.

```rust
fn init(cx: init::Context) {
    // RTIC stashes the clock in a static or somewhere
    cx.clock = Some(CycCnt());
    // OR
    rtic::init_clock(CycCnt());
}

// OR

static mut CLOCK: Option<Clock> = None;

#[rtic::app(monotonic = CLOCK)]
const APP: () = {
    #[init]
    fn init(_) {
        unsafe { CLOCK = Some(...) }
    }
}
```

### 3. Allow the `Monotonic` impl to be stateful, and put it in resources

```rust
#[rtic::app()]
const APP: () = {
    struct Resources {
        // Stateless clocks could be initialized using the #[init] attribute
        #[clock]
        #[init(rtic::CycCnt())]
        clock: CycCnt,
    }

    #[init]
    fn init(_) -> init::LateResources {
        // My custom clock using a hardware timer and an overflow counter
        let us_timer = UsTimer::new(dp.TIM2, ccdr.peripheral.TIM2, &ccdr.clocks);

        init::LateResources {
            clock: us_timer,
        }
    }

    // Now I can store the counter in the UsTimer struct rather than use a static
    #[task(binds = TIM2, priority = 8, resources = [clock])]
    fn tim2(cx: tim2::Context) {
        cx.resources.clock.handle_irq();
    }

    // But tasks would have to pull the clock impl out of resources and pass it to any code that needs it
    #[idle]
    fn idle(cx: idle::Context, resources = [&clock])
        sensor::read(&cx.resources.clock);
    }

    // We could have RTIC provide the clock in Context to simplify a bit
    #[idle]
    fn idle(cx: idle::Context)
        sensor::read(cx.clock);
    }
}
```

I like option 3 a bit. It's quite elegant, it makes it easy to create stateful `Clock`s and it makes it clear when the `clock` is initialized so you don't end up trying to use it in init() before it exists. However I really don't like having to pass the clock around to everything that wants it and I don't like having to call `clock.try_now().unwrap()`. I have logging macros that access the timer for a timestamp, do I need a separate timer for that, or a separate way of accessing the timer object, or to pass &clock to every log invocation? I'd rather make implementing the timer a bit more annoying in exchange for making it easy to use the timer.

perlindgren:
We will have to take a closer look. (@korken89  and I will sit down tomorrow for a bit, hopefully we will have time to look into this as well). (Sorry for taking such a long time...)

korken89:
Hi, for the current take at this loop at the new monotonic PR: #436

PR #313, opened 2020-05-26
Update example to use better initial value

russell:
The example above this in the documentation states

```
        // semantically, the monotonic timer is frozen at time "zero" during `init`
        // NOTE do *not* call `Instant::now` in this context; it will return a nonsense value
        let now = cx.start; // the start time of the system
```

It results in weird scheduling issues, but still eventually works.  `cx.start` is much more reliable.

Relates to https://github.com/rtfm-rs/cortex-m-rtfm/issues/196


korken89:
Thanks for fixing!

PR #315, opened 2020-05-29
allow handlers to be named 'main'

### LINKED ISSUE #311: naming the `#[idle]` function "main" produces a compiler error
## Steps to reproduce:

Modify the `examples/idle.rs` in this repository like this:

``` diff
     }

     #[idle]
-    fn idle(_: idle::Context) -> ! {
+    fn main(_: main::Context) -> ! {
         static mut X: u32 = 0;

         // Safe access to local `static mut` variable
```

Now the example fails to build:

``` console
$ cargo b --example idle --target thumbv7m-none-eabi
error[E0061]: this function takes 0 arguments but 2 arguments were supplied
  --> examples/idle.rs:19:8
   |
11 | #[rtfm::app(device = lm3s6965)]
   | -------------------------------
   | |
   | supplied 2 arguments
   | defined here
...
19 |     fn main(_: main::Context) -> ! {
   |        ^^^^ expected 0 arguments

error: aborting due to previous error
```

## The problem

If you look into the `rtfm-expansion.rs` file you'll see that the proc-macro produces a `#[no_mangle]` function called "main"; this name collides with the user provided one.

``` rust
fn main(main::Locals { X, .. }: main::Locals, _: main::Context) -> ! {
    // user code
}

const APP: () = {
    use lm3s6965 as _;
    #[no_mangle]
    unsafe extern "C" fn main() -> ! {
        // ..
        main(
            main::Locals::new(),
            main::Context::new(&rtfm::export::Priority::new(0)),
        )
    }
};
```

## The fix

Modify the `main` function inside `const APP` to have a random identifier (it won't be called by user code or proc-macro generated code) and use `export_name` instead of `no_mangle`. So, like this:

``` rust
const APP: () = {
    use lm3s6965 as _;
    #[export_name = "main"]
    unsafe extern "C" fn __rtfm_0() -> ! {
        // ..
        main(
            main::Locals::new(),
            main::Context::new(&rtfm::export::Priority::new(0)),
        )
    }
};
```

Or change the inner `main` function call to use an absolute path: `crate::main(/* args */)`. That may also work.

japaric:
`#[init]`, `#[idle]` and `#[task]` handlers can now be named `main`

fixes #311

### DIFF
--- a/examples/t-htask-main.rs
+++ b/examples/t-htask-main.rs
@@ -1,0 +1,20 @@
+#![deny(unsafe_code)]
+#![deny(warnings)]
+#![no_main]
+#![no_std]
+
+use cortex_m_semihosting::debug;
+use panic_semihosting as _;
+
+#[rtfm::app(device = lm3s6965)]
+const APP: () = {
+    #[init]
+    fn init(_: init::Context) {
+        rtfm::pend(lm3s6965::Interrupt::UART0)
+    }
+
+    #[task(binds = UART0)]
+    fn main(_: main::Context) {
+        debug::exit(debug::EXIT_SUCCESS);
+    }
+};
--- a/examples/t-idle-main.rs
+++ b/examples/t-idle-main.rs
@@ -1,0 +1,20 @@
+#![deny(unsafe_code)]
+#![deny(warnings)]
+#![no_main]
+#![no_std]
+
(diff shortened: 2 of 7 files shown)


korken89:
Hi, some tests are failing. Can you have a look?

korken89:
bors retry

PR #319, opened 2020-06-10
Use cargo feature instead of conditional compilation hacks

AfoHT:
As noted by japaric, using cargo features to do selective testing is cleaner than cfg-hacks.

Introduce new testing-feature indicating this test requires at minimum Rust v1.43.

Modify both test-suites to handle the new feature and remove the cfg-hack from the example.


korken89:
mdbook failed, merging manually

PR #326, opened 2020-06-11
Changed reference of RTFM lang

trembel:
The reference to the RTFM language has been corrected, as it was referenced to RTIC language (which does not exist)


trembel:
Will be fixed by https://github.com/rtic-rs/cortex-m-rtic/pull/327

PR #333, opened 2020-06-18
Increase MSRV to 1.41.0, as this was when profile-overrides landed

richardeoin:
profile-overrides landed in 1.41 https://github.com/rust-lang/cargo/pull/7591/

I appreciate that the intention may be to maintain support for earlier rust versions if the profile-overrides keys are deleted from the README. But if the crate throws an error when compiled with its MSRV compiler, that in my opinion/understanding defeats the purpose of the MSRV.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,7 +45,7 @@ jobs:
           - x86_64-unknown-linux-gnu
         toolchain:
           - stable
-          - 1.36.0
+          - 1.41.0
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


richardeoin:
The CI fails (as far as I can tell) on new compiler diagnostics in rust 1.41 

> (which \`$CRATE\` depends on)

AfoHT:
For the upcoming version 0.6 the current plan is to move towards a moving "MSRV" which follows Rust stable.

But for 0.5.x MSRV will remain stuck in the past. This is the reason such hacks were done.

korken89:
This has been fixed in master

PR #338, opened 2020-06-30
Fixes an issue where one could double take the cortex_m Peripheral

### LINKED ISSUE #321: `cortex_m::Peripherals` singleton can be duplicated
### STR

Modify the `init` example in this repo as follows:

``` rust
const APP: () = {
    #[init]
    fn init(cx: init::Context) {
        // Cortex-M peripherals
        let core: cortex_m::Peripherals = cx.core;

        // Cortex-M peripherals again
        let core2 = cortex_m::Peripherals::take().unwrap();
        // ^ this doesn't panic but should

        debug::exit(debug::EXIT_SUCCESS);
    }
};
```

This means that applications cannot rely on e.g. `drop(core.ITM)` to seal the configuration of the ITM peripheral (because there's another ITM handle that can still access the peripheral).

The code that runs before `init` is called should call `cortex_m::Peripherals::take` so that future calls to that method return `None`.

cc @korken89 @perlindgren

korken89:
Closes #321

### DIFF
--- a/macros/src/codegen/pre_init.rs
+++ b/macros/src/codegen/pre_init.rs
@@ -43,6 +43,13 @@ pub fn codegen(
         let mut core: rtic::export::Peripherals = core::mem::transmute(());
     ));
 
+    if app.args.cores == 1 {
+        stmts.push(quote!(
+            // To set the variable in cortex_m so the peripherals cannot be taken multiple times
+            let _ = cortex_m::Peripherals::steal();
+        ));
+    }
+
     let device = extra.device;
     let nvic_prio_bits = quote!(#device::NVIC_PRIO_BITS);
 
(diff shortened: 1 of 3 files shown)


korken89:
bors r=japaric

macros/src/codegen/pre_init.rs
```diff
@@ -43,6 +43,13 @@ pub fn codegen(
         let mut core: rtic::export::Peripherals = core::mem::transmute(());
     ));
 
+    if app.args.cores == 1 {
+        stmts.push(quote!(
+            // To set the variable in cortex_m so the peripherals cannot be taken multiple times
+            let _ = cortex_m::Peripherals::steal();
```
jonas-schievink, line 49: This caused https://github.com/rtic-rs/cortex-m-rtic/issues/351, it should probably read something like:

`let _ = rtic::export::Peripherals::steal();`

korken89: This looks correct, I'll make a fix

PR #340, opened 2020-07-09
 rtic::export::Peripherals created by into instead of transmute

perlindgren:


AfoHT:
Looks good to me :+1:

PR #341, opened 2020-07-13
Enable caching for Github Actions

AfoHT:
Using [GHA caching](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows) to store key parts of the testing environment.

One example is downloading `arm-none-eabi-gcc` (which takes roughly 1 minute) for each checkexamples.

The rustup setup is roughly 200MB and restores nicely.

Rust examples can be found [here](https://github.com/actions/cache/blob/master/examples.md#rust---cargo)

**Something to discuss:**

Several notable projects remove some problematic files in order to keep cache size reasonable

[rust-analyzer](https://github.com/rust-analyzer/rust-analyzer/blob/9a7db8fa009c612168ef16f6ed72315b5406ed09/.travis.yml#L2-L4)

[cargo duscussion](https://github.com/rust-lang/cargo/issues/5885)

[tantivity-search](https://github.com/tantivy-search/tantivy/pull/531/files)

[clap-rs](https://github.com/clap-rs/clap/pull/1658/files#diff-354f30a63fb0907d4ad57269548329e3R5-R16)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -152,15 +152,6 @@ jobs:
           restore-keys: |
             ${{ runner.OS }}-qemu
 
-      - name: Cache arm-none-eabi-gcc
-        id: cache-arm-none-eabi-gcc
-        uses: actions/cache@v2
-        with:
-          path: arm-none-eabi-gcc
-          key: ${{ runner.OS }}-gcc
-          restore-keys: |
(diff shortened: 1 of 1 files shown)


korken89 (approved):
bors merge

PR #372, opened 2020-10-01
Now late resources are always used

korken89:
Awaits the PR in `rtic-syntax`: https://github.com/rtic-rs/rtic-syntax/pull/32


AfoHT (approved):
bors merge

PR #375, opened 2020-10-02
close console text blocks on a new line

### LINKED ISSUE #369: Book include formatting is.. weird.
In the book, all the code and console examples follow a pattern like so:
     \``` rust
     {{#include ../../../../examples/resource.rs}}
     \```
     \``` console
     $ cargo run --example resource
     {{#include ../../../../ci/expected/resource.run}}```

Note that the console block is terminated with the backticks at the end of the include line, not on a new line.  This breaks the rendering, both for md previews (e.g. in vscode or github's own browsing) and in the output from `mdbook serve`.  All the following md source up to the next set of backticks gets put in as console text.

I don't understand why, but it's consistent and clearly deliberate for some reason. My guess was something related to suppressing an extra newline in the run output, perhaps related to a difference in Windows vs Unix line endings?

This has been this way for a long time. It obviously isn't coming out this way in the process that rendered the pages on rtic.rs. Maybe that's not broken because it was rendered on Windows with different line endings? Maybe it's a change in more recent `mdbook`?

If anyone knows more, please comment. If the reason is still valid, then just close this issue, otherwise I'll make a PR to add in the extra line endings and fix it that way if it's no longer needed, or noone knows why anymore.

dcarosone:
fixes #369

### DIFF
--- a/book/en/src/by-example/app.md
+++ b/book/en/src/by-example/app.md
@@ -55,7 +55,8 @@ process.
 
 ```  console
 $ cargo run --example init
-{{#include ../../../../ci/expected/init.run}}```
+{{#include ../../../../ci/expected/init.run}}
+```
 
 ## `idle`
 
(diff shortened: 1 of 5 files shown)


AfoHT (approved):
bors merge

PR #381, opened 2020-10-05
Separate example check and run-pass tests

AfoHT:
With the old 1.36.0 tests removed the full 20 free concurrent GHA jobs are not fully utilized.

Separate the run-pass testing and the cargo check of examples


korken89 (approved):

korken89:
bors merge

PR #390, opened 2020-10-11
Spawn and schedule from anywhere

korken89:
This PR moves RTIC to the spawn and schedule from anywhere syntax.

Notable changes:
 * We do no longer support non-`Send` types.
 * Some extra code is generated as any task may spawn/schedule any task. However Rust/LLVM does a great job optimizing away non used instantiations (no real code-size difference observed).
 * Worst case priority inversion has increased, but it is now predictable.

Upsides:
 * With this we should be able to support async/await.
 * RTIC tasks can now be callbacks (spawned and scheduled).
 * RTIC tasks can be stored.

Needs the following PR to land first: https://github.com/rtic-rs/rtic-syntax/pull/34

The following now works:

```rust
#[rtic::app(device = lm3s6965, monotonic = rtic::cyccnt::CYCCNT)]
mod app {
    #[init]
    fn init(mut cx: init::Context) -> init::LateResources {
        // Init stuff...

        // New spawn syntax
        foo::spawn().unwrap();

        // New schedule syntax
        bar::schedule(now + 4_000_000.cycles()).unwrap();

        init::LateResources {}
    }

    #[task]
    fn foo(_: foo::Context) {}

    #[task]
    fn bar(_: bar::Context) {}

    extern "C" {
        fn SSI0();
    }
}

```


korken89:
Look into namespace pollution before merging

AfoHT (approved):
bors merge

PR #395, opened 2020-10-15
Made relation between priority and number explicit

diondokter:
When quickly reading through the priorities chapter, I couldn't find in which order the priorities were, so I assumed it was the same as in the hardware.
In the cortex-m hardware, interrupts with the **lower** priority number will preempt the other interrupts.
RTIC does the reverse, so I think it's good to be more explicit about it.


diondokter:
So, the CI failed, but I don't think that's my fault...

korken89 (approved):
bors merge

PR #416, opened 2020-11-15
Move entry-point main into a separate module

AfoHT:
Prevents conflict with user provided tasks named main

Curious what the full test suite will make of this


korken89 (approved):
bors merge

PR #450, opened 2021-03-02
Cleanup of tests, solve duplicate panic handler error

AfoHT:
Currently `panic-halt` and `panic-semihosting` collides during testing, generating this version dependent output in the output:

```
 error: duplicate lang item in crate `panic_halt` (which `$CRATE` depends on): `panic_impl`.
Error:   |
  = note: the lang item is first defined in crate `std` (which `$CRATE` depends on)
  = note: first definition in `std` loaded from /usr/share/rust/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-e12de7683a34c500.rlib
  = note: second definition in `panic_halt` loaded from $DIR/target/tests/target/x86_64-unknown-linux-gnu/debug/deps/libpanic_halt-f6dc7d4a3dd7a21d.rmeta
  ```
  
What is worse, the checksum changes with each new release of Rust causing testv7 and testv6 to break. Additionally, the path is the path found in the GHA build environment, making it overly cumbersome to use the `TRYBUILD=overwrite` variable because you most likely do not have your rust installation in `/usr/share`.

By removing `panic-halt` this conflict goes away.

Since there are no multi-core tests "single" in the path makes less sense. Rename to "tests".


perlindgren (commented):
Looks good, but not sure where the default panic handler comes from. Should that not be included instead?

AfoHT:
> Looks good, but not sure where the default panic handler comes from. Should that not be included instead?

Should have marked that initial work as a draft, wanted a go with the CI.

Should be updated and targeting the right branch now, to reduce merge-chaos for later.

PR #453, opened 2021-03-03
Fix GHA by removing duplicate CI-check

AfoHT:
Same as done in https://github.com/rtic-rs/cortex-m-rtic/commit/fa2de97d870e0476cd7a0107193ae426ef8c3997


korken89 (approved):
bors merge

PR #454, opened 2021-03-03
Add periodic GHA job to run cargo audit

AfoHT:
With the recent generic-array issue affecting heapless it seems wise to stay up to date with the latest advisories.


korken89 (approved):
bors merge

PR #456, opened 2021-03-11
Cancel/reschedule support for monotonics

korken89:
Design document: https://hackmd.io/lhUCzrKBS-66aadO4KsSzw?view


MabezDev:
This seems to work well for me (nothing broke :D), though I do not currently have a need for the cancel/reschedule support. 

I'm not well versed in macro speak to comment on the proc-macro implementation but I had wondered if it would be better to PR the linked list impl to `heapless` instead of it residing in this crate? I think `heapless` would benefit from such an implementation and this crate already depends on it for other data structures.

korken89:
@MabezDev Yeah, it would make sense to get it into heapless.
I'm not sure if I'll fix it during this PR though, the linked list needs a lot of documentation fixes for heapless that I do not currently have time/energy for. :)

perlindgren (dismissed):
ok, let's go

AfoHT (approved):
Bors merge

PR #465, opened 2021-04-05
update russian translation of the book

burrbull:


AfoHT:
Nice that there is interest to bring the translation of the book up to date, unfortunately I do not speak Russian so I cannot judge the content.

The current book as found on the master branch is for v0.6 but incomplete, newer changes such as monotonic etc has not been properly documented (since early stages, still alpha.1).

If I understand correctly the updated translation mirrors the current docs as available in latest master?

Then later on we'll need to do an update to the translation as well when final release of 0.6 comes around.

The tests fails due to monotonic, unrelated to the documentation changes

burrbull:
> If I understand correctly the updated translation mirrors the current docs as available in latest master?

Yes, you are right. I'll disable those pages temporary as it's done in master.

korken89:
If this is rebased on master the tests will pass again.

burrbull:
Done

korken89 (approved):
bors merge

PR #471, opened 2021-04-13
Force push to gh-pages branch

AfoHT:
As suggested in https://github.com/rtic-rs/rfcs/pull/48#issuecomment-815730654


korken89 (approved):
bors merge

PR #473, opened 2021-04-15
Add cargo bloat GHA action

AfoHT:


AfoHT:
Needs to revisit this, but since we get PRs from repositories outside of `rtic-rs` organisation this way is not available.

PR #476, opened 2021-04-18
reclaim stack space used in late init

### LINKED ISSUE #474: Stack in `init()` doesn't get released, potential copies of static resources
If I have a buffer that gets initialized in RTIC's `init()` and then stored statically (like in RTIC's `Resources`), there's a pretty good chance that at least 2x memory would be allocated for it.

I noticed this when I was increasing the size of a statically allocated buffer (stored in `Resources`), and I would see both my static memory and stack memory allocations increase by the amount.  Optimizations are turned on.

Consider this example, which stores a 1KB buffer and works fine (only 1KB of memory is allocated).

```rust
type MyBuffer = Vec<u8, U1024>;

#[rtic::app(device = rtic_demo_memory::hal::raw, peripherals = true, monotonic = rtic::cyccnt::CYCCNT)]
const APP: () = {

    struct Resources {
        mybuffer: MyBuffer,
    }
    #[init()]
    fn init(c: init::Context) -> init::LateResources {
        init::LateResources {
            mybuffer: MyBuffer::new()
        }
    }
// ...
```

I check the binary `.uninit` is ~1KB bytes.  And in gdb I double check there's no stack allocation after init.  I stop the program in the idle loop and look at the stack frames:

`main::APP::main`: 0x2003fbe8

`main::idle `: 0x2003fbe0

Only 8 byte difference, no problem.

But this changes if `init()` has any stack allocation.

```rust
type MyBuffer = Vec<u8, U1024>;

#[inline(never)]
fn init_buffer() -> MyBuffer {
    MyBuffer::new()
}

#[rtic::app(device = rtic_demo_memory::hal::raw, peripherals = true, monotonic = rtic::cyccnt::CYCCNT)]
const APP: () = {

    struct Resources {
        mybuffer: MyBuffer,
    }
    #[init()]
    fn init(c: init::Context) -> init::LateResources {
        let mybuffer = init_buffer();
        init::LateResources {
            mybuffer: mybuffer
        }
    }
// ...
```

Now there's 1KB statically allocated, *and* 1KB allocated in the stack before calling the idle loop.

`main::APP::main`: 0x2003fbe8

`main::idle `: 0x2003f7d8 (1040 byte difference)

For our actual security key application, we have a dozen resources allocated this way and have around 30KB bytes on the stack that aren't used (~10% of ram).

I think if RTIC can ensure that `init()` is called in a subroutine that doesn't get inlined, this should fix the issue.

conorpp:
Fixes #474.

Tested that there is no longer any stack overhead leftover from moving init resources.

(made mistake force pushing with last PR when trying to fix lint)

The expansion for an example with 2 buffers as resources changes from:

```rust
let (late, mut monotonics) = crate::APP::init(init::Context::new(core.into()));
__rtic_internal_mybuffer.as_mut_ptr().write(late.mybuffer);
__rtic_internal_mybuffer2.as_mut_ptr().write(late.mybuffer2);
rtic::export::interrupt::enable();
crate::APP::idle(idle::Context::new(&rtic::export::Priority::new(0)))
```

to:

```rust
#[inline(never)]
fn __rtic_init_resources<F>(f: F)
where
    F: FnOnce(),
{
    f();
}
__rtic_init_resources(|| {
    let (late, mut monotonics) = crate::APP::init(init::Context::new(core.into()));
    __rtic_internal_mybuffer.as_mut_ptr().write(late.mybuffer);
    __rtic_internal_mybuffer2.as_mut_ptr().write(late.mybuffer2);
    rtic::export::interrupt::enable();
});
crate::APP::idle(idle::Context::new(&rtic::export::Priority::new(0)))
```

### DIFF
--- a/macros/src/codegen.rs
+++ b/macros/src/codegen.rs
@@ -66,9 +66,17 @@ pub fn app(app: &App, analysis: &Analysis, extra: &Extra) -> TokenStream2 {
 
                 #(#pre_init_stmts)*
 
-                #call_init
+                #[inline(never)]
+                fn __rtic_init_resources<F>(f: F) where F: FnOnce() {
+                    f();
+                }
 
-                #(#post_init_stmts)*
+                // Wrap late_init_stmts in a function to ensure that stack space is reclaimed.
+                __rtic_init_resources(||{
+                    #call_init
+
+                    #(#post_init_stmts)*
+                });
 
                 #call_idle
             }


korken89 (approved):
Simple yet effective, thanks for the contribution!

PR #484, opened 2021-05-02
Reference monotonics by type alias

### LINKED ISSUE #470: Const generic scoping with monotonic macro
I'm looking at the [`schedule.rs` example](https://github.com/rtic-rs/cortex-m-rtic/blob/43c5ad79c27fbdefa00e2373eba554ec11e1d9df/examples/schedule.rs#L11-L18)'s monotonic type alias, simplified below.

```rust
#[rtic::app(device = lm3s6965, dispatchers = [SSI0])]
mod app {
    use dwt_systick_monotonic::DwtSystick;

    #[monotonic(binds = SysTick, default = true)]
    type MyMono = DwtSystick<8_000_000>; // 8 MHz

    // ...
}
```

This builds as expected. But suppose I want to move `8_000_000` to a `const`, making it easy to reference in both the const argument, and in [the `DwtSystick` constructor](https://github.com/rtic-rs/cortex-m-rtic/blob/43c5ad79c27fbdefa00e2373eba554ec11e1d9df/examples/schedule.rs#L26):

```rust
#[rtic::app(device = lm3s6965, dispatchers = [SSI0])]
mod app {
    use dwt_systick_monotonic::DwtSystick;

    const HZ: u32 = 8_000_000; // 8 MHz

    #[monotonic(binds = SysTick, default = true)]
    type MyMono = DwtSystick<HZ>; 

    // ...
}
```

Building reveals a sequence of errors. The compiler's hints are helpful, and we can make it work with a fully-qualified, brace-delimited constant:

```rust
#[monotonic(binds = SysTick, default = true)]
type MyMono = DwtSystick<{ crate::app::HZ }>; 
```

However, I was anticipating that it could work something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f019217823f93bffa817e3069e7305e4). We can skip the fully-qualified path, because the type alias and constant are in the same scope.

Could RTIC support this UX? Though it's unexpected, requiring the complete path isn't too troublesome, particularly when the compiler guides us to the fix.

- rustc version 1.51.0
- cortex-m-rtic version 0.6.0-alpha.2

mciantyre:
The PR tries to fix #470. I think we can correct the monotonic codegen with const generics by changing the way we reference user-defined monotonics.

Here's what the user does:

```rust
const MONO_HZ: u32 = 8_000_000;

#[monotonic(binds = SysTick, default = true)]
type MyMono = DwtSystick<MONO_HZ>;
```

Before this PR, RTIC generated code that referenced the monotonic by the right side of the type alias, `DwtSystick<MONO_HZ>`. RTIC could access the `DwtSystick` identifier from the user imports. But, RTIC doesn't seem to have a way to reference the `MONO_HZ` user-defined const.

As of this PR, RTIC references monotonics by the left side, `MyMono`. To disambiguate the type alias from the generated module of the same name (withing the `monotonics` module), we use an absolute path afforded by `app_path`.

Tested by compiling the `schedule.rs` example:

```
cargo build --example schedule --target thumbv7em-none-eabihf --all-features
```

### DIFF
--- a/examples/schedule.rs
+++ b/examples/schedule.rs
@@ -14,8 +14,11 @@ mod app {
     use dwt_systick_monotonic::DwtSystick;
     use rtic::time::duration::Seconds;
 
+    /// 8 MHz
+    const MONO_HZ: u32 = 8_000_000;
+
     #[monotonic(binds = SysTick, default = true)]
-    type MyMono = DwtSystick<8_000_000>; // 8 MHz
+    type MyMono = DwtSystick<MONO_HZ>;
 
     #[init()]
     fn init(cx: init::Context) -> (init::LateResources, init::Monotonics) {
@@ -23,7 +26,7 @@ mod app {
         let dwt = cx.core.DWT;
         let systick = cx.core.SYST;
 
-        let mono = DwtSystick::new(&mut dcb, dwt, systick, 8_000_000);
+        let mono = DwtSystick::new(&mut dcb, dwt, systick, MONO_HZ);
 
         hprintln!("init").unwrap();
 


korken89:
Hi, thanks for the PR!

It just so happens that this was part of a much larger problem.
Could you try the following branch out? It should fix the issue: https://github.com/rtic-rs/cortex-m-rtic/tree/flattened

mciantyre:
Thanks for the help! The commit on `flattened` works for the proposed test.

If you'd like to include it, I've reduced this PR to just the test case. Otherwise, I'm happy to close this PR. Looking forward to the next release.

korken89:
Thanks for testing! I have added the test case to the `flattened` branch.
I'm preparing for a new release in the coming week.

PR #489, opened 2021-06-06
Allow zero sized LinkedList

jhillyerd:
If one configures a monotonic in alpha4, but doesn't use it, TimerQueue attempts to create a zero-sized LinkedList, which causes an underflow.

This PR allows for zero-sized linked lists.


korken89 (approved):
bors merge

PR #498, opened 2021-07-21
book: update the resources chapter

japaric:
see individual commit messages for details.

what's still left to do is adjust the very last section about `#[task_local]` and `#[lock_free]` but I plan to do that as a follow up. I didn't find an in-tree example for those two attributes (are they field attributes? where do they fit in the syntax?); a quick scan of the rtic-syntax crate seems to indicate that `task_local` has been removed (?) and that `lock_free` still exists.

### DIFF
--- a/examples/resource.rs
+++ b/examples/resource.rs
@@ -61,7 +61,7 @@ mod app {
         let local_to_uart0 = cx.local.local_to_uart0;
 
         // error: no `local_to_uart1` field in `uart0::LocalResources`
-        cx.local.local_to_uart1 += 1;
+        // cx.local.local_to_uart1 += 1;
 
         hprintln!("UART0: local_to_uart0 = {}", local_to_uart0).unwrap();
     }
--- a/examples/resource.rs
+++ b/examples/resource.rs
@@ -13,55 +13,68 @@ mod app {
     use lm3s6965::Interrupt;
 
     #[shared]
-    struct Shared {
-        shared: u32,
-    }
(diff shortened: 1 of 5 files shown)


AfoHT:
I believe `task_local` attribute has been removed in favor of using the `locals`-resource instead.

examples/resource.rs
```diff
@@ -13,55 +13,69 @@ mod app {
...
-        });
-
-        hprintln!("UART0: shared = {}", shared).unwrap();
+    #[task(binds = UART0, local = [local_to_uart0])]
+    fn uart0(cx: uart0::Context) {
+        *cx.local.local_to_uart0 += 1;
+        let local_to_uart0 = cx.local.local_to_uart0;
+
+        // error: no `local_to_uart1` field in `uart0::LocalResources`
+        cx.local.local_to_uart1 += 1;
```
AfoHT, line 64: Causes the examples CI test to fail, guessing intention was to have it commented out

AfoHT (changes requested):

examples/resource.rs
japaric: ah, yes. I uncommented that to test the error message but forget to comment it again :sweat_smile:

japaric:
@AfoHT CI is green again

japaric:
while I was at it I updated the last section of this chapter (see last commit)

AfoHT (approved):
bors merge

Issue #488, opened 2021-05-29
[book] 0.6 migration: mention `lock()` being required also in highest-priority tasks now, and suggest `#[lock_free]` where applicable


birkenfeld:

japaric:
I think this was done in PRs #498 and #500 . Closing as done.

PR #500, opened 2021-07-22
migration/0.5: cover #[lock_free]

japaric:
I think this completes #488

### DIFF
--- a/book/en/src/migration/migration_v5.md
+++ b/book/en/src/migration/migration_v5.md
@@ -201,6 +201,49 @@ fn bar(cx: bar::Context) {
 
 Note that the performance does not change thanks to LLVM's optimizations which optimizes away unnecessary locks.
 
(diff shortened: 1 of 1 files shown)


crawford (approved):
It might be worth calling out that the compiler will still protect against data races when using `#[lock_free]`: `error: Lock free shared resource "test_resource" is used by tasks at different priorities`. Very cool work.

japaric:
@crawford agreed! that'd be good to highlight. I'll add that to the [main section of the book](https://github.com/rtic-rs/cortex-m-rtic/blob/c62fd967d77ccf72c08ea720c21777e367ab4125/book/en/src/by-example/resources.md#lock-free-resource-access-of-mutable-resources) in a separate PR.

AfoHT (approved):
bors merge

PR #502, opened 2021-07-22
book/resources: highlight that `#[lock_free]` includes a compile-time check

japaric:
for the "same priority requirement"; this prevents data races


AfoHT (approved):
bors merge

PR #506, opened 2021-07-28
book: Add note to remember to choose target

CuriouslyCurious:
While following the instructions in the book I ran into a wall. I was getting linker errors and had no idea why. Turns out trying to run `cargo build --example init` on an x86 build target maybe doesn't work too well, so I added a note about it. :sweat_smile:

I'll leave this as a draft since my Russian isn't fluent enough to add the note there as well. Also, adding a modified version of `.config/cargo.toml` from the embedded example might be a decent idea but wanted to hear from you guys first. :P

### DIFF
--- a/book/en/src/by-example/app.md
+++ b/book/en/src/by-example/app.md
@@ -53,6 +53,10 @@ $ cargo run --example init
 {{#include ../../../../ci/expected/init.run}}
 ```
 
+> **NOTE**: Remember to always specify your chosen target device or configure
+> one to be used by default in `.cargo/config.toml`. In this case, we use
+> a Cortex M3 running in QEMU so the target is `thumbv7m-none-eabi`.
+
 ## `idle`
 
 A function marked with the `idle` attribute can optionally appear in the
@@ -120,7 +124,7 @@ crate. When the `priority` argument is omitted, the priority is assumed to be
 `1`. The `idle` task has a non-configurable static priority of `0`, the lowest priority.
 
 > A higher number means a higher priority in RTIC, which is the opposite from what
-> Cortex-M does in the NVIC peripheral.  
+> Cortex-M does in the NVIC peripheral.
 > Explicitly, this means that number `10` has a **higher** priority than number `9`.
 
 When several tasks are ready to be executed the one with highest static


mriise:
It would also be helpful to add a note about passing the target from the command line.

CuriouslyCurious:
> It would also be helpful to add a note about passing the target from the command line.

Fair point. Will add that.

Should I also add the `.cargo/config.toml` from [cortex-m-quickstart](https://github.com/rust-embedded/cortex-m-quickstart)?

mriise:
That might be a bit much to add inline (though I am not the one to discern that), it might just be enough to refer to the quickstart for more reading.

AfoHT:
I think it is sufficient to refer to [1.6 Starting a new project](https://rtic.rs/dev/book/en/by-example/new.html) which discusses and refers to `cortex-m-quickstart`.

Otherwise this is really nice, to be merged when done :)

CuriouslyCurious:
> I think it is sufficient to refer to [1.6 Starting a new project](https://rtic.rs/dev/book/en/by-example/new.html) which discusses and refers to `cortex-m-quickstart`.
> 
> Otherwise this is really nice, to be merged when done :)

Hmm, fair point. I'll add a link to the note.

CuriouslyCurious:
The note is a bit word-y, but I think it is sufficient. Should lead people on the right track at least.

book/en/src/by-example/app.md
```diff
@@ -6,32 +6,6 @@ This is the smallest possible RTIC application:
 {{#include ../../../../examples/smallest.rs}}
 ```
 
-All RTIC applications use the [`app`] attribute (`#[app(..)]`). This attribute
```
mriise, line 9: was this intentional?

CuriouslyCurious: Completely uninentional. :eyes:

CuriouslyCurious:
I apologise for the messiness. Should not push commits when half-asleep. :sweat_smile:

AfoHT (approved):
bors merge

AfoHT:
Sadly it's blocked on #507, but it's going in ASAP

AfoHT:
bors retry

PR #509, opened 2021-08-10
Use cortex-m `InterruptNumber` instead of bare_metal `Nr`

jorgeig-space:
The trait `Nr`from crate `bare-metal` is deprecated in 1.0, and `cortex-m` moved to `InterruptNumber`, which new PACs are using. As a result, `cortex-m-rtic` does not compile when using newer PACs because the `Nr` trait is not implemented:

```
error[E0277]: the trait bound `Interrupt: Nr` is not satisfied
 --> src/main.rs:7:1
  |
7 | #[rtic::app(device = stm32wl::stm32wle5, peripherals = true)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Nr` is not implemented for `Interrupt`
  |
  = note: this error originates in the attribute macro `rtic::app` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Older PACs that still use `Nr` will still work because `cortex_m` implements `Nr` for `InterruptNumber`. This was tested using the minimal code at [https://github.com/jorgeig-space/rtic-test](https://github.com/jorgeig-space/rtic-test).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ license = "MIT OR Apache-2.0"
 name = "cortex-m-rtic"
 readme = "README.md"
 repository = "https://github.com/rtic-rs/cortex-m-rtic"
-version = "0.5.7"
+version = "0.5.8"
 
 [lib]
 name = "rtic"
@@ -51,7 +51,7 @@ name = "types"
 required-features = ["__v7"]
 
 [dependencies]
-cortex-m = "0.6.2"
+cortex-m = "0.7.3"
 cortex-m-rtic-macros = { path = "macros", version = "0.5.3" }
 rtic-core = "0.3.0"
 cortex-m-rt = "0.6.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,11 +51,20 @@ name = "types"
 required-features = ["__v7"]
 
 [dependencies]
-cortex-m = "0.7.3"
 cortex-m-rtic-macros = { path = "macros", version = "0.5.3" }
 rtic-core = "0.3.0"
 cortex-m-rt = "0.6.9"
 heapless = "0.6.1"
+
+[dependencies.cortex-m]
+package = "cortex-m"
+version = "0.6.2"
+optional = true
+
+[dependencies.cortex-m-7]
+package = "cortex-m"
+version = "0.7.3"
+optional = true
 
 [build-dependencies]
 version_check = "0.9"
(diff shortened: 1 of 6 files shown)


korken89:
Thanks for the PR!
Hmm there seems to be a version issue somewhere.
Could you have a look at what is causing this?

korken89:
Also I'm fairly sure this is a breaking change, do we would probably not be able to release it.
You'd need to add a feature flag to select the new cortex-m  for us to be able and make a release.

jorgeig-space:
> Thanks for the PR!
> Hmm there seems to be a version issue somewhere.
> Could you have a look at what is causing this?

There is an issue with the `lm3s6965` crate that is making the `checkexamples` test fail because the published version does not implement `Copy` for `Interrupt`. The latest GitHub version (published Jan 2020) does, though. Do you mind that I use the GH version throughout the crate or do you prefer to stick to published crates?

The other issue I see is with the crate `hashbrown` which is making all other tests fail. I just ran [v0.5.7](https://github.com/rtic-rs/cortex-m-rtic/commit/356f53ddc3d4b719c369a6063729b4f0e9e22c16) through the CI checks and it fails all the `check` tests with the same error as this PR.

If I try to compile the examples with Rust 1.36, they also fail with the same errors coming from `hashbrown`.

> Also I'm fairly sure this is a breaking change, so we would probably not be able to release it.
> You'd need to add a feature flag to select the new cortex-m for us to be able and make a release that's not breaking the MSRV.

If you think as maintainers that it's worth it fixing/digging into what's causing the above (it breaks the MSRV as far as I can tell), I can try helping and then implementing the feature gate, although I think it's not going to be pretty 😅

korken89:
With the feature gate it should fix itself I think. :)

jorgeig-space:
> With the feature gate it should fix itself I think. :)

I don't think so... Here is a CI run of 0.5.7 as released, with the CI forked directly from this repo: https://github.com/jorgeig-space/cortex-m-rtic/actions/runs/1118645218

Please let me know if I am doing something wrong! As it is, it doesn't matter if the feature gate is implemented or not, this PR won't pass CI.

jorgeig-space:
I've added the feature gate with the name `cortex-m-7`. Happy to rename it to what you think fits best with the crate.

Since there is no way I know of disabling a crate in Cargo, both cortex-m versions are optional and enabled via features. By default, cortex-m 0.6.2 is selected so that nothing breaks. To use cortex-m 0.7.3 and InterruptNumber, run with `--no-default-features -features cortex-m-7`.

The CI tests that use Rust Stable pass: https://github.com/jorgeig-space/cortex-m-rtic/runs/3297040636?check_suite_focus=true, but all the 1.36 / nightly fail because of other issues (see the comment above).

korken89:
Thanks for fixing!
I see the issue with CI.. I think we have to live with that.
I'll look into merging and releasing this soon!

korken89 (approved):

korken89:
Published as v0.5.8

PR #511, opened 2021-08-16
Remove linked list impl - use heapless, linked list init now const fn

korken89:


AfoHT (approved):
bors merge

:+1: Very nice!

PR #515, opened 2021-08-19
`mark_internal_ident` cleanup

### LINKED ISSUE #512: [refactor] repetitive `mark_internal_ident` API calls
there are several instances in the macros/codegen module where this pattern is used:
https://github.com/rtic-rs/cortex-m-rtic/blob/74b396497861ad69dcab0414851ddb8b5c624036/macros/src/codegen/software_tasks.rs#L39-L40

The first call creates a `task_name_fq` identifier; the second call prepends `internal_rtic` to it.

(AFAICT) the `task_name_fq` identifier is never used in the generated code so we could save the extra call by moving the `mark_internal_ident` function call into the `fq_ident` function.

That way one cannot 'forget' to call `mark_internal_ident` because `fq_ident` already returns an "internal" ident.
I think this also applies to the `inputs_ident` API.
I have checked the rest of the `util` API in detail but there may be more candidates for this simplification.

datdenkikniet:
Refactor so that all `idents` that had to call `mark_ident_internal` before now call `mark_name_internal` in the `util` module by default.

The commits can (and probably should) be squashed, they're separated for clarity.

Fixes #512

### DIFF
--- a/macros/src/codegen.rs
+++ b/macros/src/codegen.rs
@@ -108,7 +108,6 @@ pub fn app(app: &App, analysis: &Analysis, extra: &Extra) -> TokenStream2 {
             let name = &monotonic.ident;
             let name_str = &name.to_string();
             let ident = util::monotonic_ident(&name_str);
-            let ident = util::mark_internal_ident(&ident);
             let panic_str = &format!(
                 "Use of monotonic '{}' before it was passed to the runtime",
                 name_str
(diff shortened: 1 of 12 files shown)


korken89:
Thanks for the PR! Looks good to me, can you just squash the commits?
I'll merge it after that :)

datdenkikniet:
Done!

korken89 (approved):
bors merge

PR #526, opened 2021-09-16
implement run-pass tests as xtasks

Lotterleben:
resolves https://github.com/rtic-rs/cortex-m-rtic/issues/499 .

With this PR, you should be able to run `cargo xtask --target <desired target>` or `cargo xtask --target all` locally. Of course, it also reconfigures the CI workflow to do the same.

Note that I've translated the old `Run-pass tests` verbatim for now, which means the code includes checks for a `"types"`example which doesn't exist anymore.

The examples could be collected much more nicely to prevent leftovers like this in the future, but imo that could also be achieved in a separate PR.


japaric (commented):
I've overseen this work and it looks good to me. A second opinion would be great so not approving / merging this PR myself.

Lotterleben:
Converting to draft until I've debugged the Actions failure (of course, it works on my machine..)

Lotterleben:
problem resolved; un-draft-ing. I've left my fixup PRs for transparency, happy to squash after review.

xtask/src/main.rs
```diff
@@ -0,0 +1,388 @@
...
+        }
+    }
+}
+
+impl Error for TestRunError {}
+
+fn main() -> anyhow::Result<()> {
+    // if there's an `xtask` folder, we're *probably* at the root of this repo (we can't just
+    // check the name of `env::current_dir()` because people might clone it into a different name)
+    let probably_running_from_repo_root = Path::new("./xtask").exists();
```
AfoHT, line 72: Might consider checking for `.git`?

AfoHT (commented):
Interesting! Looks nice, need to try this out on computer too 😃

korken89 (approved):

korken89:
bors merge

xtask/src/main.rs
korken89: YOLO, missed this comment. We can fix later :)

Lotterleben: > Might consider checking for `.git`?

I intentionally didn't do that in case someone has some wild submodule-ing going on :)

PR #528, opened 2021-09-20
The great 0.6 docs update

### LINKED ISSUE #422: Document the critical section available in init as `cx.cs`

### LINKED ISSUE #440: [Book] Add chapter for hardware tasks/interrupts
The RTIC book would benefit from having an Interrupts/Hardware tasks chapter. As it stands, they aren't introduced formally, so the reader is left scrolling around the Resources and Software tasks chapters to learn about them.

Suggested topics:

- How to define a hardware task
- ` check_interrupt()`, `clear_interrupt_pending_bit()`, `clear_update_interrupt_flag()` still relevant or required in RTIC.

Points to make:

- A task fn can only be bound to single interrupt
- Bound interrupts must be removed from the app(dispatchers) list
- If you are defining a high priority task, you may want to leverage messaging queuing to do larger chunks of work in a lower priority software task

### LINKED ISSUE #448: [Book] Using `extern` in 0.6 is missing
As the title states, needs updating.

### LINKED ISSUE #461: [Book] Add 0.6 migration note about `Monotonic` priority
I think it's worth noting that the default timer queue priority is now `1`, unless specified. I have a feeling many people will run into https://github.com/rtic-rs/cortex-m-rtic/issues/458 without this information.

Perhaps an example setting the `Monotonic` priority would be a good idea too.

### LINKED ISSUE #487: dev book: `examples/baseline.rs` missing
https://rtic.rs/dev/book/en/by-example/timer-queue.html at the bottom shows just 
```
{{#include ../../../../examples/baseline.rs}}
```

instead of the example.

### LINKED ISSUE #527: refactor: move test metadata into test file source code
something that came up during PR #526 is that there's test specific information hard coded in the xtask logic.
this information could be moved into the test file source code (`examples/*.rs`)

the Rust compiler test suite uses this approach: it encodes test metadata in code comments.

for the particular case of RTIC the information that we need to encode is (off the top of my head)
- whether the test is ARMv7 only
- whether the example should NOT be run in QEMU (no CYCCNT in QEMU = spawn_at API doesn't work)

what this could look like is (straw man proposal)

``` console
$ head -n1 examples/t-schedule.rs
// skip-run-test
```

that's the metadata for compile this example but don't run it in QEMU

``` console
$ head -n1 examples/pool.rs
// v7-only
```

that's the metadata for "only compile this for the thumbv7m-none-eabi target"

an advantage of this approach is that you don't need to list the examples in the `xtask` source code, which is bound to get out of date
instead you can have the `xtask` scan *all*  the files in `examples/` and decide what to do with the example (compile-pass for v6 and v7; run-pass; compile-pass for v7 only; etc.) based on the metadata

### LINKED ISSUE #530: RUSTSEC-2018-0017: `tempdir` crate has been deprecated; use `tempfile` instead
> `tempdir` crate has been deprecated; use `tempfile` instead

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Status              | unmaintained                |
| Package             | `tempdir`                      |
| Version             | `0.3.7`                   |
| URL                 | [https://github.com/rust-lang-deprecated/tempdir/pull/46](https://github.com/rust-lang-deprecated/tempdir/pull/46) |
| Date                | 2018-02-13                         |

The [`tempdir`](https://crates.io/crates/tempdir) crate has been deprecated
and the functionality is merged into [`tempfile`](https://crates.io/crates/tempfile).

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2018-0017.html) for additional details.

korken89:
Closes #530
Closes #527
Closes #487
Closes #461
Closes #448
Closes #440
Closes #422


AfoHT (approved):
Time to get the ball rolling

bors merge

PR #529, opened 2021-09-21
Updated codegen for the updated syntax (default monotonic priority)

korken89:
Needs new syntax release before merge.


AfoHT (approved):
bors merge

PR #531, opened 2021-09-23
v0.5: Remove cortex-m-rt dependency.

Dirbaio:
It's not needed, since the user has to supply a PAC which will already depend on a suitable
cortex-m-rt version.

This allows using RTIC v0.5 with either cortex-m-rt 0.6 or 0.7.


korken89 (approved):

Dirbaio:
I have no idea where the microamp issue comes from. I think it worked with older nightly but the CI is using latest nightly.

PR #537, opened 2021-09-27
[PoC] actor API

japaric:
this PR contains a proof-of-concept implementation of the actor API proposed in rtic-rs/rfcs#52
the files added to the `examples` directory are tests of the feature semantics
a full example, with multiple crates (firmware + testable `no_std` library), has been included in the `actor-example` directory


japaric:
@rtic-rs/devs this PR has been rebased onto rtic 1.0 and the deps of the firmware have also been bumped :rocket:

japaric:
ping @korken89 this and the rtic-syntax PRs have been rebased

jscatena88:
Looking at the `actor-example` multi-crate example I am curious as to why `FakeTempratureSensor` is defined as an actor in the actor crate then not listed in the `Actors` struct. Instead an instance of it saved in `Local`. Why would you not just save a copy of the Poster in `Local` and have the periodic task use that to post a message into the Actor system. Seems confusing to have an Actor that isn't quite used as an Actor.

japaric:
indeed, having the temp sensor as an actor instance makes more sense. updated the example.

actor-example/actors/src/temperature_monitor.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+impl<P> Receive<TemperatureReadingCelsius> for TemperatureMonitor<P>
+where
+    P: Post<TemperatureAlert>,
+{
+    fn receive(&mut self, temperature: TemperatureReadingCelsius) {
+        if temperature.0 >= self.threshold {
+            self.outbox.post(TemperatureAlert).ok().expect("OOM");
```
jscatena88, line 28: Why is `ok()` called here when you could call expect directly on the return from `post()` i.e.
`self.outbox.post(TemperatureAlert).expect("OOM");`

japaric: doing `expect` on the `Result<T, E>` value requires that `E` implements the `fmt::Debug` trait. doing `ok().expect()` places no trait requirements on the type `E`. `TemperatureAlert` does not implement the `fmt::Debug` trait so `post(TemperatureAlert).expect()` would not compile.

deriving `Debug` on the type and leaving out the `ok()` is also an option (no pun intended). I tend to write `ok().expect()` mostly out of bad habit. (less `Result::unwrap` calls produces smaller (code size) binaries but that's micro-optimizing and this is an example so that hardly matters; as I said: bad habit)

jscatena88: Ahh, ok makes sense. Thanks for the reply, I was mainly just curious more than critiquing

bb010g:
What's the status of this, now that v2.0 is stable?

AfoHT:
> What's the status of this, now that v2.0 is stable?

The proposed path forward is to create a separate branch for actor, just like other older releases.
I did rebase this actor branch onto latest RTIC v1, then they could be released as alphas or similar.

PR #546, opened 2021-10-26
Lockall

perlindgren:
Lockall api [53](https://github.com/rtic-rs/rfcs/issues/53).

Issue with:
```shell
> cargo xtask   --target thumbv7m-none-eabi
```
Builds in --release which fails run-test for 
```shell
cargo run --example cfg-whole-task --target thumbv7m-none-eabi --release
```
Not sure what to do...


AfoHT:
@perlindgren Is this still relevant for v1 and something you plan to pursue further?

PR #550, opened 2021-11-06
Release 0.6.0-rc.3 due to breaking change in rtic-syntax naming of shared and local

AfoHT:
`rtic-syntax` got the old multicore concept of `Locations` removed, see https://github.com/rtic-rs/rtic-syntax/pull/61

This updates `cortex-m-rtic` to match this


korken89 (approved):
bors merge

PR #562, opened 2021-12-14
Replace default WFI with simpler NOP

### LINKED ISSUE #561: Default idle task (`SLEEPONEXIT`/`wfi()`) breaks RTT and flashing
In RTIC v0.6.x, it is no longer necessary to explicitly declare the background idle task as it has an implicit default implementation as documented in https://rtic.rs/dev/book/en/by-example/app_idle.html:

https://github.com/rtic-rs/cortex-m-rtic/blob/235484565d706d362f37c8bd388dbfff78401e8a/macros/src/codegen/idle.rs#L87-L89

Unfortunately, both `SLEEPONEXIT` and `wfi()` have the unintended side-effect that debug connections, in particular RTT breaks on some devices. This is explained in more detail in [this section](https://wiki.segger.com/RTT#Low-power_modes) from the SEGGER Wiki:

> Many CPU cores support low-power modes to save power while waiting for an event or interrupt. On Cortex-M for example this is achieved by executing the WFI and WFE instructions. Unfortunately, the exact behavior during low-power modes is depending on the actual device (two devices from different vendors incorporating the same Cortex-M core may behave different).
> 
> Some devices disable the whole debug unit during low-power modes, causing J-Link to lose the connection to the device.
Some keep the debug unit enabled but disable the internal flash and RAM, making it effectively impossible for J-Link to perform RTT requests while the target is in a low-power mode
> As the target will be idle most of the time, using low-power modes will effectively make the usage of RTT impossible, as the target memory cannot be accessed 99% of the time while the application is running. Halting the target application will make the device to make up and RAM etc. becomes accessible again. This may be seen as effects like "RTT output is only seen each time the target is halted".
> 
> **Solution:** When using RTT, make sure that low-power modes are **not used**.

The problem happens not only with J-Link but with other probes as well, see these issues:

- https://github.com/knurling-rs/probe-run/issues/128
- https://github.com/probe-rs/probe-rs/issues/350

There doesn't seem to be an easy and universal solution but to disable sleep modes when debugging.

Since the default implementation of the background idle task going to sleep seems practical for most use-cases nonetheless, I suggest an addition to the documentation instead:

- In [The background task `#[idle]`](https://rtic.rs/dev/book/en/by-example/app_idle.html), add a prominent note that states the expected problems with debugging when the default idle task is not overwritten.
- In [The `#[app]` attribute and an RTIC application](https://rtic.rs/dev/book/en/by-example/app.html), add a comment to the example idle task implementation that states we don't enter sleep mode here on purpose.

Alternatively, the low-power default implementation could be made an explicit opt-in, e.g.

```rust
#[rtic::app(device = lm3s6965, dispatchers = [SSI0, QEI0], sleep = true)]
```

Unless given, the default idle implementation would neither employ `SLEEPONEXIT` nor `wfi()` but spin in a busy loop.

AfoHT:
As noted by #561 there are multiple issues with various hardware implementations and debugging together with sleep modes.

As RTIC strives to be a generic framework (even though this is an implementation targeting cortex-m) the decision having `WFI` as the default `idle` task causes issues in some hardware, raising the barrier to entry.

This changes the default behaviour to do a simple `NOP` instead, adding documentation how to provide a custom `idle` task achieving the old default behaviour. Also removes the automatic setting of SLEEPONEXIT bit when no `idle` was provided, delegating this to the user.

This was discussed on the weekly meeting 2021-12-14 and this was the favoured solution.

Fixes #561


korken89 (approved):
bors merge

PR #581, opened 2022-01-05
Relax spawn_after parameter

ZoeyR:
With RTIC 1.0 if you use `embedded_time` on your `Monotonic` impl any call to `task::spawn_after(...)` has to put a `.into()` after the time type.
```rust
task::spawn_after(Milliseconds(250).into())
```
This is because the implementation of `Monotonic` requires using `embedded_time::duration::Generic` for the `Duration` associated type. This PR changes the bounds on the `spawn_after` function to take any type that can be converted into the target duration type. Removing the need to type all the `into()`s.

There are some backward-compatibility concerns since any code that previously used .into() will now fail the type inference. I'm also not sure how this affects the type inference for any `Monotonic` using `fugit` since I don't have a firmware repo to test with that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 - CI changelog entry enforcer
 
+### Changed
+
+- `spawn_after` now takes an `Into<Duration>` instead of bare `Duration`
+
 ## [v1.0.0] - 2021-12-25
 
 ### Changed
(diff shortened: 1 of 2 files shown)


korken89:
This seems fine to me. 
But it would be good to check if this had any negative effects such as increased size when the concrete type is not available anymore and must go through a trait.
Both for embedded time and fugit.

korken89:
Also, could you add a changelog line so it passes CI?

ZoeyR:
Sure thing! I've pushed the update.

AfoHT:
Was there any size increase when using embedded time or fugit in this way?

burrbull:
fugit will not work with Into

PR #585, opened 2022-01-08
Fix the locals.rs comment

keadwen:
I believe that is a typo s/shared/local_to_bar


korken89 (approved):
Thanks for fixing!

bors merge

PR #588, opened 2022-01-21
Highlight how to run examples locally

### LINKED ISSUE #587: [Book] Make it easy to follow along with the examples
It would be nice to add instructions on how to follow along with the 1.0 version of the book. At the moment it just points to The embedded rust book.

I followed all the examples from The embedded rust book previously, so I have all the tools installed. However, just using the [cortex-m-quickstart](https://docs.rust-embedded.org/book/start/qemu.html) and pasting the code of the [first example](https://github.com/rtic-rs/cortex-m-rtic/blob/master/examples/common.rs) to main.rs will not work out of the box. It leaves me to solve all the compile errors caused by missing packages and unstable features.

It would be nice to have a similar template as cortex-m-quickstart for rtic, and/or a bit more elaborate instructions on what to do if you want to follow along with the examples.

Some of the issues are:
- missing `panic-semihosting` in `Cargo.toml`
- missing `systick-monotonic` in `Cargo.toml`
- panicking in constants is unstable (currently stuck here)

AfoHT:
How examples are run never gets properly introduced before used in `by-example/resources.md`

Fixes #587

### DIFF
--- a/book/en/src/by-example.md
+++ b/book/en/src/by-example.md
@@ -15,3 +15,18 @@ Check [the embedded Rust book] for instructions on how to set up an
 embedded development environment that includes QEMU.
 
 [the embedded Rust book]: https://rust-embedded.github.io/book/intro/install.html
+
+To run the examples found in `examples/` locally, cargo needs a supported `target` and
+either `--examples` (run all examples) or `--example NAME` to run a specific example.
+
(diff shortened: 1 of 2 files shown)


perlindgren (approved):
bors merge

PR #591, opened 2022-01-24
Fix running command examples in chapter 1.5.5 Monotonic

GreenPenguino:
Run the right examples in chapter 1.5.5 in the book


AfoHT (approved):
Nice catch!

bors merge

PR #594, opened 2022-02-02
Fix link to nRF monotonic

AfoHT:


jordens:
Better use a permalink. This one and the stm32f4 one are both also fixed in #595.

AfoHT:
> Better use a permalink. This one and the stm32f4 one are both also fixed in #595.

Indeed, nice work on that one btw, will close this one and merge #595 instead.

PR #595, opened 2022-02-03
Update tips_monotonic_impl.md

### LINKED ISSUE #597: [Book] references 0.6 instead of 1.0
https://rtic.rs/1.0/book/en/by-example/tips_monotonic_impl.html

`For RTIC 0.6 we instead assume the user has a time library,`

jordens:
* There is no RTIC 0.6, only several RCs.
* Timers without interrupts (like the DWT cycle counter alone) will not be useful for `Monotonic` impls.
* Clarified some of the descriptions of the various implementations.

Close #597 
Close #594


AfoHT (approved):
bors merge

PR #599, opened 2022-02-05
Docs: SW and HW tasks

### LINKED ISSUE #576: Dispatcher documentation
As a newcomer to this framework, I'm trying to get software tasks to work within my application running on the `stm32h7xx_hal` and I don't fully understand whats going on with Dispatchers and their dependency on the hardware, specifically I find this line [in the book](https://rtic.rs/1.0/book/en/by-example/software_tasks.html) is a little misleading `Software tasks are tasks which are not directly assigned to a specific interrupt vector`. At first I interpreted this as if a Dispatcher does not use any kind of hardware based interrupt at all, but then the documentation goes on to say that you need to assign a "free" interrupt vector to act as a dispatcher.

`stm32h7xx_hal` has no `SSI0` or `QEI0` symbols available, can _any_ unused interrupt vector be used? E.g. If my app is not using `EXTI0`, then this could be used? Are there any interrupts that generally shouldn't be used?

I think the documentation around this could be improved to help people

AfoHT:
Reword and (hopefully) clarify SW and HW tasks

Closes #576

### DIFF
--- a/book/en/src/by-example/hardware_tasks.md
+++ b/book/en/src/by-example/hardware_tasks.md
@@ -14,11 +14,18 @@ start execution in reaction to a hardware event.
 Specifying a non-existing interrupt name will cause a compilation error. The interrupt names
 are commonly defined by [PAC or HAL][pacorhal] crates.
 
+Any available interrupt vector should work, but different hardware might have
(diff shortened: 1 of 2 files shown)


korken89 (approved):
bors merge

PR #603, opened 2022-02-08
Add a remark about `Sync` and `Send` traits requirement for resources

Glaeqen:


book/en/src/by-example/resources.md
```diff
@@ -30,6 +30,11 @@ task.
 Thus, a task `#[local]` resource can only be accessed by one singular task.
 Attempting to assign the same `#[local]` resource to more than one task is a compile-time error.
 
+Types of `#[local]` resources must implement [`Send`] trait as they are being sent from `init`
```
Glaeqen, line 33: Is it appropriate to add the example of why this is reasonable? Here I'm referring to @japaric 's example with `Rc` type.

korken89: It's fine without IMO.

AfoHT (approved):
bors merge

PR #612, opened 2022-02-15
Allow naming lock type of shared resources

### LINKED ISSUE #611: Naming lock types
I want to pass a shared resource to a function, but I'm unable to name the lock.
The path `app::shared_resources::resource_that_needs_to_be_locked` is not accessible, as `app::shared_resources` is private.
Also the generated name `resource_that_needs_to_be_locked` is inconvenient.

LU15W1R7H:
This PR allows users to name the lock types of shared resources. This is useful to pass the locks to other functions.
It's not always possible to just specify the argument type as `impl rtic::Mutex<Resource>`, because the function then is no longer trait object safe. The type can now be named as `app::shared_resources::{}_lock`.

Example:
This trait is now object safe!
```rust
trait TraitObject {
  fn func(
    &self,
    lock: &mut app::shared_resources::resource_lock,
  );
}

```

changes:

- change lock type ident: `{}_that_needs_to_be_locked` => `{}_lock`
- change visibility of `shared_resources` module to pub

closes #611

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - `examples/periodic-at.rs`, an example of a periodic timer without accumulated drift.
 - `examples/periodic-at2.rs`, an example of a periodic process with two tasks, with offset timing. Here we depict two alternative usages of the timer type, explicit and trait based.
 - book: Update `Monotonic` tips.
+- rename lock types `{}_lock` (instead of `{}_that_needs_to_be_locked`)
+- change visibility of `app::shared_resources` module to pub
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


LU15W1R7H:
What's the state of this PR? Are you interested in this addition?

I provided a MWE as mentioned in [this comment](https://github.com/rtic-rs/cortex-m-rtic/issues/611#issuecomment-1040720488). Showing what this PR solves. I think this is a nice improvement, especially since it involves little modifications of RTIC.

perlindgren:
I think the main concern here is that it exposes the implementation details. I don't think it would be unsound to expose though. The underlying problem seems to be related to the creation vtables in Rust (so it cannot be made into a trait object). 

I've been experimenting with another internal representation of locks that might solve this particular problem. It won't solve the underlying problem with vtables totally though. There might be other issues with the new lock implementation that comes up along the way, thus as of today its unclear if it will be implemented or not. I think there was also some other potential fix discussed, need to get back on that.
 
/Per

korken89:
Hi,

After today's' meeting we decided that we don't want to give this guarantee of an implementation detail.
As long as you find this useful you are welcome to roll your own fork of RTIC.

Thank you for the effort!

PR #623, opened 2022-03-04
[testing] Embedded ci

korken89:


.github/workflows/build.yml
```diff
@@ -172,12 +172,21 @@ jobs:
...
       - name: Run-pass tests
-        run:
-          cargo xtask --target ${{ matrix.target }}
+        env:
+          EMBEDDED_CI_TOKEN: ${{ secrets.EMBEDDED_CI_TOKEN }}
+          EMBEDDED_CI_SERVER: ${{ secrets.EMBEDDED_CI_SERVER }}
+        run: |
+          cd examples-runner
+          cargo xtask --target ${{ matrix.target }} --runner embedded-ci
+          cargo xtask --target ${{ matrix.target }} --runner qemu
```
AfoHT, line 189: I think we should split this and have two different `checkexamples` allowing them to run concurrently as this is the slowest task, we're currently not using all available 20 concurrent runners :+1:

PR #626, opened 2022-03-20
Fix error in book, shared resource need only `Send`

korken89:


AfoHT (approved):
bors merge

korken89:
bors merge

PR #629, opened 2022-04-03
Attempting to add conceptual model description

jvanderk:
This is an attempt to provide a description of the execution model of RTIC at an abstract level.

I have had an interesting learning experience with git in this attempt. I have tried to disentangle as much as I can. Please bear with me.


AfoHT:
From @01joja in Matrix chatroom:

A different perspective, but also a visual interpretation

image

PR #633, opened 2022-04-14
armv6m MASKS issue, temporary fix

AfoHT:
- allow const_err on MASKS
- example added
- Assume default target armv7m


AfoHT:
This is reported as not a fix, not even a hot-glue one, thus we need a proper solution instead.

Closing

PR #635, opened 2022-04-20
Masks take 3

korken89:
This solves the `MASKS` generation issue by having `rtic::export` do the feature gating.

### DIFF
--- a/macros/src/codegen/shared_resources.rs
+++ b/macros/src/codegen/shared_resources.rs
@@ -110,12 +110,32 @@ pub fn codegen(
 
     let mut prio_to_masks = HashMap::new();
     let device = &extra.device;
+    let mut uses_exceptions_with_resources = false;
 
     for (&priority, name) in interrupt_ids.chain(app.hardware_tasks.values().flat_map(|task| {
         if !util::is_exception(&task.args.binds) {
             Some((&task.args.priority, &task.args.binds))
         } else {
-            // TODO: exceptions not implemented
+            // If any resource to the exception uses non-lock-free or non-local resources this is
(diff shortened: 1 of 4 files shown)


macros/src/codegen/shared_resources.rs
```diff
@@ -124,18 +119,31 @@ pub fn codegen(
             None
```
korken89, line 119: What to do when this happens? Panic? If this just "falls through" we have have UB.

perlindgren: We could reject RTIC applications on armv6 that uses shared lock free resources tied to exceptions (the exceptions cannot be easily masked by the NVIC). This would be a very rare case though, e.g., using SysTick as a task with a shared resource, the monotonic implementation using SysStick should be fine as it does not use any shared resource (at least not the current implementation). A short summary of the v6 exceptions, and usability as generic tasks.

1 | Reset, not usable
2 | NMI, not usable
3 | HardFault, not usable
4-10 | Reserved, not usable
11 | SVCall, ... hmm perhaps, but never seen it....
12-13 | Reserved
14 | PendSV, ... hmm perhaps, but never seen it
15 | SysTick, ... seems like the only usable case

perlindgren: The latest fix proposal should take care of that, our current way to perform tests on v6 vs v7 is not covering all cases though, and the whole test approach needs to be revisited at some point. I think however we can live with this atm.

perlindgren (approved):
bors r+

AfoHT:
bors merge

PR #643, opened 2022-05-24
Fix clash with defmt

### LINKED ISSUE #642: RTIC introduces ambiguous assert!
i'm using `defmt::*` which imports `defmt::assert`

since my last `cargo update` (`cortex-m-rtic-macros v1.0.0 -> v1.1.2`)
a conflict with another `assert` macro was introduced
image

switching to `use defmt::{trace, debug}` resolved this
(i was already using `defmt::assert!` explicitly, possibly for a similar problem)

the RTIC matrix channel suggested i look at `rtic-expansion.rs`
here is the snippet from there i believe to conflict with `defmt::assert` (via `defmt::*`)
```
{
    if rtic :: export :: is_armv6()
    {
        assert!
        ((stm32f4xx_hal :: pac :: Interrupt :: OTG_FS as u32) < 32)
        ;
    } else {}
} ;
```
@jamesmunns suggested that this problem could be avoided if RTIC namespaced its use of the assert macro

korken89:
Fixes #642

### DIFF
--- a/macros/Cargo.toml
+++ b/macros/Cargo.toml
@@ -12,7 +12,7 @@ license = "MIT OR Apache-2.0"
 name = "cortex-m-rtic-macros"
 readme = "../README.md"
 repository = "https://github.com/rtic-rs/cortex-m-rtic"
-version = "1.1.2"
+version = "1.1.3"
 
 [lib]
 proc-macro = true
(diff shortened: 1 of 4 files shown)


AfoHT (approved):
bors merge

PR #645, opened 2022-06-07
fix ci: use SYST::PTR

japaric:
SYST::ptr has been deprecated in cortex-m v0.7.5
SYST::PTR is available since cortex-m v0.7.0

CI was failing due to a warning turned into an error by `deny(warnings)`


korken89 (approved):
bors merge

PR #646, opened 2022-06-12
Async/await support

korken89:
Async/await codegen

- [x] ~~`async idle`~~
- [ ] Support 0-priority async tasks for non-`Send` argument
- [x] Disallow task local resources in `async` tasks

Design document: https://hackmd.io/6ejCFNBJTuKBwnuz0O41iQ

Depends on https://github.com/rtic-rs/rtic-syntax/pull/78


macros/src/codegen/dispatchers.rs
```diff
@@ -73,27 +76,88 @@ pub fn codegen(app: &App, analysis: &Analysis, _extra: &Extra) -> Vec<TokenStrea
...
+                                let #tupled =
+                                    (&*#inputs
+                                    .get())
+                                    .get_unchecked(usize::from(index))
+                                    .as_ptr()
+                                    .read();
+                                (&mut *#fq.get_mut()).split().0.enqueue_unchecked(index);
+
+                                let priority = &rtic::export::Priority::new(PRIORITY);
+                                (&mut *#exec_name.get_mut()).spawn(#name(#name::Context::new(priority), #(,#pats)*));
```
Dirbaio, line 97: there's one extra comma:

```suggestion
                                (&mut *#exec_name.get_mut()).spawn(#name(#name::Context::new(priority) #(,#pats)*));
```

korken89: Nice Catch! That would make task arguments stop working.

examples/async-infinite-loop.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    #[idle]
+    fn idle(_: idle::Context) -> ! {
+        loop {
+            cortex_m::asm::wfi(); // put the MCU in sleep mode until interrupt occurs
+        }
+    }
+
+    // Infinite loops are not allowed in RTIC, however in async tasks they are - in there is an
```
MabezDev, line 41:
```suggestion
    // Infinite loops are not allowed in RTIC, however in async tasks they are - if there is an
```

MathiasKoch:
Minor thing I just found while toying around with this.

It doesn't seem like externalizing tasks with `extern "Rust"` works with async tasks.

korken89:
Correct, you can however just make an external task and call it. This works just as well

datdenkikniet:
Should this PR be closed in favour of #694 ?

PR #649, opened 2022-06-23
Bump rtic-syntax to v1.0.2 and fix Changelog

AfoHT:
Use the latest rtic-syntax, update the changelog with the last few undocumented releases


korken89 (approved):
bors merge

PR #652, opened 2022-07-03
Remove use of basepri register on thumbv8m.base

neonquill:
The basepri register appears to be aviable on thumbv8m.main but not thumbv8m.base. At the very least, attempting to compile against a Cortex-M23 based Microchip ATSAML10E16A generates an error:

```
error[E0432]: unresolved import `cortex_m::register::basepri`
  --> /Users/dwatson/.cargo/registry/src/github.com-1ecc6299db9ec823/cortex-m-rtic-1.1.3/src/export.rs:25:5
   |
25 | use cortex_m::register::basepri;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `basepri` in `register`
```

I wasn't sure if it made more sense to replace the `armv7m` config flag with something related to basepri availability or to get closer to matching the cortex-m use of several architecture specific flags. In the end i chose to make the minimal change possible and just narrowed the existing `thumbv8m` check.

Context:
[cortex-m:src/register/mod.rs](https://github.com/rust-embedded/cortex-m/blob/4e908625204a1e95dd3fd5bdcd8d66d6bc11c3bc/src/register/mod.rs#L33):
```
#[cfg(all(not(armv6m), not(armv8m_base)))]
pub mod basepri;
```

[cortex-m:build.rs](https://github.com/rust-embedded/cortex-m/blob/4e908625204a1e95dd3fd5bdcd8d66d6bc11c3bc/build.rs#L21):
```
    } else if target.starts_with("thumbv8m.base") {
        println!("cargo:rustc-cfg=cortex_m");
        println!("cargo:rustc-cfg=armv8m");
        println!("cargo:rustc-cfg=armv8m_base");
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ### Fixed
 
+- Distinguish between thumbv8m.base and thumbv8m.main for basepri usage.
+
 ### Changed
 
 ## [v1.1.3] - 2022-06-23
--- a/build.rs
+++ b/build.rs
@@ -7,15 +7,21 @@ fn main() {
         println!("cargo:rustc-cfg=rustc_is_nightly");
     }
 
-    if target.starts_with("thumbv6m") {
-        println!("cargo:rustc-cfg=armv6m");
-    }
-
+    // These targets all have know support for the BASEPRI register.
     if target.starts_with("thumbv7m")
         | target.starts_with("thumbv7em")
-        | target.starts_with("thumbv8m")
+        | target.starts_with("thumbv8m.main")
     {
-        println!("cargo:rustc-cfg=armv7m");
+        println!("cargo:rustc-cfg=have_basepri");
+
+    // These targets are all known to _not_ have the BASEPRI register.
+    } else if target.starts_with("thumb")
+        && !(target.starts_with("thumbv6m") | target.starts_with("thumbv8m.base"))
+    {
+        panic!(
+            "Unknown target '{}'. Need to update BASEPRI logic in build.rs.",
+            target
+        );
     }
 
     println!("cargo:rerun-if-changed=build.rs");
--- a/macros/src/codegen/assertions.rs
+++ b/macros/src/codegen/assertions.rs
@@ -22,15 +22,16 @@ pub fn codegen(app: &App, analysis: &Analysis, extra: &Extra) -> Vec<TokenStream
     }
 
     let device = &extra.device;
-    let arm_v6_checks: Vec<_> = app
+    let chunks_name = util::priority_mask_chunks_ident();
+    let no_basepri_checks: Vec<_> = app
         .hardware_tasks
         .iter()
         .filter_map(|(_, task)| {
             if !util::is_exception(&task.args.binds) {
                 let interrupt_name = &task.args.binds;
                 Some(quote!(
-                    if (#device::Interrupt::#interrupt_name as u32) > 31 {
-                        ::core::panic!("An interrupt above value 31 is used while in armv6");
+                    if (#device::Interrupt::#interrupt_name as usize) >= (#chunks_name * 32) {
+                        ::core::panic!("An interrupt out of range is used while in armv6 or armv8m.base");
                     }
                 ))
             } else {
@@ -41,8 +42,8 @@ pub fn codegen(app: &App, analysis: &Analysis, extra: &Extra) -> Vec<TokenStream
 
     let const_check = quote! {
         const _CONST_CHECK: () = {
-            if rtic::export::is_armv6() {
-                #(#arm_v6_checks)*
+            if !rtic::export::have_basepri() {
+                #(#no_basepri_checks)*
             } else {
                 // TODO: Add armv7 checks here
             }
--- a/macros/src/codegen/shared_resources.rs
+++ b/macros/src/codegen/shared_resources.rs
@@ -112,6 +112,8 @@ pub fn codegen(
     let device = &extra.device;
     let mut uses_exceptions_with_resources = false;
 
+    let mut mask_ids = Vec::new();
+
     for (&priority, name) in interrupt_ids.chain(app.hardware_tasks.values().flat_map(|task| {
         if !util::is_exception(&task.args.binds) {
             Some((&task.args.priority, &task.args.binds))
@@ -141,12 +143,13 @@ pub fn codegen(
     })) {
         let v = prio_to_masks.entry(priority - 1).or_insert(Vec::new());
         v.push(quote!(#device::Interrupt::#name as u32));
+        mask_ids.push(quote!(#device::Interrupt::#name as u32));
     }
 
-    // Call rtic::export::create_mask([u32; N]), where the array is the list of shifts
+    // Call rtic::export::create_mask([Mask; N]), where the array is the list of shifts
 
     let mut mask_arr = Vec::new();
-    // NOTE: 0..3 assumes max 4 priority levels according to M0 spec
+    // NOTE: 0..3 assumes max 4 priority levels according to M0, M23 spec
     for i in 0..3 {
         let v = if let Some(v) = prio_to_masks.get(&i) {
             v.clone()
@@ -159,18 +162,26 @@ pub fn codegen(
         ));
     }
 
+    // Generate a constant for the number of chunks needed by Mask.
+    let chunks_name = util::priority_mask_chunks_ident();
+    mod_app.push(quote!(
+        #[doc(hidden)]
+        #[allow(non_upper_case_globals)]
+        const #chunks_name: usize = rtic::export::compute_mask_chunks([#(#mask_ids),*]);
+    ));
+
     let masks_name = util::priority_masks_ident();
     mod_app.push(quote!(
         #[doc(hidden)]
         #[allow(non_upper_case_globals)]
-        const #masks_name: [u32; 3] = [#(#mask_arr),*];
+        const #masks_name: [rtic::export::Mask<#chunks_name>; 3] = [#(#mask_arr),*];
     ));
 
     if uses_exceptions_with_resources {
         mod_app.push(quote!(
             #[doc(hidden)]
             #[allow(non_upper_case_globals)]
-            const __rtic_internal_V6_ERROR: () = rtic::export::v6_panic();
+            const __rtic_internal_V6_ERROR: () = rtic::export::no_basepri_panic();
         ));
     }
 
--- a/macros/src/codegen/util.rs
+++ b/macros/src/codegen/util.rs
@@ -253,6 +253,11 @@ pub fn static_shared_resource_ident(name: &Ident) -> Ident {
     mark_internal_name(&format!("shared_resource_{}", name))
 }
 
+/// Generates an Ident for the number of 32 bit chunks used for Mask storage.
+pub fn priority_mask_chunks_ident() -> Ident {
+    mark_internal_name("MASK_CHUNKS")
+}
+
 pub fn priority_masks_ident() -> Ident {
     mark_internal_name("MASKS")
 }
--- a/src/export.rs
+++ b/src/export.rs
@@ -21,10 +21,43 @@ pub use rtic_monotonic as monotonic;
 pub type SCFQ<const N: usize> = Queue<u8, N>;
 pub type SCRQ<T, const N: usize> = Queue<(T, u8), N>;
 
-#[cfg(armv7m)]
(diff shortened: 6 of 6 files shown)


neonquill:
So, I have little insight into the correctness questions, but I can add some data to the conversation:

In the [Armv8M Architecture Reference](https://developer.arm.com/documentation/ddi0553/latest) section B3.32:

> In a PE with the Main Extension, the PRIMASK, FAULTMASK, and BASEPRI registers can be used as follows.
> A PE without the Main Extension implements PRIMASK, but does not implement FAULTMASK and BASEPRI.

From section A1.3.2 of the same doc:

> If an implementation does not include the Main Extension CPUID.Architecture reads as 0b1100.

From the [ARM Cortex-M23 Processor Technical Reference Manual r1p0 CPUID docs](https://developer.arm.com/documentation/ddi0550/c/system-control/system-control-register-summary/cpuid-register?lang=en):

> Indicates the architecture, ARMv8-M baseline:
> 0xC.

From my perspective, it feels relatively conclusive that M23 doesn't include BASEPRI.

Back to your question about correctness:
> but does that guarantee that mask based scheduling will work? It assumes the m0/m0+ limitation of 32 interrupts (and does not work on the exceptions, such as Systick).

That is a question I didn't think to ask 🙂! I do not understand the internals of rtic enough to answer this question, but here's some data.

From the [SAML10/11 datasheet](https://ww1.microchip.com/downloads/aemDocuments/documents/MCU32/ProductDocuments/DataSheets/SAM-L10L11-Family-Data-Sheet-DS60001513G.pdf):

> 11.2.1 Overview
> The Nested Vectored Interrupt Controller (NVIC) in the SAM L10/L11 supports up to 45 interrupt lines with four different priority levels + 1 Non Maskable Interrupt (NMI) line.
> For more details, refer to the Cortex-M23 Technical Reference Manual (www.arm.com).

From section A1.3.3 of the Armv8M Architecture Reference:

> The Nested Vectored Interrupt Controller (NVIC) is used for integrated interrupt and exception handling and prioritization. Armv8-M increases the number of interrupts that can potentially be supported by the NVIC is 480 for external sources, and includes automatic vectoring and priority management, and automatic state preservation.

Hopefully that helps a little.

neonquill:
For the record, it seems likely that this is not a good fix for M23 support. I'm happy to have this PR closed and I can either open an issue or leave this alone until there's a better approach.

I'm going to leave this open for now on the off chance that I'm wrong about that assumption.

perlindgren:
Thanks, for the info, I was just doing a quick read, but I think you are correct, baseline is not supporting BASEPRI (in any meaningful manner). Hence we need to do interrupt masking. Your fix will indeed "work" but has a soundness issue as interrupts with vector index above 32 will not be filtered and can therefore run even when they should be blocked. Interrupts bindings are just "names" to RTIC and will be bound to vector positions later (at link time). 

The current implementation needs to be revised so that it sees the "set/clear mask" register as an array of `set/clear mask[index/32]`. It was a while since I implemented it, and @korken89 also did some changes in the code since the first implementation.

perlindgren:
The current implementation is quite efficient, the mask will be computed as constant, not sure if it will for an "array" like implementation though, but LLVM is remarkably good at optimizing, so maybe its possible. 

If it becomes too complex for LLVM to see through, maybe its better to revert to the global masking for the m23 (v8M base), its cheap regarding cycles but costly regarding blocking.

neonquill:
It looks like there's logic to protect against this, but the config flags are still messed up in my case: looks like `thumbv6m` and `thumbv7m` should be mutually exclusive, but I'm not preserving that behavior in my patch (both are unset). I'm probably going to change up the config flags and see if I can generate the panic. If I'm reading the code correctly that should preserve correct behavior as long as you don't use an interrupt with an index > 31. That might be more annoying than not working at all 🙂.

Research:

[src/export.rs:create_mask](https://github.com/rtic-rs/cortex-m-rtic/blob/981fa1fb30a70ec2416778b3e8daa491108deb41/src/export.rs)
```rs
        if shift > 31 {
            panic!("Generating masks for thumbv6 failed! Are you compiling for thumbv6 on an thumbv7 MCU?");
        }
```

I think shift is coming from [macros/src/codegen/shared_resources.rs](https://github.com/rtic-rs/cortex-m-rtic/blob/981fa1fb30a70ec2416778b3e8daa491108deb41/macros/src/codegen/shared_resources.rs#L143)
```rs
        v.push(quote!(#device::Interrupt::#name as u32));
```

which in my case is probably coming from [here](https://github.com/neonquill/saml10-test-rs/blob/1e2c679b22fbae2d4dd2179504d98e5394b0650f/atsaml10e16a/src/lib.rs#L189)

But, the issue is that `create_mask` has another version for `#[cfg(not(armv6m))]` that just always returns zero, which I suspect is the version I'm ending up with.

neonquill:
Here's an attempt to rework the config flags. This version appears to correctly detect using an interrupt handler with index > 31 at compile time. If I stay <= 31 then everything compiles without errors. I kind of hacked this together as more of a proof of concept, if there's any desire to merge this I'll spend more time cleaning it up.

Errors with interrupts > 31:
```
336 | ...   panic!("Generating masks for thumbv6/thumbv8m.base failed! Are you compiling for thumbv6 on an thumbv7 MCU or using an unsupported thumbv8m.base MCU?...
    |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |       |
    |       the evaluated program panicked at 'Generating masks for thumbv6/thumbv8m.base failed! Are you compiling for thumbv6 on an thumbv7 MCU or using an unsupported thumbv8m.base MCU?', /Users/dwatson/work/tools/cortex-m-rtic/src/export.rs:336:13
    |       inside `create_mask::<2_usize>` at /Users/dwatson/.rustup/toolchains/1.61.0-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/panic.rs:57:9
...
260 | #[rtic::app(device = atsaml10e16a)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'An interrupt above value 31 is used while in armv6 or armv8m.base', fridge_sensor/src/main.rs:260:1
```

perlindgren:
Yea, you are on the right track.
Next thing up would be to implement the masking for `.base`, and look at the generated assembly.

neonquill:
I think I have something that appears to work. No clue if it's good or not, but tests appear to pass.

Testing with this sample code:
```rs
#![no_std]
#![no_main]

use panic_halt as _;

#[rtic::app(device = atsaml10e16a)]
mod app {

    #[shared]
    struct Shared {
        count: u32,
    }

    #[local]
    struct Local {}

    #[init]
    fn init(_cx: init::Context) -> (Shared, Local, init::Monotonics) {
        let count = 0;

        (Shared { count }, Local {}, init::Monotonics())
    }

    #[idle(shared = [count])]
    fn idle(_cx: idle::Context) -> ! {
        loop {}
    }

    #[task(binds = PTC, shared = [count], priority = 1)]
    fn ptc(mut cx: ptc::Context) {
        cx.shared.count.lock(|count| *count += 1);
    }

    #[task(binds = TC0, shared = [count], priority = 2)]
    fn tc0(mut cx: tc0::Context) {
        cx.shared.count.lock(|count| *count += 1);
    }
}
```
Attached  are the raw assembly output generated using `cargo objdump --release -- -S --no-leading-addr` corresponding to the two commits I just pushed. The first (ddd3d3) uses a fixed mask size of `[u32; 2]` the second version (e34fed) dynamically calculates the size based on the used interrupts but should also end up using `[u32; 2]` (PTC is 42, TC0 is 34):
[ddd3d3.txt](https://github.com/rtic-rs/cortex-m-rtic/files/9059615/ddd3d3.txt)
[e34fed.txt](https://github.com/rtic-rs/cortex-m-rtic/files/9059620/e34fed.txt)

I believe the relevant sections are (ddd3d3):
```asm
<PTC>:
; #[rtic::app(device = atsaml10e16a)]
 4e f2 80 10            movw    r0, #57728
 ce f2 00 00            movt    r0, #57344
 00 21                  movs    r1, #0
 01 60                  str     r1, [r0]
 41 60                  str     r1, [r0, #4]
;         cx.shared.count.lock(|count| *count += 1);
```

and from (e34fed):
```asm
<PTC>:
; #[rtic::app(device = atsaml10e16a)]
 4e f2 80 10            movw    r0, #57728
 ce f2 00 00            movt    r0, #57344
 00 21                  movs    r1, #0
 01 60                  str     r1, [r0]
 04 22                  movs    r2, #4
 42 60                  str     r2, [r0, #4]
;         cx.shared.count.lock(|count| *count += 1);
```

but I could be completely wrong about that! Interpreting ARM assembly is not something I can do without a lot of looking things up, so that will be tomorrow's task 🙂

neonquill:
(I'm hoping these incremental updates aren't annoying, if so, please just let me know and I'll hold off until I think I'm done).

I figured out enough of the assembly to spot a potential optimization and feel relatively confident this approach is working reasonably well. Next step from my perspective is to clean up the code, add some comments, and maybe some new unit tests.

More testing:

I compared the assembly for a M0+ chip with a stock 1.1.3 release of rtic with my latest ([2be8bfd](https://github.com/rtic-rs/cortex-m-rtic/pull/652/commits/2be8bfdda5537ebef1916dc3a4e377b60d7a5c94)) version of the code:

```diff
--- armv6m_rtic_1.1.3.asm	2022-07-07 21:15:14.000000000 -0400
+++ armv6m_2be8bf.asm	2022-07-07 21:28:54.000000000 -0400
@@ -98,7 +98,7 @@
  04 20                 	movs	r0, #4
  10 60                 	str	r0, [r2]
 ; #[rtic::app(device = efm32zg222)]
- 00 f0 05 f8           	bl	0x13c <test::app::rtic_ext::main::__rtic_init_resources::h663fb9b84d6fed13> @ imm = #10
+ 00 f0 05 f8           	bl	0x13c <test::app::rtic_ext::main::__rtic_init_resources::h2f286d05035dfd3a> @ imm = #10
 ;         loop {}
  fe e7                 	b	0x132 <main+0x36>       @ imm = #-4

@@ -106,7 +106,7 @@
      134:	00 e4 00 e0	.word	0xe000e400
      138:	00 e1 00 e0	.word	0xe000e100

-<test::app::rtic_ext::main::__rtic_init_resources::h663fb9b84d6fed13>:
+<test::app::rtic_ext::main::__rtic_init_resources::h2f286d05035dfd3a>:
 ; #[rtic::app(device = efm32zg222)]
  80 b5                 	push	{r7, lr}
  00 af                 	add	r7, sp, #0
```

The fact that I'm getting identical code for thumbv6m makes me optimistic that the approach of using an array of 32 bit masks will get optimized out in cases of only needing 1 chunk of masks.

I also dug more into the assembly on thumbv8m.base. As far as I can tell, I'm adding extra instructions to deal with the fact that I now need to store more values in ISER and ICER which seems unavoidable. I did add an optimization to not write to those two registers if the value was zero. That was happening in my test code because the interupts I am using are both > 31.

Reference assembly files:
 - [armv6m_2be8bf.asm.txt](https://github.com/rtic-rs/cortex-m-rtic/files/9068256/armv6m_2be8bf.asm.txt) M0+ with commit 2be8bf of rtic.
 - [armv6m_rtic_1.1.3.asm.txt](https://github.com/rtic-rs/cortex-m-rtic/files/9068258/armv6m_rtic_1.1.3.asm.txt) M0+ with stock 1.1.3 rtic.
 - [2be8bf.asm.txt](https://github.com/rtic-rs/cortex-m-rtic/files/9068260/2be8bf.asm.txt) M23 with commit 2be8bf of rtic.

neonquill:
I added a few comments and added a panic inside of build.rs to protect against future unknown targets.

I tried to see if I could write some tests against `thumbv8m.base` but ran into some issues with dependent crates. I'm going to stop poking at things for now. Let me know if there's anything else I can do.

neonquill:
Tests are failing because of:
> thread 'main' panicked at 'Unknown target 'x86_64-unknown-linux-gnu'. Need to update BASEPRI logic in build.rs.', build.rs:19:9

This is totally my fault, I saw it when i ran the tests locally and somehow dismissed it as something unrelated to my changes.  I'll work on finding a better solution.

neonquill:
Excluding targets that don't start with `thumb` seems to work better for me.

I still run into issues when running on macos, but every test I tried worked fine inside a linux vm.

AfoHT (approved):
bors merge

Looks great! Thank you for not just finding the issue but fixing it too!

PR #653, opened 2022-07-06
Allow custom `link_section` attributes for late resources

vccggorski:
This commit makes RTIC aware of user-provided `link_section` attributes,
letting user override default section mapping.


AfoHT (approved):
bors merge

PR #661, opened 2022-09-28
Fix new lint in the compiler

korken89:


AfoHT (dismissed):
bors merge

AfoHT (approved):
bors merge

PR #666, opened 2022-11-23
Modular basic

01joja:
Have implemented software codegen and hardware codegen


01joja:
Made a mistake sorry.

PR #669, opened 2022-12-15
CI: Run rustup and cargo directly

AfoHT:
actions-rs seems abandoned: See [link](https://github.com/actions-rs/toolchain/issues/216)

As GHA bundles rustup using it directly is trivial

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,7 +58,7 @@ jobs:
         run: sed -i 's,//deny_warnings_placeholder_for_ci,#![deny(warnings)],' src/lib.rs macros/src/lib.rs
 
       - name: Cache Dependencies
-        uses: Swatinem/rust-cache@v1
+        uses: Swatinem/rust-cache@v2
 
       - name: cargo check
         run: cargo check --target=${{ matrix.target }}
@@ -75,7 +75,7 @@ jobs:
         run: sed -i 's,//deny_warnings_placeholder_for_ci,#![deny(warnings)],' src/lib.rs macros/src/lib.rs
 
       - name: Cache Dependencies
-        uses: Swatinem/rust-cache@v1
+        uses: Swatinem/rust-cache@v2
 
       - name: cargo clippy
         run: clippy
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -21,22 +21,14 @@ jobs:
...
-          profile: minimal
-          toolchain: stable
-          override: true
-          components: rustfmt
-
       - name: Fail on warnings
         run: sed -i 's,//deny_warnings_placeholder_for_ci,#![deny(warnings)],' src/lib.rs macros/src/lib.rs
 
+      - name: Add Rust component
+        run: rustup component add rustfmt
```
rursprung, line 28: is this even needed? i think this component is present by default (at least i'm using `cargo fmt` in my GH CI builds w/o installing it first). previously you used `profile: minimal` which might've led to rustfmt not being present, but the default image has the full rust toolchain installed.

AfoHT: I've set the profile to minimal on the other jobs, but "missed" it here, maybe that's another way of doing it, just not doing minimal and then not having to add the component instead.

If profile is `default`, adding an already existing component is a no-op? Will check the log

AfoHT: Looks like a no-op, just states it's up to date and takes 0 seconds :D

.github/workflows/build.yml
```diff
@@ -21,22 +21,14 @@ jobs:
...
-        with:
-          profile: minimal
-          toolchain: stable
-          override: true
-          components: rustfmt
-
       - name: Fail on warnings
         run: sed -i 's,//deny_warnings_placeholder_for_ci,#![deny(warnings)],' src/lib.rs macros/src/lib.rs
 
+      - name: Add Rust component
```
AfoHT, line 27:
```suggestion

```

.github/workflows/build.yml
AfoHT:
```suggestion
```

AfoHT: Got the opportunity to try pushing the "suggestions" to a new level, noted they now support batching suggestions together :)

Thank you for pointing out the redundant stuff :+1:

korken89 (approved):
bors merge

PR #675, opened 2022-12-28
enable GitHub Dependabot

rursprung:
this ensures that the dependencies are kept up to date. see [the docs][] for further information.

[the docs]: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "cargo"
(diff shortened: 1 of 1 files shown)


rursprung:
> Could you also add for the `macro` folder?

that shouldn't be needed as they're part of the workspace and as far as i could observe dependabot handles workspaces (i hope this also works here as the setup here is a bit special with the main `Cargo.toml` being both a workspace definition and a normal project in one...):
https://github.com/rtic-rs/cortex-m-rtic/blob/4d7aeaf6c48e323509b0cb44446e032e91ac4478/Cargo.toml#L49-L53

AfoHT (approved):
bors merge

PR #690, opened 2023-01-22
NVIC prio bits must be in the range 1..=255: Handled by logical2hw()

### LINKED ISSUE #687: Maximum Task Priority
The [book](https://rtic.rs/1/book/en/by-example/app_priorities.html?highlight=priori#priorities) states, that task priorities are in range `1..=(1 << NVIC_PRIO_BITS)`, where `NVIC_PRIO_BITS` could be up to `8` bits. The logical priority of the monotonic task however [is set to](https://github.com/rtic-rs/cortex-m-rtic/blob/master/macros/src/codegen/pre_init.rs#L101) `1 << NVIC_PRIO_BITS`.  In case of `NVIC_PRIO_BITS == 8`, this would lead to the logical priority being `256`, i.e. 1 unit higher than allowed and an overflow would happen. This would make the timer the least prioritized interrupt.

I think the issue is not very urgent, since (to my limited knowledge) cortex-m devices normally do not offer 8 priority bits. However, I stumbled upon this when I was porting RTIC to the RISC-V processor architecture (with the CLIC extension), that offers 8 priority bits.

Best,
Noah

AfoHT:
Fixes #687

### DIFF
--- a/ui/task-priority-too-high.rs
+++ b/ui/task-priority-too-high.rs
@@ -9,7 +9,7 @@ mod app {
     struct Local {}
 
     #[init]
-    fn init(cx: init::Context) -> (Shared, Local, init::Monotonics) {
+    fn init(_: init::Context) -> (Shared, Local, init::Monotonics) {
         (Shared {}, Local {}, init::Monotonics())
     }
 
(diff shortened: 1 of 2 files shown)


korken89 (approved):
bors merge

PR #691, opened 2023-01-22
Basic cfg support, kind of, for Monotonics

### LINKED ISSUE #664: Support for #[cfg] on #[monotonic]
I would like to feature-gate the monotonic type to only have it in some builds and not others. I tried the following but it doesn't work:

```rust
    #[cfg(feature = "log")]
    #[monotonic(binds = SysTick, default = true)]
    type MonoSysTick = Systick<100>;
```

Then in `init` I have:

```rust
        (
            Shared { ... },
            Local { ... },
            #[cfg(not(feature = "log"))]
            init::Monotonics(),
            #[cfg(feature = "log")]
            init::Monotonics(systick),
        )

```

Here's what I get when building without the feature (it works with the feature):

```
error[E0412]: cannot find type `MonoSysTick` in module `super::super`
  --> src/main.rs:51:10
   |
51 |     type MonoSysTick = Systick<100>;
   |          ^^^^^^^^^^^ not found in `super::super`

error[E0412]: cannot find type `Systick` in this scope
  --> src/main.rs:51:24
   |
51 |     type MonoSysTick = Systick<100>;
   |                        ^^^^^^^ not found in this scope

error[E0061]: this struct takes 1 argument but 0 arguments were supplied
   --> src/main.rs:117:13
    |
117 |             init::Monotonics(),
    |             ^^^^^^^^^^^^^^^^-- an argument is missing
    |
```

Is this something that might eventually get support for?

AfoHT:
- Enable at least masking out a Monotonic
- Add example cfg-ing a Monotonic, showing limitations imposed by rtic-syntax
- Update changelog

The use case detailed in linked issue seems to be covered: Fixes #664

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ### Added
 
+- CFG: Slightly improved support for #[cfg] on Monotonics
 - CI: Check examples also for thumbv8.{base,main}
 - Allow custom `link_section` attributes for late resources
 
--- a/examples/cfg-monotonic.rs
+++ b/examples/cfg-monotonic.rs
@@ -1,0 +1,121 @@
+//! examples/cfg-monotonic.rs
+
+#![deny(unsafe_code)]
+#![deny(warnings)]
+#![deny(missing_docs)]
+#![no_main]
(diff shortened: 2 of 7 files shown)


korken89 (approved):
bors merge

PR #703, opened 2023-03-15
fix(rtic-sync): used fully qualified paths in Channel Macro

Yandrik:
Other imports with the name `Channel` (e.g. from Embassy) broke this macro. Now they don't.

### DIFF
--- a/rtic-monotonics/src/rp2040.rs
+++ b/rtic-monotonics/src/rp2040.rs
@@ -144,12 +144,12 @@ macro_rules! make_rp2040_monotonic_handler {
         #[no_mangle]
         #[allow(non_snake_case)]
         unsafe extern "C" fn TIMER_IRQ_0() {
-            rtic_monotonics::rp2040::Timer::__tq().on_monotonic_interrupt();
+            $crate::rp2040::Timer::__tq().on_monotonic_interrupt();
         }
 
         pub struct Rp2040Token;
 
-        unsafe impl rtic_monotonics::InterruptToken<rtic_monotonics::rp2040::Timer>
+        unsafe impl $crate::InterruptToken<rtic_monotonics::rp2040::Timer>
             for Rp2040Token
         {
         }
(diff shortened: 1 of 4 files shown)


alexmoon:
If you use $crate instead of ::rtic_sync the macro will be robust to cases where users rename the crate in Cargo.toml.

korken89:
Good catch to use `$crate`, I had totally forgotten about that.

Yandrik:
$crate should now be used in all macros in RTIC
(edit: the `macro_rules!` ones at least)

rtic-monotonics/src/systick.rs
```diff
@@ -162,12 +162,12 @@ macro_rules! make_systick_handler {
...
         #[allow(non_snake_case)]
         unsafe extern "C" fn SysTick() {
-            rtic_monotonics::systick::Systick::__tq().on_monotonic_interrupt();
+            $crate::systick::Systick::__tq().on_monotonic_interrupt();
         }
 
         pub struct SystickToken;
 
-        unsafe impl rtic_monotonics::InterruptToken<rtic_monotonics::systick::Systick>
+        unsafe impl $crate::InterruptToken<rtic_monotonics::systick::Systick>
```
korken89: Missed an `rtic_monotonics` here

rtic-monotonics/src/rp2040.rs
```diff
@@ -144,12 +144,12 @@ macro_rules! make_rp2040_monotonic_handler {
...
         #[allow(non_snake_case)]
         unsafe extern "C" fn TIMER_IRQ_0() {
-            rtic_monotonics::rp2040::Timer::__tq().on_monotonic_interrupt();
+            $crate::rp2040::Timer::__tq().on_monotonic_interrupt();
         }
 
         pub struct Rp2040Token;
 
-        unsafe impl rtic_monotonics::InterruptToken<rtic_monotonics::rp2040::Timer>
+        unsafe impl $crate::InterruptToken<rtic_monotonics::rp2040::Timer>
```
korken89: Missed an `rtic_monotonics` here

korken89 (changes requested):

korken89 (approved):
Thanks for fixing!

bors merge

PR #706, opened 2023-03-15
ci/tuneverboselevel

AfoHT:
- xtask: Change default loglevel
- xtask: Fix cargo fmt logmsg
- ci: xtask default verbosity


korken89 (approved):
bors merge

PR #710, opened 2023-03-19
Start async tasks at lowest priority

korken89:


DusterTheFirst:
Would this not cause a problem with the idle task also being at priority 0, causing it to not be able to be preempted, as outlined in the book? 

> Notice: software task at zero priority cannot co-exist with the [idle] task. The reason is that idle is running as a non-returning Rust function at priority zero. Thus there would be no way for an executor at priority zero to give control to software tasks at the same priority.

https://rtic.rs/dev/book/en/by-example/software_tasks.html

korken89:
@DusterTheFirst It will not interfere, one simply can have async tasks at prio 0 OR an idle.
If you want to use idle, then all async tasks need a prio > 0.

AfoHT (approved):
bors merge

PR #712, opened 2023-03-22
book: fix link to example sources

ottojo:


AfoHT:
The build error seems unrelated to the change, we might have to pin the rust-nightly version

let's retry and see:

bors retry

AfoHT (approved):
bors merge

PR #721, opened 2023-04-01
Added nRF monotonics

korken89:
Testing completed

### DIFF
--- a/rtic-monotonics/src/nrf/rtc.rs
+++ b/rtic-monotonics/src/nrf/rtc.rs
@@ -5,19 +5,22 @@
 //! ```
 //! use rtic_monotonics::nrf::rtc::*;
 //!
-//! # async fn usage() {
-//! # let rtc = unsafe { core::mem::transmute(()) };
-//! // Generate the required token
-//! let token = rtic_monotonics::create_nrf_rtc0_monotonic_token!();
+//! fn init() {
+//!     # // This is normally provided by the selected PAC
+//!     # let rtc = unsafe { core::mem::transmute(()) };
+//!     // Generate the required token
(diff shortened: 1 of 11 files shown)


rtic-monotonics/src/nrf/rtc.rs
```diff
@@ -106,7 +106,10 @@ macro_rules! make_rtc {
 
                 $tq.initialize(Self {});
 
-                unsafe { pac::NVIC::unmask(Interrupt::$rtc) };
+                unsafe {
```
AfoHT, line 115: Maybe a good time to add an (un)safety

```
# Safety
blabla
```
comment (here and for other similar ones)?

perlindgren (approved):
Looks good.

bors merge

rtic-monotonics/src/nrf/rtc.rs
```diff
@@ -0,0 +1,241 @@
...
+            type Instant = fugit::TimerInstantU64<32_768>;
+            type Duration = fugit::TimerDurationU64<32_768>;
+
+            fn now() -> Self::Instant {
+                // In a critical section to not get a race between overflow updates and reading it
+                // and the flag here.
+                critical_section::with(|_| {
+                    let rtc = unsafe { &*$rtc::PTR };
+                    let cnt = rtc.counter.read().bits() as u64;
+                    let ovf = if Self::is_overflow() {
```
Dirbaio, line 193: if overflow happens right between these lines, it'll give the wrong result

```rust
let cnt = rtc.counter.read().bits() as u64;
// OVERFLOW HAPPENS HERE
let ovf = if Self::is_overflow() {
```

perlindgren: So, a fix might be to re-read the `cnt` value on overflow.

PR #728, opened 2023-04-10
Add test for timer queue & monotonic

datdenkikniet:
Add an `std`-runnable test fixture and a `Monotonic` implementation for testing the `TimerQueue`


korken89 (approved):
bors merge

PR #730, opened 2023-04-11
remove vis restriction for local and shared resources

andrewgazelka:


korken89 (changes requested):
Thanks for the PR! I see one UI test is now panicking the macro, could you have a look at fixing that?

rtic-macros/ui/shared-pub.stderr
korken89, line 1: Specifically this one

andrewgazelka: What action do you suggest I take? My PR has no impact on the error. 
https://github.com/rtic-rs/rtic/blob/89e060d0c1e7d34aa4751e10b4ea1ae3c727adf3/rtic-macros/src/syntax/parse/app.rs#L435-L436
If `x` were not public, it would have triggered a panic before my commit.

korken89: After taking a closer look at the test case, I'd just remove it. Lets not complicate it.

andrewgazelka: @korken89 fixed :)

korken89 (approved):
Looks good, thanks!

bors merge

PR #733, opened 2023-04-14
Adding support for the ESP32C3

onsdagens:
As discussed, this adds support for the ESP32C3.

It relies on forks of both the runtime and the hal for interrupt preemption, my  next goal is possibly getting those upstreamed as well.

### DIFF
--- a/rtic-macros/src/codegen/bindings/cortex.rs
+++ b/rtic-macros/src/codegen/bindings/cortex.rs
@@ -3,7 +3,7 @@ use crate::{
     codegen::util,
     syntax::{analyze::Analysis as SyntaxAnalysis, ast::App},
 };
-use proc_macro2::TokenStream as TokenStream2;
+use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::quote;
 use std::collections::HashSet;
 use syn::{parse, Attribute, Ident};
@@ -28,6 +28,10 @@ fn is_exception(name: &Ident) -> bool {
             | "PendSV"
             | "SysTick"
     )
+}
+pub fn interrupt_ident() -> Ident {
+    let span = Span::call_site();
+    Ident::new("interrupt", span)
 }
 
 #[cfg(feature = "cortex-m-source-masking")]
@@ -323,6 +327,14 @@ pub fn interrupt_exit(_app: &App, _analysis: &CodegenAnalysis) -> Vec<TokenStrea
     vec![]
 }
 
+pub fn async_entry(
+    _app: &App,
+    _analysis: &CodegenAnalysis,
+    _dispatcher_name: Ident,
+) -> Vec<TokenStream2> {
+    vec![]
+}
+
 pub fn async_prio_limit(app: &App, analysis: &CodegenAnalysis) -> Vec<TokenStream2> {
     let max = if let Some(max) = analysis.max_async_prio {
         quote!(#max)
@@ -337,4 +349,11 @@ pub fn async_prio_limit(app: &App, analysis: &CodegenAnalysis) -> Vec<TokenStrea
         #[no_mangle]
         static RTIC_ASYNC_MAX_LOGICAL_PRIO: u8 = #max;
     )]
+}
+pub fn handler_config(
+    _app: &App,
+    _analysis: &CodegenAnalysis,
+    _dispatcher_name: Ident,
+) -> Vec<TokenStream2> {
+    vec![]
 }
--- a/rtic-macros/src/codegen/util.rs
+++ b/rtic-macros/src/codegen/util.rs
@@ -3,13 +3,11 @@ use core::sync::atomic::{AtomicUsize, Ordering};
 use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::quote;
 use syn::{Ident, PatType};
+//hook the target specific interrupt_ident function
+pub use super::bindings::interrupt_ident;
 
 const RTIC_INTERNAL: &str = "__rtic_internal";
 
-pub fn interrupt_ident() -> Ident {
-    let span = Span::call_site();
-    Ident::new("interrupt", span)
-}
 
 /// Mark a name as internal
 pub fn mark_internal_name(name: &str) -> Ident {
--- a/rtic/Cargo.toml
+++ b/rtic/Cargo.toml
@@ -31,6 +31,8 @@ features = ["rtic-macros/test-template"]
 name = "rtic"
 
 [dependencies]
+esp32c3 = { version = "0.17.0", optional = true}
+riscv = {version = "0.10.1", optional = true}
 cortex-m = { version = "0.7.0", optional = true }
 bare-metal = "1.0.0"
 #portable-atomic = { version = "0.3.19" }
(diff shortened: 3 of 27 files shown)


onsdagens:
As a sidenote, this should generalize well to the other ESP RISC-V MCUs but i don't have any to play around with.

onsdagens:
@romancardenas has suggested splitting up export like here https://github.com/greenlsi/rtic/tree/master/rtic%2Fsrc%2Fexport for better modularity. I think for now it might be overkill, but depending on how many unique backends we expect to support, it could be a good idea.

romancardenas:
If you don't like all those submodule folders, we could also leave it as `export::riscv_common`, `export::riscv_esp32c3`... And forget about things like `export::riscv_backend::esp32c3`.

In any case, I do believe that we should move the feature-gated definitions of functions like `pend` from `rtic::export.rs` to a backend-specific module for easier maintenance and better separation of concerns.

onsdagens:
I've been able to move things around a bit, now this only relies on changes to the hal, the rt remains untouched. I've also fixed some weird errors regarding feature selection, which i'm not sure how they made it in there to begin with.

romancardenas:
@steewbsd and I have re-structured the export module of our fork so it has less folders. Have a look, maybe you prefer it that way.

rtic-macros/src/codegen/bindings/cortex.rs
```diff
@@ -323,6 +323,14 @@ pub fn interrupt_exit(_app: &App, _analysis: &CodegenAnalysis) -> Vec<TokenStrea
     vec![]
 }
 
+pub fn async_entry(
```
romancardenas, line 330: I guess this new hook is required by ESP32C3 for un-pending the dispatchers' interrupt, right?

rtic-macros/src/codegen/main.rs
```diff
@@ -20,7 +20,7 @@ pub fn codegen(app: &App, analysis: &Analysis) -> TokenStream2 {
         quote!(#dispatcher();)
     } else {
         quote!(loop {
-            rtic::export::nop()
+            continue;
```
romancardenas: Just curious: why do you avoid the `nop` function?

rtic-macros/src/codegen/util.rs
```diff
@@ -6,11 +6,20 @@ use syn::{Ident, PatType};
 
 const RTIC_INTERNAL: &str = "__rtic_internal";
 
+//this is named something else in the esp32c3 pac
+#[cfg(not(feature = "riscv-esp32c3"))]
```
romancardenas: Check [this](https://github.com/greenlsi/rtic/blob/6f1f6605bf2ac03a238c003b7322c16dfe6c3ddb/rtic-macros/src/codegen/bindings.rs#L31). We had a similar issue and tackled this way. The idea is to avoid feature gates out of `codegen::bindings`.

rtic/Cargo.toml
```diff
@@ -31,6 +31,10 @@ features = ["rtic-macros/test-template"]
 name = "rtic"
 
 [dependencies]
+esp32c3 = {version = "0.13.0", optional = true}
+#next goal is getting the neccessary changes to the hal upstreamed
+#esp32c3-hal = { git = "https://github.com/onsdagens/esp-hal", branch="scaled_back", optional = true}
+esp32c3-hal = { git = "https://github.com/onsdagens/esp-hal", branch="interrupt_preemption", optional = true}
```
romancardenas: I see you depend both on the PAC and the HAL. Could you explain to me why? Maybe we need this as well for the SLIC implementation

rtic/src/export.rs
```diff
@@ -47,6 +50,65 @@ where
     NVIC::pend(interrupt);
 }
 
+#[cfg(feature = "riscv-esp32c3")]
+mod riscv_esp32c3;
+#[cfg(feature = "riscv-esp32c3")]
+pub use riscv_esp32c3::*;
+
+///I think all of these "pends" and "unpends"  should be moved to /export/<device>.rs
```
romancardenas: Same. We propose [this](https://github.com/greenlsi/rtic/tree/6f1f6605bf2ac03a238c003b7322c16dfe6c3ddb/rtic/src/export) structure.

rtic/src/export/riscv_esp32c3.rs
```diff
@@ -0,0 +1,105 @@
+use esp32c3::INTERRUPT_CORE0; //priority threshold control
+pub use esp32c3::{Interrupt, Peripherals};
```
romancardenas, line 2: I see you set your core peripherals as the PAC's peripherals. So this is basically what we have done. Thus, the core peripherals and device peripherals will be exactly the same. This is why I wonder if the core peripherals are needed for RISC-V targets, as there is no such thing. Maybe we can make it optional and include it only for Cortex. However, I think @korken89 would rather not modify struct fields depending on the backend.

romancardenas (commented):
Nice piece of work! I see we more or less reached the same point. Let's try to clarify the structure stuff and then I will adapt our fork to include yours :D

rtic-macros/src/codegen/bindings/cortex.rs
onsdagens: Yep

rtic-macros/src/codegen/main.rs
onsdagens: It feels like it achieves the same end result (this is just an empty idle generated when you don't give one), without the need for exporting nop. I guess there is no specific reason, just the solution i came up with.

rtic/src/export.rs
onsdagens: Yep, i'll align this with your proposal

rtic-macros/src/codegen/util.rs
onsdagens: I'll get back to this in like an hour, need to get moving

rtic-macros/src/codegen/main.rs
romancardenas: If this is effectively equivalent, then I'm in

romancardenas (dismissed):
LGTM.

As a side note, it would be worth discussing removing from the API `rtic::export::peripherals` and `rtic::export::nop` at the next meeting. The prior looks too Cortex-M-driven, and the latter can just be `continue`.

onsdagens:
Some breaking changes if you've been using this fork. I've moved it to use the new upcoming ``direct-vectoring`` feature in the esp-riscv-rt. This changes the syntax of the interrupt handlers slightly, and takes advantage of the interrupt vector table, improving task dispatch latency dramatically.

If you're using the ``esp32c3-hal``, you must enable the ``interrupt-preemption`` and  ``direct-vectoring`` features, and make sure you're using the latest version of the crate from the ``esp-hal`` repo. Similarly, if you're only running ``esp-riscv-rt``, you must enable those same features.

For some minimal examples check out https://github.com/perlindgren/esp32c3-test/tree/rtic_sw_tasks

onsdagens:
Rebased on upstream now with Per adding a drop-in replacement of atomics in `rtic-common` and `rtic-sync` with `portable-atomic` to move us away from the `imac` target and the shaky emulation it requires. There is also a new, less PoC, example in https://github.com/perlindgren/esp32c3-test/tree/rtic_sw_tasks using the `UsbSerialJtag` peripheral to communicate with the host, and a more advanced example here https://github.com/perlindgren/esp32c3-rtic-tau , with more things to come.

perlindgren:
Manually testing for toxic effects of `portable_atomic`.

```rust
extern "C" fn test_atomics(offset: u8) -> AtomicU8 {
    (84 + offset).into()
}
```

```
42000174 <test_atomics>:
42000174: 13 05 45 05  	addi	a0, a0, 84
42000178: 13 75 f5 0f  	andi	a0, a0, 255
4200017c: 82 80            	ret
```

It's not a complete test of course, but it looks as expected (compiled in release mode, without overflow checks).

There is no reason to believe any other natively supported atomic data structures should cause problems.

rtic/src/export/riscv_esp32c3.rs
korken89: I don't think that API is set in stone - if there is a good reason to deviate I see no reason to strictly follow the old one :)

korken89 (dismissed):
LGTM, thank you for the effort!

korken89 (approved):

PR #745, opened 2023-04-23
Check rtic-monotonics with eh-async enabled

datdenkikniet:
@korken89 discovered that the embedded-hal-async stuff had been broken for a longer time. This should've been caught by CI, but we don't test it with this feature enabled. This PR fixes that.


korken89 (approved):
bors merge

PR #752, opened 2023-05-17
relax rtic-time generics

burrbull:


datdenkikniet:
I think this is a pretty good idea.

For clarity, it may be good to rename the `TimerQueue` to `PriorityQueue` (since that is what it becomes with this relaxation), and add a `pub type TimerQueue<M> = PriorityQueue<<M as Monotonic>::Instant>` to indicate what the types' respective goals are, and what they' re bound to. Not sure if it is feasible or any more/less usable in that case, though

rtic-time/src/lib.rs
```diff
@@ -77,25 +77,25 @@ pub struct TimeoutError;
...
         &mut *self.0
     }
 }
 
-unsafe impl<Mono: Monotonic> Send for LinkPtr<Mono> {}
-unsafe impl<Mono: Monotonic> Sync for LinkPtr<Mono> {}
+unsafe impl<Instant: Copy + Ord> Send for LinkPtr<Instant> {}
+unsafe impl<Instant: Copy + Ord> Sync for LinkPtr<Instant> {}
 
-impl<Mono: Monotonic> TimerQueue<Mono> {
```
datdenkikniet, line 98: I think we should rename `TimerQueue` to `PriorityQueue` and add a type alias similar to `type TimerQueue<T> = PriorityQueue<<T as Monotonic>::Instant>` so that we can maintain the ergonomics of using the timer queue

datdenkikniet (changes requested):

Finomnis:
Might collide with and/or become obsolete through #874.

PR #753, opened 2023-05-20
Use xtasks for book

datdenkikniet:


datdenkikniet:
Too old :) Will do smaller incremental changes to hopefully land this sometime in the future.

PR #759, opened 2023-05-31
ci/releasetuning

AfoHT:
- CI: Cargo doc for both pre and post v2
- CI: Make it clearer why a branch push was omitted
- ci: Use job outputs


korken89 (approved):
bors merge

PR #768, opened 2023-06-14
rtic-sync: Fix possible UB in make_channel!

### LINKED ISSUE #763: rtic-sync: make_channel is unsafe
```rust
/// Creates a split channel with `'static` lifetime.
#[macro_export]
macro_rules! make_channel {
    ($type:path, $size:expr) => {{
        static mut CHANNEL: $crate::channel::Channel<$type, $size> =
            $crate::channel::Channel::new();

        // SAFETY: This is safe as we hide the static mut from others to access it.
        // Only this point is where the mutable access happens.
        unsafe { CHANNEL.split() }
    }};
}
```

You can get multiple receivers to one (mpsc) channel using something like this:
```rust
use rtic_sync::channel::{Sender, Receiver};

fn main() {
    let (s1, r1) = test();
    let (s2, r2) = test();
}

fn test() -> (Sender<'static, i32, 1>, Receiver<'static, i32, 1>) {
    rtic_sync::make_channel!(i32, 1)
}
```

I suggest using [static_cell](https://docs.rs/static_cell/latest/static_cell/) or doing a manual runtime check inside the Channel impl, like `thingbuf`'s [StaticChannel](https://docs.rs/thingbuf/0.1.4/thingbuf/mpsc/struct.StaticChannel.html#method.split)

korken89:
Closes #763

### DIFF
--- a/rtic-sync/CHANGELOG.md
+++ b/rtic-sync/CHANGELOG.md
@@ -13,4 +13,10 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ### Fixed
 
+## [v1.0.1]
+
+### Fixed
+
+- `make_channel` could be UB
+
 ## [v1.0.0] - 2023-xx-xx
--- a/rtic-sync/Cargo.toml
+++ b/rtic-sync/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rtic-sync"
-version = "1.0.0"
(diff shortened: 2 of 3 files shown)


datdenkikniet (dismissed):

datdenkikniet:
bors r-

Because we may want to add some docs/a version number

datdenkikniet (approved):

PR #791, opened 2023-08-22
ritc-sync: Fix make_channel forcing dependency into user crate

nilfit:
Before, `make_channel` assumed that `critical_section` would be available in the namespace of the code that invoked the macro. Access `critical_section` through `rtic-sync` instead.


AfoHT (approved):
Nice 👍

And welcome back 😃

PR #792, opened 2023-08-25
Book: Graph in preface.md example

GHF:
The graph in the example has two tasks named `A` with different priorities, which seems like a typo.


AfoHT (approved):
👍

PR #796, opened 2023-09-03
Monotonic impl for stm32 based on `stm32-metapac`

andresv:
This PR uses [stm32-metapac](https://lib.rs/crates/stm32-metapac) to add support for STM32 64 bit monotonic timer.
It defines the same timers as [embassy-stm32](https://github.com/embassy-rs/embassy/blob/2e6f4237f2410aa18c9866a5a1a5ed1f3bec8a4e/embassy-stm32/build.rs#L103-L108) is using: tim2, tim3, tim4, tim5, tim12 and tim15.

# TODO
- [ ] [MAYBE] currently it does not take timer peripheral, so user can accidentally use the same timer later somewhere in the app
- [x] expose STM32 16bit timers as 64 bit monotonics
- [x] figure out how to get `NVIC_PRIO_BITS` for `crate::set_monotonic_prio(pac::NVIC_PRIO_BITS, pac::Interrupt::$timer);`
- [x] figure out how to deal with `stm32-metapac` feature flags - there are a lot of them: https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/Cargo.toml#L143
- [x] figure out how to use `make_timer!` based on `stm32-metapac` feature flags - not all timers are available for all STMs

# TEST
`TIM2`, `TIM3` and `TIM15` are tested on `stm32g081kb`


rtic-monotonics/src/stm32.rs
```diff
@@ -0,0 +1,216 @@
...
+                    r.set_cen(true);
+                });
+
+                $tq.initialize(Self {});
+
+                // SAFETY: We take full ownership of the peripheral and interrupt vector,
+                // plus we are not using any external shared resources so we won't impact
+                // basepri/source masking based critical sections.
+                unsafe {
+                    //crate::set_monotonic_prio(pac::NVIC_PRIO_BITS, pac::Interrupt::$timer);
```
andresv: @Dirbaio maybe you can tell me how to get `NVIC_PRIO_BITS` from `stm32-metapack`. It seems it is gated with `rt` feature flag, but when this was enabled then I got conflict with interrupts.

Dirbaio: hm I think I made it gated by `rt` because that's what other PACs do. It's true that enabling `rt` makes stm32-metapac generate the interrupt vectors etc which will conflict if your'e also using a `stm32-rs` PAC in the same project. 

I'm not sure if there's any downsides to always exposing it regardless of the `rt` feature, I don't think so? PRs welcome to always expose it.

Alternatively you can get it from [here](https://docs.embassy.dev/stm32-metapac/git/stm32wle4cb/metadata/struct.Metadata.html#structfield.nvic_priority_bits) but the metadata is intended to be used from `build.rs`, so maybe it's not what you're looking for.

andresv: I couldn't figure out if it is possible to use `stm32-metapac` with `rt` in `rtic-monotonics` and at the same time import `embassy-stm32` to your main app without getting interrupt vectors conflict

```toml
embassy-stm32 = { git = "https://github.com/embassy-rs/embassy.git", features = ["defmt", "stm32g081kb", "nightly", "unstable-traits", "unstable-pac", "exti"] }
rtic-monotonics = { path = "../rtic/rtic-monotonics", features = ["stm32g081kb"]}
```
So for now I am gonna use `METADATA.nvic_priority_bits.unwrap()` as you suggested.

rtic-monotonics/Cargo.toml
```diff
@@ -40,6 +41,14 @@ nrf5340-app-pac = { version = "0.12.2", optional = true }
 nrf5340-net-pac = { version = "0.12.2", optional = true }
 nrf9160-pac = { version = "0.12.2", optional = true }
 
+# STM32
+stm32-metapac = { version = "13.0.0", features = ["metadata"], optional = true }
+
+[build-dependencies]
+stm32-metapac = { version = "13.0.0", default-features = false, features = [
```
andresv: build.rs must be probably removed, stm32-metapac expects a stm32xx feature flag in order to build. So it cannot be used like this.

rtic-monotonics/Cargo.toml
```diff
@@ -40,6 +41,9 @@ nrf5340-app-pac = { version = "0.12.2", optional = true }
 nrf5340-net-pac = { version = "0.12.2", optional = true }
 nrf9160-pac = { version = "0.12.2", optional = true }
 
+# STM32
+stm32-metapac = { version = "13.0.0", features = ["metadata"], optional = true }
```
Dirbaio, line 45: note `metadata` is supposed to be used from `build.rs` only (`build-dependencies`), not from the main binary. It's a giant static with a lot of strings, I'm not sure if it's guaranteed to *not* be included in the final binary if all you use is nvic_prio_bits.

andresv:
I think adding timer peripheral to `start` function might be too restrictive. Currently only `stm32-metapac` is used as a dependency but if we want to move a timer peripheral then this depends on which kind of HAL is going to be used.
Current implementation would work with any STM32 HAL (embassy-stm32, stm32-hal2, etc).

rtic-monotonics/Cargo.toml
andresv: I seems it does not affect size. I removed metadata and setting priorities and checked with `cargo objcopy --bin app --release -- -O binary app.bin` before and after and it was the same size. Actually it was exactly the same size, I assumed it would be a little smaller because priority was not set. But did it twice and it indeed rebuilt and used correct thing.

rtic-monotonics/Cargo.toml
```diff
@@ -63,3 +67,1350 @@ nrf52840 = ["dep:cortex-m", "dep:nrf52840-pac", "dep:critical-section"]
 nrf5340-app = ["dep:cortex-m", "dep:nrf5340-app-pac", "dep:critical-section"]
 nrf5340-net = ["dep:cortex-m", "dep:nrf5340-net-pac", "dep:critical-section"]
 nrf9160 = ["dep:cortex-m", "dep:nrf9160-pac", "dep:critical-section"]
+
+# STM32 timers
+# Use as `features = ["stm32g081kb", "tim15"]`
+tim2 = []
```
korken89: Please prepend these with `stm32_` to not mix them with other impls in the future. :)

korken89 (commented):
Awesome work! :rocket: 
I only have one small comment, please have a look.

I don't have hardware to test thing today, I hope I can give it a try tomorrow!

rtic-monotonics/Cargo.toml
andresv: `features = ["stm32g081kb", "tim15"]` looked nicer than `features = ["stm32g081kb", "stm32_tim15"]` but sure there might be name clashes in the future, i'll fix it

andresv: done

korken89 (approved):
Thank you for the hard work! :clap: 

Initial testing on HW looks good from my side, so lets merge and do more PRs if we find issues.

PR #802, opened 2023-09-14
rtic-monotonics: Implement blocking DelayUs from embedded-hal 1

nilfit:


korken89 (approved):

burrbull:
Can it be generic implementation?

nilfit:
> Can it be generic implementation?

My first thought was `impl<T: Monotonic> DelayUs for T`, but that's not possible due to the orphan rule. A wrapper type `struct W(T)` could `impl <T: Monotonic> DelayUs for W<T>`, but that would complicate the API as you have to wrap the monotonic before using `DelayUs`. There is already precedent in `embedded_hal_async::delay::DelayUs` for repeating the `impl`.

PR #810, opened 2023-09-27
Removing `nop` from the generated empty idle

### LINKED ISSUE #747: Removing `nop` from `rtic::export`
As discussed in #733, re-exporting `nop` in `rtic::export` may be unnecessary, and something like `continue` would do the trick in most cases. Personally, I'd keep the *mandatory* elements of `rtic::export` to the minimum. If anyone needs by chance the `cortex_m::nop` instruction, I'd say that they use it directly from the `cortex-m` crate instead of RTIC's re-export.

onsdagens:
This addresses #747 , the empty ` loop{} ` turning into a UDF instruction has been fixed ( https://github.com/rust-lang/rust/issues/28728 ), so we can avoid exporting ` nop `. Unsure if this warrants a changelog entry.

### DIFF
--- a/rtic-macros/src/codegen/main.rs
+++ b/rtic-macros/src/codegen/main.rs
@@ -20,7 +20,6 @@ pub fn codegen(app: &App, analysis: &Analysis) -> TokenStream2 {
         quote!(#dispatcher();)
     } else {
         quote!(loop {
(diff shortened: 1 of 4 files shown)


AfoHT (approved):
👍

PR #813, opened 2023-10-03
rtic-monotonics: Fix stm32-metapac use

### LINKED ISSUE #803: STM32 rtic-monotonics APB register names incompatible
Hi,
I am trying to use most recent STM32 `rtic-monotonics` addition #796 with STM32H730VB and build fails, because `apb*` register names don't match expectations

https://github.com/rtic-rs/rtic/blob/54aec9b3989162a60534620b80cde78d2d5cc1ba/rtic-monotonics/src/stm32.rs#L300
needs to be changed to
```rust
enable_timer!(apb1lenr, set_tim3en, apb1lrstr, set_tim3rst);
```
and similarly for other timers.

nilfit:
Previously, the stm32 monotonics only compiled for some chip families (#803). For example, stm32g081kb worked, but not stm32f407*.

The stm32-metapac does not directly unify peripheral names between the many stm32 families, but provides tools for build scripts to generate code that uses the right names for the selected chip. Use that mechanism instead of targeting a specific family.

The build script is based on https://github.com/embassy-rs/embassy/blob/ad524373edf8dde6ff816a41c6dd3fedf1072ba6/embassy-stm32/build.rs and probably still contains some things that are not needed for rtic-monotonics.

### DIFF
--- a/rtic-monotonics/CHANGELOG.md
+++ b/rtic-monotonics/CHANGELOG.md
@@ -7,6 +7,10 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ## Unreleased
 
+### Fixed
+
+- Fix STM32 support for other chip families
+
 ## v1.2.0 - 2023-09-19
 
 ### Added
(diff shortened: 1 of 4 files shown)


AfoHT (approved):
👍

PR #815, opened 2023-10-06
RISC-V support over CLINT

### LINKED ISSUE #746: (partially) removing core peripherals from RTIC context
As discussed in the previous meeting, now that RTIC v2 targets multiple architectures, some items seem too biased toward Cortex-M microcontrollers.

One of these elements is `rtic::export::Peripherals`. For instance, the RISC-V architecture does not have a common set of peripherals, and the current successful ports (#733 or [greenlsi/rtic](https://github.com/greenlsi/rtic)) end up re-exporting the device peripherals twice.

The main options we discussed to tackle this issue are:
1. Leave `rtic::export::Peripherals` as `()` when not needed.
2. Feature-gate the core peripherals of the context so it is only active when it makes sense.
3. Completely remove core peripherals from RTIC.

It would be nice to have a discussion about this topic. Personally, I'm inclined to the third option, as we would keep a common interface for all the targets. However, it may break already working examples/projects.

### LINKED ISSUE #816: rtic-macros: add backend-specific AppArgs
As mentioned in #815 , it would be very useful to allow backends have additional configuration parameters. This is specially necessary for RISC-V, since there will be different interrupt controllers that, in multihart/multicore chips, would need extra configuration to specify which peripheral we want to use.

romancardenas:
The [Software-Level Interrupt Controller (SLIC)](https://github.com/romancardenas/riscv-slic) is a software interrupt controller that aims to mimic the PLIC but using a single interrupt source. It currently works for the CLINT peripheral, but I plan to add additional mechanisms so a broader amount of RISC-V targets can benefit from it.

This PR is a work-in-progress effort for adapting the SLIC to RTIC. I already modified some aspects:
- New `interrupt_mod` codegen function.
- New `extra_modules` codegen function.

But I'd like to do two additional major changes before this PR is accepted:
- Remove core peripherals from `rtic::export` (I currently use a very dirty patch to avoid issues with my implementation)
- Add a new `backend` field to the `AppArgs` struct that, depending on the selected backend, can potentially parse additional configuration parameters. 

Additionally, as the SLIC is able to generate any number of synthetic interrupt sources, I would like to modify the current implementation so users won't need to define their dispatchers. Instead, the dispatchers will be automatically generated as needed by the application.

All the feedback is more than welcome, let me know what do you think about my proposals :)


romancardenas:
I added backend-specific configuration arguments to the app macro.  This solves #816

rtic-macros/src/codegen/pre_init.rs
```diff
@@ -11,10 +11,16 @@ pub fn codegen(app: &App, analysis: &Analysis) -> Vec<TokenStream2> {
     // Disable interrupts -- `init` must run with interrupts disabled
     stmts.push(quote!(rtic::export::interrupt::disable();));
 
+    // TODO dirty patch, needs to be addressed
+    #[cfg(not(feature = "riscv-slic"))]
```
onsdagens: Maybe the following line could come from a new binding?

onsdagens (commented):
awesome, one small comment from a quick look

rtic-macros/src/codegen/pre_init.rs
romancardenas: That's a good idea. We can use bindings to partially remove the Peripherals struct depending on the backend.

romancardenas:
I followed the `peripherals` macro argument approach and added a `core` macro argument. This argument defaults to `true` in order to be backwards compatible. However, when set to `false`, RTIC will remove the core peripherals from its internals.

In case of my RISC-V SLIC, this field is mandatory to be false (again, I still need a pre-processing step in which I can manipulate the app args before generating all the code).

With this approach, removing core peripherals from RTIC is very similar to #720.

Related issue: #746

romancardenas:
I added the `pre_init_preprocessing` binding to allow backends to modify the `App` structure according to their needs before any other action is triggered.

You can see how the SLIC backend uses this binding [here](https://github.com/rtic-rs/rtic/blob/3920a8618c47e10d06c2d8b0f60769fcc218fe8d/rtic-macros/src/codegen/bindings/riscv_slic.rs#L66). In this case, I force the `core` argument to be `false`. I also check that the `dispatchers` are empty, because I later generate as much dispatchers as needed by the SLIC.

With all these changes I am happy with the final result. So it's time to discuss my proposal and fine-tune/reorder the code before merging it :) Let me know what you think.

romancardenas:
Last week I've been working on integrating use cases of RTIC on a [HiFive1](https://github.com/romancardenas/hifive1) RISC-V microcontroller in the CI. I adapted the CI of the repo to test examples for a RISC-V target using my new backend. The main changes are explained below:

- The `rtic` crate does not contain examples anymore.
- Now, `xtask` accepts a new parameter `platform` to select which [platform](https://github.com/rtic-rs/rtic/blob/fb20aeec1f43a47e25d9b29b3f7d271853bf581a/xtask/src/argument_parsing.rs#L219) we want to test. By default, it is set to `lm3s6965`.
- Each platform has custom [rust flags](https://github.com/rtic-rs/rtic/blob/fb20aeec1f43a47e25d9b29b3f7d271853bf581a/xtask/src/argument_parsing.rs#L245) and a [default backend](https://github.com/rtic-rs/rtic/blob/fb20aeec1f43a47e25d9b29b3f7d271853bf581a/xtask/src/argument_parsing.rs#L283). Additionally, the [`features`](https://github.com/rtic-rs/rtic/blob/fb20aeec1f43a47e25d9b29b3f7d271853bf581a/xtask/src/argument_parsing.rs#L298) method allows us to define which backends are valid for every platform and, if needed, which features must be enabled when building their examples.
- All the examples for a given platform are in the `examples/<platform>` crate. We can continue expanding the number of platforms.
- For QEMU runs, the `ci/expected/<platform>` directory will contain a file with the expected outcome of a given example. `xtask` will then check that the execution matches the expected outcome.

About usage examples builds, I'm not sure if these are longer needed, as we can now treat them as regular examples. Of course, we now should adapt all the examples to be emulated in QEMU when possible. In this way, the CI will be enriched, as it does not limit to a single platform.

perlindgren:
Small comment, RTIC typically should require no unsafe user code. For the Cortex and ESP32 targets setting up the interrups are done by the framework. This also ensures that the `init` is guaranteed to run to end in a "global critical section".

romancardenas:
@perlindgren I'll take a loop to ESP's implementation to see where is the proper place to enable the interrupts right after the init code finishes

onsdagens:
> @perlindgren I'll take a loop to ESP's implementation to see where is the proper place to enable the interrupts right after the init code finishes

the global interrupt enable is already there [here](https://github.com/rtic-rs/rtic/blob/efb82b7b05785352d0003b565f15e73ca4e656a1/rtic-macros/src/codegen/post_init.rs#L44), so it's probably just a matter of removing it from the example. We were also discussing offline the premature interrupt enable being a possible cause of some concurrency issues leading to the CI weirdness.

the ``slic.set_interrupts()`` can probably be done anywhere in the pre_init, my reasoning here being it shouldn't have any side effects with the interrupts disabled

romancardenas:
I fixed the unsafe code issue.

The weird thing is that the shared variable is correctly read and written from all the tasks but when it comes to idle, it reads the initial value as if the other tasks didn't run before

romancardenas:
Hey wait! The issue is that idle locks the shared counter before the medium task preempts it! That makes more sense. I'll take a look to find the bug.

(still weird that CI fails depending on which host system is running QEMU)

romancardenas:
QEMU version via apt is 6.2, while I'm using 8.2.0 on my machine. I'll try to build the latest version of QEMU from source and see if that was it.

romancardenas:
That was it! CI passes with QEMU 8.2.0

.github/workflows/build.yml
```diff
@@ -213,7 +247,59 @@ jobs:
...
+        run: | 
+          sudo apt-get -y install build-essential procps curl file git
+          NONINTERACTIVE=1 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+          (echo; echo 'eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"') >> /home/runner/.bashrc
+          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+          brew upgrade
+      - name: Install QEMU
+        run: |
+          eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+          brew install qemu
```
AfoHT: This can be replaced now by 
```
      - name: Download built QEMU
        uses: actions/download-artifact@v4
        with:
          name: qemu

      - name: Extract QEMU into local path
        run: tar -xf qemu.tar -C /usr/local/bin
```
:)

.github/workflows/build.yml
```diff
@@ -213,7 +247,59 @@ jobs:
           sudo apt install -y qemu-system-arm
 
       - name: Run-pass tests
-        run: cargo xtask --deny-warnings --backend ${{ matrix.backend }} qemu
+        run: cargo xtask --deny-warnings --platform lm3s6965 --backend ${{ matrix.backend }} qemu
+  
+  # Platform hifive1: verify the example output with run-pass tests
+  testexampleshifive1:
+    name: QEMU run (hifive1)
```
AfoHT, line 298:
```suggestion
    name: QEMU run (hifive1)
    needs: buildqemu
```

AfoHT (commented):
I have not yet gone through the actual code changes but looks really promising!

Another suggestion to make things easier for you and us later on when merging, I would recommend you rebase your branch on master rather than merging `rtic-rs:master` into this branch to match rtic repo workflow :)

image

What I had in mind is you make sure you've `git fetch` so you got latest `rtic-rs:master` locally, then ```git rebase -i where-you-have-upstream-rtic/master```

In my case that's my `upstream`
```
❯ git remote -v
origin  git@github.com:AfoHT/rtic.git (fetch)
origin  git@github.com:AfoHT/rtic.git (push)
upstream        git@github.com:rtic-rs/rtic.git (fetch)
upstream        git@github.com:rtic-rs/rtic.git (push)
```
So I did
```
git rebase -i upstream/master
```
Git will try to place all your commits of this PR on top/following the rtic:master branch, omitting any that already exists in the branch

There will be some conflicts on `rtic/Cargo.toml` and `.github/worflows/build.yml`

* fix inside the file/resolve
* `git add` the file
 when done `git rebase --continue`

I've pushed my attempt at that to https://github.com/AfoHT/rtic/tree/romancardenas/master if it somehow can help you. That branch also includes the "suggestions" related to QEMU, might be useful.

Git is a powerful  tool with so many ways of using it, my intention is not to question your abilities, rather to help you land this PR ASAP :D

romancardenas:
DW @AfoHT it's good to learn new tools :D I'll take a look ASAP and try to adapt the PR as you suggested

romancardenas:
@AfoHT I addressed all your comments and everything should work now. I think it is time to improve the xtask tool for multiple platforms and try to reduce duplicates in the build.yaml file. What do you think?

.github/workflows/build.yml
```diff
@@ -282,8 +316,80 @@ jobs:
...
+          name: qemu
+
+      - name: Extract QEMU into local path
+        run: tar -xf qemu.tar -C /usr/local/bin
+
+      - name: Check which QEMU is used
+        run: |
+          which qemu-system-arm
+          which qemu-system-riscv32
+
```
AfoHT:
```suggestion
```

Seems we happened to get double of these :)

rtic/build.rs
```diff
@@ -16,7 +16,7 @@ fn main() {
         //even within the same target
         //need some other way to discern
     } else if target.starts_with("riscv32i") {
-        println!("cargo:rustc-cfg=feature=\"riscv-esp32c3\"");
```
AfoHT, line 19: I'm sure you have some plan for these TODOs :)

AfoHT (commented):
Nice! The git commit log is much nicer without all the unrelated merges, thank you!

I have yet to play around with this a bit more locally, but apart from a tiny duplication in build.yaml and some TODOs you've saved yourself this is approaching master branch :D I don't have any proper devices, but that makes the QEMU really interesting, that we have that at least. But I know @korken89 had some devices?

I expect we might find things to iron out, but this can be done on master branch too.

onsdagens:
> Nice! The git commit log is much nicer without all the unrelated merges, thank you!
> 
> 
> 
> I have yet to play around with this a bit more locally, but apart from a tiny duplication in build.yaml and some TODOs you've saved yourself this is approaching master branch :D I don't have any proper devices, but that makes the QEMU really interesting, that we have that at least. But I know @korken89 had some devices?
> 
> 
> 
> I expect we might find things to iron out, but this can be done on master branch too.

I think we have a CLINT/PLIC core we can play around with on an FPGA @perlindgren .

rtic/build.rs
romancardenas: I'm not sure about the purpose of this `build.rs` file. Doesn't Cargo deal with all these feature flags?

romancardenas:
I don't know what changed, but now some tetsts fails because an enum in the `build.rs` file of `rtic-monotonic` is not used. I'll check it ASAP

rtic/build.rs
romancardenas: Ok, so I changed the approach in `build.rs`. Instead of analyzing the target, I check the active features and depending on the backend, I enable the additional feature.

romancardenas:
I pushed a new version of the `build.rs` for `rtic`. Now, I parse the backend selected by the user and enable the proper `rustc` flag. It should be more flexible now :)

@AfoHT let me know what you think or if you prefer any other approach.

PS: as the new way is backend-driven, we can get rid of all the test features 😁

romancardenas:
> I don't know what changed, but now some tetsts fails because an enum in the `build.rs` file of `rtic-monotonic` is not used. I'll check it ASAP

Looks like the issue was the new Nightly 1.78.0 version. Clippy is more clever now. I added some changes to allow dead code from now.

romancardenas:
I adapted some previous examples for lm3s6965 but for the e310x uC. They work as expected.

There is a lot of code duplicates though. We should re-think the current stack to avoid these issues. But I'd leave this task to a future PR, this is already quite big.

romancardenas:
closes #746, closes #816

rtic/src/export/riscv_common.rs
```diff
@@ -1,10 +1,10 @@
 /// GENERIC RE-EXPORTS: needed for all RTIC backends
-pub use riscv::interrupt;
+// pub use riscv::interrupt;
```
AfoHT: Can it be removed?

xtask/src/argument_parsing.rs
```diff
@@ -175,22 +178,28 @@ impl Backends {
...
-    }
+    // #[allow(clippy::wrong_self_convention)]
+    // pub fn to_rtic_uitest_feature(&self) -> &'static str {
+    //     match self {
+    //         Backends::Thumbv6 | Backends::Thumbv8Base => "rtic-uitestv6",
+    //         Backends::Thumbv7 | Backends::Thumbv8Main => "rtic-uitestv7",
+    //         Backends::RiscvEsp32C3 => "rtic-uitestesp32c3",
+    //         Backends::Riscv32ImcClint | Backends::Riscv32ImacClint => "rtic-uitestclint",
+    //     }
+    // }
```
AfoHT: Looks like this can be removed too :)

xtask/src/main.rs
```diff
@@ -221,12 +243,13 @@ fn main() -> anyhow::Result<()> {
...
         Commands::Test(args) => {
             info!("Running cargo test on backend: {backend:?}");
-            cargo_test(globals, &args, backend)
+            cargo_test(globals, args, backend)
         }
         Commands::Book(args) => {
             info!("Running mdbook");
             cargo_book(globals, &args.arguments)
         }
+        // TODO these two commands are not needed anymore
```
AfoHT: Great! Then I think you should remove them together with the associated "usage" functions etc.! :+1:

rtic-monotonics/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 ### Changed
 
 - Removed unused `rust-toolchain.toml`
+Tweak `build.rs` to avoid warnings in Nightly 1.78+
```
AfoHT: Nitpick: Add '-' and put latest entry at the top

```suggestion
- Tweak `build.rs` to avoid warnings in Nightly 1.78+
- Removed unused `rust-toolchain.toml`
```

rtic-macros/src/syntax/backend/cortex.rs
```diff
@@ -0,0 +1,16 @@
...
+};
+
+#[derive(Debug)]
+pub struct BackendArgs();
+
+impl Parse for BackendArgs {
+    fn parse(input: ParseStream) -> Result<Self> {
+        Err(Error::new(
+            input.span(),
+            "cortex backend does not accept any arguments",
```
AfoHT, line 13: Discussion point: the features exposed are `cortex-m-*`, and some modules are `cortex`, while others are `riscv_xyz`.

I wonder if it makes sense to keep to `cortex-m`/`cortex_m` here too for consistency? That includes module name etc.

rtic-macros/src/codegen/bindings/riscv_slic.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    if app.args.backend.is_none() {
+        return Err(parse::Error::new(
+            Span::call_site(),
+            "SLIC requires backend-specific configuration",
+        ));
+    }
+
+    Ok(()) // TODO
```
AfoHT: TODO?

rtic-macros/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 ### Added
 
 - RTIC v2 now works on stable.
+- Unstable support for RISC-V targets compatible with `riscv-slic`
```
AfoHT: This PR predates stable support, just it will land after :P Rebase does not care though

```suggestion
- Unstable support for RISC-V targets compatible with `riscv-slic`
- RTIC v2 now works on stable.
```

examples/hifive1/examples/prio_inheritance.rs
```diff
@@ -0,0 +1,141 @@
+#![no_main]
+#![no_std]
+// #![feature(type_alias_impl_trait)]
```
AfoHT: I think this applies to all the examples copied/touched before v2.1.0, can be cleaned

examples/hifive1/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+edition = "2021"
+
+[workspace]
+
+[dependencies]
+rtic = { path = "../../rtic" }
+heapless = "0.8.0"
+hifive1 = { git = "https://github.com/romancardenas/hifive1.git", features = ["board-redv"] }
+e310x = { git = "https://github.com/greenlsi/e310x.git", features = ["rt"]}
+# riscv-rt = {path = "../../../riscv/riscv-rt", features = ["single-hart"]}
```
AfoHT: TODO before publish

rtic/Cargo.toml
```diff
@@ -31,24 +31,22 @@ features = ["rtic-macros/test-template"]
...
 
 [dependencies]
+# riscv-slic = { path = "../../riscv-slic/riscv-slic", optional = true }
+riscv-slic = { git = "https://github.com/romancardenas/riscv-slic.git", rev = "2a91edb", optional = true }
 esp32c3 = { version = "0.20.0", optional = true}
 riscv = {version = "0.11.0", optional = true}
 cortex-m = { version = "0.7.0", optional = true }
 bare-metal = "1.0.0"
-#portable-atomic = { version = "0.3.19" }
+# portable-atomic = { version = "0.3.19" }
```
AfoHT, line 39: Not used at all, do we need it? :thinking: @korken89

AfoHT (commented):
This is looking great!

Love the cleanup of the examples structure, the platform makes it tidy and extensive!

Some comments and nitpicks here and there

Do you have a plan for `riscv-slic` crate? For long term I guess there will be some release we'll depend on, rather than the current git-rev dep :)

romancardenas:
> Do you have a plan for `riscv-slic` crate? For long term I guess there will be some release we'll depend on, rather than the current git-rev dep :)

I will ship a 0.1.0 release as soon as I get your OK to merge the PR :) I'm still waiting in case we detect any deficiency during the review process

rtic/src/export/riscv_common.rs
romancardenas: Done!

xtask/src/argument_parsing.rs
romancardenas: Done!

xtask/src/main.rs
romancardenas: Done (maybe I forgot something, a second check would be very useful :D)

rtic-monotonics/CHANGELOG.md
romancardenas: Done

rtic-macros/src/syntax/backend/cortex.rs
romancardenas: Good point. I guess that we should use `cortex-m` / `cortex_m`, as it follows the name of the architecture crate (for RISC-V, the crate is `riscv`).

rtic-macros/src/codegen/bindings/riscv_slic.rs
romancardenas: Ooops! removed

rtic-macros/CHANGELOG.md
romancardenas: Swapped

examples/hifive1/examples/prio_inheritance.rs
romancardenas: Cleaned

rtic/Cargo.toml
romancardenas: I think it is a "reminder" that we should move from atomic-polyfill to portable-atomic :D

examples/hifive1/Cargo.toml
romancardenas: The issue with the official e310x and hifive1 crates is that they do not follow the new riscv-peripheral crate fashion. It is something I have to start moving, but I'm now kind of busy with riscv-pac and adding support to vectored mode in riscv-rt

romancardenas: Once RTIC provides support to any RISC-V uC with a CLINT peripheral provided the PAC uses riscv-peripheral, maintainers will have a strong motivation to adopt the new crate 😁 

I also plan to look at svd2rust and contribute there to ease the port task.

romancardenas:
@AfoHT I addressed all your comments :)

AfoHT (approved):
Nice! 🚀 

I think we can have this on master to really start playing with it. 

@perlindgren you also had some devices to test on?

PR #817, opened 2023-10-09
rtic-sync: Add SPI bus sharing with arbiter

nilfit:
The pre-release embedded-hal-async 1.0.0-rc.1 has two traits for SPI: [`SpiBus`], which is typically implemented by HALs and [`SpiDevice`], which is implemented by pairing a chip select GPIO with a sharing mechanism. [`embedded-hal-bus`](https://docs.rs/embedded-hal-bus/0.1.0-rc.1/embedded_hal_bus/) provides implementations for some synchronization mechanisms.

Implement `SpiDevice` using `rtic_sync::arbiter::Arbiter`.

[`SpiBus`]: https://docs.rs/embedded-hal-async/1.0.0-rc.1/embedded_hal_async/spi/trait.SpiBus.html
[`SpiDevice`]: https://docs.rs/embedded-hal-async/1.0.0-rc.1/embedded_hal_async/spi/trait.SpiDevice.html

### DIFF
--- a/rtic-sync/CHANGELOG.md
+++ b/rtic-sync/CHANGELOG.md
@@ -9,6 +9,8 @@ For each category, _Added_, _Changed_, _Fixed_ add new entries at the top!
 
 ### Added
 
+- `arbiter::spi::ArbiterDevice` for sharing SPI buses using `embedded-hal-async`
+
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


korken89 (changes requested):
Thank you for the addition!

Can you please feature gate this under e.g. `embedded-hal-async` so it's not forced to be nightly only?

Dirbaio:
perhaps `unstable` since it'll be a breaking change to ugprade when embedded-hal 1.0.0 is out.

korken89 (approved):
Thank you for fixing!

PR #820, opened 2023-10-13
[book] Remove non-existent anchor reference

fwolter:
There is no anchor named `timeout_at` in the example source file. By looking at the context, the reference doesn't look like it's supposed to be there at all.


korken89 (approved):
Thanks for fixing!

PR #822, opened 2023-10-25
incr: making embedded-hal optional

### LINKED ISSUE #821: rtic-monotonics depends on embedded-hal-1.0.0-rc1
Hi,
I'm trying to use `rtic-monotonics` in my project. However, the newest release depends on `embedded-hal` version `1.0.0-rc1`, where many traits are removed as they are trying to ship a stable release. This creates dependency conflicts with most HAL implementations, as they depend on `embedded-hal` version `1.0.0-alpha.11` or lower. A practical solution is to simply downgrade. If you think this is a good idea, I'm happy to do a PR. Thanks!

explocion:
A simple way to fix #821

### DIFF
--- a/rtic-monotonics/Cargo.toml
+++ b/rtic-monotonics/Cargo.toml
@@ -4,12 +4,12 @@ version = "1.2.0"
 
 edition = "2021"
 authors = [
-    "The Real-Time Interrupt-driven Concurrency developers",
-    "Emil Fresk <emil.fresk@gmail.com>",
-    "Henrik Tjäder <henrik@tjaders.com>",
-    "Jorge Aparicio <jorge@japaric.io>",
-    "Per Lindgren <per.lindgren@ltu.se>",
-    "Andres Vahter <andres@vahter.me>",
+  "The Real-Time Interrupt-driven Concurrency developers",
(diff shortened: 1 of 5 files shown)


Dirbaio:
you can mix crates using `1.0.0-rc1` and `1.0.0-alpha.11` perfectly fine, they don't conflict at all. Rust considers them separate traits, so you can't use them together, but it still compiles. 

Making e-h optional doesn't help, because it also means you still can't use these crates together.

explocion:
> error: failed to select a version for `embedded-hal`.
    ... required by package `stm32f4xx-hal v0.17.1`
    ... which satisfies dependency `stm32f4xx-hal = "^0.17.1"`
versions that meet the requirements `=1.0.0-alpha.11` are: 1.0.0-alpha.11

> all possible versions conflict with previously selected packages.

 >previously selected package `embedded-hal v1.0.0-rc.1`
    ... which satisfies dependency `embedded-hal = "^1.0.0-rc.1"` of package `rtic-monotonics v1.2.0`
    ... which satisfies dependency `rtic-monotonics = "^1.2.0"`

> failed to select a version for `embedded-hal` which could resolve this conflict

I guess the cargo cannot resolve this by default.

PR #833, opened 2023-11-08
rtic-monotonics: Add documentation for i.MX RT and other doc fixes

Finomnis:


rtic-monotonics/src/systick.rs
```diff
@@ -5,7 +5,7 @@
 //!
 //! | Feature          | Tick rate | Precision |
 //! |:----------------:|----------:|----------:|
-//! | (none / default) |  1 Hz     |      1 ms |
+//! | (none / default) |  1 Hz     |       1 s |
```
korken89: This should say 1 kHz, 1 ms.

korken89 (changes requested):
Thanks for the PR! There is a small error that needs fixing, after that I'll merge :D

rtic-monotonics/src/systick.rs
Finomnis: My bad, I corrected it in the wrong direction.

Finomnis:
@korken89 fixed :)

korken89 (approved):

Finomnis:
@korken89 Any rough idea on when 1.3.0 will be released to crates.io?

PR #841, opened 2023-11-24
rtic-sync: add Arbiter for I2C

andresv:
Adds `Arbiter` for I2C to easily share I2C bus with many I2C sensor drivers.
Usage example is also added to rustdoc.


korken89 (approved):
Thanks for adding!

andresv:
https://github.com/rtic-rs/rtic/pull/845

PR #843, opened 2023-11-27
Fix mono delay

### LINKED ISSUE #844: Monotonic delays are too short
Given a monotonic tick rate of `1ms` and we call `DelayUs::delay_us(1200)` on the monotonic.
- *Desired behavior:* wait between `2ms` and `3ms`
- *Current behavior:* wait between `0ms` and `1ms` (which isn't even correct sometimes)

Note that there is a discrepancy of `2ms` (or rather: two timer ticks) between desired and current behavior.
- `1ms` comes from the fact that we need to compensate for timer uncertainty
- `1ms` comes from the fact that we use `.micros()` instead of `.micros_at_least()` in our `DelayUs` impls.

---

This is not only a problem in general, but also directly breaks the requirements of [`DelayUs`](https://docs.rs/embedded-hal/1.0.0-rc.1/embedded_hal/delay/trait.DelayUs.html):
> Pauses execution for at minimum `us` microseconds. Pause can be longer if the implementation requires it due to precision/timing issues.

Finomnis:
Fixes #844.

# Solution
- Compensate for timer uncertainty by always waiting for one extra timer tick.
- Replace all `.micros()` with `.micros_at_least()` and all `.millis()` with `.millis_at_least()`.

# Open questions / Problems
- [x] This is technically a breaking change for `rtic-time`, and indirectly therefore also for `rtic-monotonics` and `rtic` itself. This would require all of those to be updated to a new major version.
    - **Answer:** It's a soundness issue, so it is fine to release it as a patch.
- [x] The previous versions technically did not comply with `embedded-hal`'s definitions of `DelayX`. So should we yank those?
    - **Answer:** Yes. Let's yank previous versions.

### DIFF
--- a/examples/teensy4_blinky/Cargo.lock
+++ b/examples/teensy4_blinky/Cargo.lock
@@ -440,7 +440,7 @@ dependencies = [
 
 [[package]]
 name = "rtic-monotonics"
-version = "1.2.1"
+version = "1.3.0"
 dependencies = [
  "atomic-polyfill",
  "cfg-if",
--- a/rtic-monotonics/CHANGELOG.md
+++ b/rtic-monotonics/CHANGELOG.md
@@ -7,6 +7,10 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ## Unreleased
 
+### Changed
+
+- Update to new `Monotonic` version
+
 ## v1.3.0 - 2023-11-08
 
 ### Added
--- a/rtic-monotonics/CHANGELOG.md
+++ b/rtic-monotonics/CHANGELOG.md
@@ -10,6 +10,8 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 ### Changed
 
 - Update to new `Monotonic` version
+### Fixed
+- **Soundness fix:** Monotonics did not wait long enough in `Duration` based delays.
 
 ## v1.3.0 - 2023-11-08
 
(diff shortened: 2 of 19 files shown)


burrbull:
https://docs.rs/fugit/latest/fugit/trait.ExtU64Ceil.html

Finomnis:
> https://docs.rs/fugit/latest/fugit/trait.ExtU64Ceil.html

That doesn't fix the problem.
- 1ms tick rate, clock is at 0.99ms
- "900.micros_at_least()" -> 1ms delay scheduled
- actual wait time 0.01ms, although `at_least()` was used

To fulfill the requirement, both is actually needed.

With our fix, but without `at_least()`:
- 1ms tick rate, clock is at 0.99ms
- "1200.micros()" - rounds to `1ms` delay
- still only 0.01ms waited, although even with rounding, everyone would expect at least 1ms

Finomnis:
Delays for blocking eh is still broken.
- Implement both delay_us and delay_ms and add one timer tick to the wait.

burrbull:
Add more tests, please. Those cases which this PR fixes.

Finomnis:
> Add more tests, please. Those cases which this PR fixes.

I can add some to `rtic-time`. I can't really add some to the monotonic impls, as this would require an external time measurement separate from the microcontroller.

Would that be good enough then?

burrbull:
Yes, please.

Finomnis:
@korken89 still working on more tests. Need to rewrite the test to do a sub-tick test

korken89:
Alright!

Finomnis:
@burrbull I wasn't able to add tests to the blocking functions; those are hard to test reliably. But I now test all `async` functions. Is that enough for now?

**EDIT:** I added some tests for the blocking functions and they do work, but they are unreliable. So I disabled them by default.

korken89 (approved):
Thanks for the effort!

PR #845, opened 2023-11-28
rtic-sync Arbiter: impl more I2C trait fns

andresv:
For example embassy-stm32 I2C [does not impl transaction yet](https://github.com/embassy-rs/embassy/blob/main/embassy-stm32/src/i2c/mod.rs#L200) but other functions are available. So it would be better to implement all I2c trait functions.


korken89 (approved):
Nice expansion, thanks!

Issue #848, opened 2023-11-29
Race condition in `nrf:rtc` and `nrf:timer` monotonics


Finomnis:
Same problem as #825, and will probably have the same solution.

PR #850, opened 2023-12-01
Refactor race condition free timer helper

Finomnis:
This change is in preparation for fixing #848, to reduce code duplication and increase test coverage of some code that is hard to understand. Also, it makes this method of extending a timer public, so that other people might benefit from it as well.

### DIFF
--- a/rtic-time/Cargo.toml
+++ b/rtic-time/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rtic-time"
-version = "1.0.0"
+version = "1.1.0"
 
 edition = "2021"
 authors = [
--- a/rtic-time/Cargo.toml
+++ b/rtic-time/Cargo.toml
@@ -20,6 +20,7 @@ license = "MIT OR Apache-2.0"
 critical-section = "1"
 futures-util = { version = "0.3.25", default-features = false }
 rtic-common = { version = "1.0.0", path = "../rtic-common" }
+atomic-polyfill = "1"
 
 [dev-dependencies]
 embedded-hal = { version = "1.0.0-rc.2" }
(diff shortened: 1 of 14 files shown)


rtic-time/Cargo.toml
```diff
@@ -20,6 +20,7 @@ license = "MIT OR Apache-2.0"
 critical-section = "1"
 futures-util = { version = "0.3.25", default-features = false }
 rtic-common = { version = "1.0.0", path = "../rtic-common" }
+atomic-polyfill = "1"
```
burrbull: `atomic-polyfill` is deprecated. Use `portable-atomic` instead.

Finomnis: While this is true, `rtic` and `rtic-monotonics` already use `atomic-polyfill`. Should we switch those as well, or use `atomic-polyfill` for now?

Finomnis: I switched `rtic-monotonics` and `rtic-time` to `portable-atomic`. I was scared of touching `rtic` itself, though, as I'm not sure if those are exposed and hence whether it would be a breaking change.

Finomnis: Actually I couldn't make it work with `portable-atomic`, so I switched back to `atomic-polyfill` for now.

Finomnis: Background: https://github.com/taiki-e/portable-atomic/issues/133
Either we use `atomic-polyfill` here, or we wait for this issue to be resolved.

Finomnis: Fix for now: remove `atomic-polyfill` completely and replace it with `core::sync::atomic`s.

korken89 (approved):
Very nice refactor! After a rebase on `master` this should pass CI and I'll merge!

Finomnis:
> Very nice refactor! After a rebase on `master` this should pass CI and I'll merge!

Done. If you merge, make sure to squash ;)

PR #852, opened 2023-12-02
Fix nrf monotonics

### LINKED ISSUE #848: Race condition in `nrf:rtc` and `nrf:timer` monotonics

Finomnis:
Fixes #848.


rtic-monotonics/src/nrf/timer.rs
```diff
@@ -139,17 +140,42 @@ macro_rules! make_timer {
                 // 1 MHz
                 timer.prescaler.write(|w| unsafe { w.prescaler().bits(4) });
                 timer.bitmode.write(|w| w.bitmode()._32bit());
-                timer
-                    .intenset
-                    .modify(|_, w| w.compare0().set().compare1().set());
-                timer.cc[1].write(|w| unsafe { w.cc().bits(0) }); // Overflow
+
+                // Disable interrupts, as preparation
+                timer.intenclr.modify(|_, w| w
```
jamesmunns, line 145: just checking, isn't intenclr "set to one to disable interrupts"? (if there's only a `clear` method that sets high, then ignore me :)

Finomnis: I think yes, that's what clear does

rtic-monotonics/src/nrf/timer.rs
```diff
@@ -139,17 +140,42 @@ macro_rules! make_timer {
...
+                timer.cc[2].write(|w| unsafe { w.cc().bits(0x8000_0000) }); // Half-period
+
+                // Reset the timer
                 timer.tasks_clear.write(|w| unsafe { w.bits(1) });
                 timer.tasks_start.write(|w| unsafe { w.bits(1) });
 
-                $tq.initialize(Self {});
-
+                // Clear pending events.
+                // Should be close enough to the timer reset that we don't miss any events.
```
Finomnis: Might require a critical section to actually guarantee this.

korken89 (approved):
Thank you!

PR #858, opened 2023-12-06
Fix `nrf::rtc` errata workaround

### LINKED ISSUE #851: `rtic_monotonic::nrf::rtc` errata workaround is probably not what was intended
The errata says that the RTC timer only fires if the scheduled CC is at least 2 ticks in the future.

The workaround for this was to provide the `Monotonic::should_dequeue_check` function, which, if the leftover duration is less than 4 ticks, dequeues and wakes the future.

The timer's `delay_until` function sadly doesn't know anything about that, so it reacts with re-inserting the delay back onto the timer queue and triggering an interrupt.

The interrupt again dequeues it and wakes the future.

This infinite loop goes on for up to 4 timer ticks, or 122us. Due to the fact that this ping-pong involves the timer interrupt, which has the highest priority on the system, most certainly this will cause disturbances on the rest of the tasks.

---

There are two ways to mitigate that:
1) Check `should_dequeue_check` inside of the `delay_until` function, to prevent re-queueing. Then, simply wake the waker and return `Pending`, so that the busy-waiting is limited to the low priority task and no longer involves the interrupt handler.
2) Schedule the queue wakeup 4 ticks in the future, even if there are tasks that would have to be woken earlier. `embedded-hal`'s `DelayNs` specifies waiting "at least `ms` milliseconds" anyway.

So, it is a tradeoff between "in time, but higher power usage" (busy waiting) and "delayed, but low power usage" (schedule the wakup late)

A discussion in chat resulted in that it is propably best to go with the second option, as people who use the RTC driver most certainly prefer low power usage over 150us latency.

Further discussion is required.

Another question is also what we do with `Monotonic::should_dequeue_check`, as it's already published, and removing it would be a breaking change. If we go with option 2, I think we should at least deprecate it.

Finomnis:
The problem is that the RTC peripheral will not properly deliver an interrupt if the scheduled compare is less than two ticks in the future. Together with the timer uncertainty it was determined that it is necessary to schedule always at least `now + 3` ticks in the future.

The current workaround, however, did the following. If it checked the queue and if there is a delay in the queue that is less than 3 ticks in the future, it will dequeue it. This means that the delay will only get re-checked and released at whatever timer interrupt comes in the future, or not at all if it was the last thing to ever wait for, causing a deadlock.

This was the current behavior when scheduling a bunch of delays all at once:
```rust
// <desired_release_time_ticks> => <actual_release_time_ticks>
 50 =>  50
100 => 100
101 => 150
150 => 150
200 => 200
201 => ... never
```

This is obviously not desired behavior.

With the rework, the same situation now plays out as follows:
```rust
// <desired_release_time_ticks> => <actual_release_time_ticks>
 50 =>  50
100 => 100
101 => 103
150 => 150
200 => 200
201 => 203
```

Fixes #851.

### DIFF
--- a/rtic-monotonics/CHANGELOG.md
+++ b/rtic-monotonics/CHANGELOG.md
@@ -9,8 +9,9 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ### Fixed
 
-- Race condition in `nrf::timer`.
-- Race condition in `nrf::rtc`.
+- Fix Race condition in `nrf::timer`.
+- Fix Race condition in `nrf::rtc`.
+- Fix errata in `nrf::rtc`.
 - Add internal counter integrity check to all half-period based monotonics.
 
 ## v1.4.0 - 2023-12-04
--- a/rtic-monotonics/CHANGELOG.md
+++ b/rtic-monotonics/CHANGELOG.md
@@ -9,8 +9,8 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ### Fixed
 
-- Fix Race condition in `nrf::timer`.
-- Fix Race condition in `nrf::rtc`.
+- Fix race condition in `nrf::timer`.
+- Fix race condition in `nrf::rtc`.
 - Fix errata in `nrf::rtc`.
 - Add internal counter integrity check to all half-period based monotonics.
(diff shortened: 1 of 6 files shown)


korken89 (approved):
Thanks!

PR #860, opened 2023-12-06
Fix race condition in `calculate_now`

Finomnis:
In fast overflowing timers like the 16bit 1MHz timer on stm32, preempting the current thread while it's inside of `calculate_now` might cause erroneous behavior.

A critical section was therefore added.

**IMPORTANT:** this is a breaking change for `rtic-time`. How to deal with it?
- Would probably require to yank `1.1.0`.


korken89 (approved):
Thanks for fixing!

PR #877, opened 2024-01-07
Changes `make_channel!` macro to accept more types

ansg191:
Changes `type` macro argument from `path` to `ty`, allowing more complex types like tuples, arrays, & pointers.

See https://doc.rust-lang.org/reference/types.html#type-expressions.


korken89 (approved):
Thank you for fixing!

PR #878, opened 2024-01-10
Bump esp32c3 to 0.20.0

onsdagens:
0.20.0 brings some breaking changes to the way registers are accessed, this addresses that and bumps the PAC.


onsdagens:
don't know if this warrants a changelog entry

AfoHT (approved):

PR #883, opened 2024-01-11
ci/qemu build

AfoHT:
- CI: Add extra step building and uploading QEMU
- CI: Print which QEMU are found


perlindgren (approved):
Looks good

PR #888, opened 2024-01-29
Make RTIC 2 not require nightly by using `main`'s stack as an allocator

korken89:
With this PR RTIC v2 will work on stable from 1.75 and forward.

This is done by instead of using Type Alias Impl Trait to do static allocations of async executors (causing us to require nightly), we are now using the stack of `main` (before user `init` is called) to allocate the async executors there. Basically doing a manual `alloca`.

From testing this seems to not have any performance impact compared to TAIT other than startup taking a few extra cycles to move the pointer to the global storage.

### DIFF
--- a/book/en/src/migration_v1_v2.md
+++ b/book/en/src/migration_v1_v2.md
@@ -2,7 +2,7 @@
 
 Migrating a project from RTIC `v1.0.x` to `v2.0.0` involves the following steps:
 
-1. `v2.0.0` requires [`#![type_alias_impl_trait]`](https://github.com/rust-lang/rust/issues/63063) and Rust Nightly.
+1. `v2.1.0` works on Rust Stable from 1.75 (**recommended**), while older versions require a `nightly` compiler via the use of [`#![type_alias_impl_trait]`](https://github.com/rust-lang/rust/issues/63063).
 2. Migrating from the monotonics included in `v1.0.x` to `rtic-time` and `rtic-monotonics`, replacing `spawn_after`, `spawn_at`.
 3. Software tasks are now required to be `async`, and using them correctly.
 4. Understanding and using data types provided by `rtic-sync`.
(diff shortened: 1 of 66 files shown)


rtic-monotonics/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ## Unreleased
 
+### CHanged
```
glaeqen: Typo

book/en/src/preface.md
```diff
@@ -141,7 +141,7 @@ Asynchronous programming in various forms are getting increased popularity and l
 
 The Rust standard library provides collections for dynamically allocated data-structures which are useful to manage execution contexts at run-time. However, in the setting of resource constrained real-time systems, dynamic allocations are problematic (both regarding performance and reliability - Rust runs into a *panic* on an out-of-memory condition). Thus, static allocation is the preferable approach!
 
-RTIC provides a mechanism for `async`/`await` that relies solely on static allocations. However, the implementation relies on the `#![feature(type_alias_impl_trait)]` (TAIT) which is undergoing stabilization (thus RTIC v2.x currently requires a *nightly* toolchain). Technically, using TAIT, the compiler determines the size of each execution context allowing static allocation.
+RTIC provides a mechanism for `async`/`await` that relies solely on allocations in `main` before `init` is called. The implementation does not rely on the `#![feature(type_alias_impl_trait)]` (TAIT) and thus RTIC v2.x **does not** require *nightly* toolchain. A stable Rust compiler from 1.75 is enough.
```
glaeqen: If someone didn't know that RTIC used TAIT before this remark sounds weird, why would someone care what RTIC does not need? It probably should mention since which specific RTIC version TAIT and nightly are not necessary

rtic-macros/src/codegen/bindings/cortex.rs
```diff
@@ -29,11 +29,35 @@ fn is_exception(name: &Ident) -> bool {
...
+            pub static __ebss: u32;
+        }
+
+        let stack_start = &_stack_start as *const _ as u32;
+        let ebss = &__ebss as *const _ as u32;
+
+        if stack_start > ebss {
+            // No flip-link usage, check the MSP for overflow.
+            if rtic::export::msp::read() <= ebss {
+                panic!("Stack overflow after allocating executors");
```
glaeqen, line 55: Is there some kind of test suite that could be extended to verify that panic happens? I'm not sure is there's any "embedded-ci" like test suite that gates the master or not

glaeqen (commented):
Stable yoooo ♥️

rtic-monotonics/CHANGELOG.md
korken89: Fixed

book/en/src/preface.md
korken89: Fixed

rtic-macros/src/codegen/bindings/cortex.rs
korken89: I tried a bit, but was not able to make a good test for this

rtic-macros/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 ### Added
 
 - Unstable ESP32-C3 support.
+- RTIC v2 now works on stable.
```
AfoHT:
```suggestion
- RTIC v2 now works on stable.
- Unstable ESP32-C3 support.
```

Super nitpick :D

AfoHT (dismissed):
Exciting! One tiny thing

rtic-macros/src/codegen/bindings/cortex.rs
glaeqen: Np, maybe in the future then!

book/en/src/migration_v1_v2.md
```diff
@@ -2,7 +2,7 @@
 
 Migrating a project from RTIC `v1.0.x` to `v2.0.0` involves the following steps:
 
-1. `v2.0.0` requires [`#![type_alias_impl_trait]`](https://github.com/rust-lang/rust/issues/63063) and Rust Nightly.
+1. `v2.0.0` works on Rust Stable from 1.75.
```
jonathanpallant: I imagine you'll release this as 2.1.0 or 2.0.2 or something? So maybe it should make clear that 2.0.0 requires nightly but 2.1.0 only requires Rust 1.75 or higher.

korken89: Thank you for the feedback, it is fixed now!

PR #891, opened 2024-02-22
Add example of using Embassy HAL(stm32) with RTIC.

cschuhen:
The RTIC book mentions Embassy+RTIC but gives no examples.


cschuhen:
Not sure what's up with all the failures. Can't see how I triggered that. I didn't even think CI would know how to build what I added.

korken89 (approved):

PR #874, opened 2023-12-27
Monotonic rewrite

Finomnis:
See #873.

PR #893, opened 2024-02-23
Try testing all RTIC monotonics

korken89:


Finomnis:
Clashes with #874.

Also only tests the `tim2` peripheral; as many of the macros are hand-synced code duplication, it would make much sense to test all the timers.

korken89:
This won't be merged, the compile time goes outside the max limits of the CI xD

PR #905, opened 2024-03-17
Bump rp2040-pac to 0.6 in rtic-monotonics

martinsp:
⚠️ `examples/rp2040_local_i2c_init` depends on unreleased https://github.com/rp-rs/rp-hal-boards/pull/60

### DIFF
--- a/examples/rp2040_local_i2c_init/Cargo.lock
+++ b/examples/rp2040_local_i2c_init/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
 
 [[package]]
 name = "atomic-polyfill"
-version = "1.0.3"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


korken89:
Thanks for the PR! Can you have a look at fixing the failing test?

martinsp:
Looks like this needs to wait for https://github.com/rp-rs/rp-hal-boards/pull/60 otherwise there are two versions of `rp204-pac` pulled in in the `examples/rp2040_local_i2c_init` and causes conflicting types.
- rp2040-pac 0.5 through rp-pico -> rp2040-hal (0.9.2)
- rp2040-pac 0.6 through rtic-monotonic

korken89:
I gave this a look now, I think this is a breaking change no? As this would make the RP2040 monotonic stop working with the older PACs.
We've not set a policy yet on how we see breaking PAC changes, but lets add this to tomorrows meeting.
I'm fine with tracking the latest PACs as the contract of `rtic-monotonics` but lets see what others think.

Else one needs to add a feature flag for specific PAC version spans as breaking changes come.

korken89 (approved):
Since we will do a major release this is now fine!

PR #914, opened 2024-04-04
CI: Add ESP32-C3 QEMU support

AfoHT:
Use [forked and patched QEMU from espressif](https://github.com/espressif/qemu)

- **CI: ESP32-C3: QEMU support**
- **CI: ESP32C3: Script for preparing and running**
- **CI: gitignore qemu.log**
- **ESP32-C3: Example sw_and_hw output**


korken89 (approved):
ESP32C3 testing :tada:

PR #922, opened 2024-04-16
Breaking change: portable-atomic

### LINKED ISSUE #918: `atomic-polyfill` is deprecated
Hi! While scraping crates.io I've found this repository to be using `atomic-polyfill` despite it being declared as deprecated (see https://crates.io/crates/atomic-polyfill).

You may want to replace it with [`portable-atomic`](https://crates.io/crates/portable-atomic).

AfoHT:
- **Atomics: Replace polyfill with portable-atomic**


korken89:
The failing tests is because you need the set the feature flag that allows for the critical section to be used on thumbv6, I think it is `unsafe-assume-single-core`.

korken89 (approved):

PR #931, opened 2024-05-01
rtic-sync/channel: Make debug asserts into asserts to help find the issue

korken89:


AfoHT (approved):
Let the :bug:-hunt begin! :eyes:

PR #934, opened 2024-05-07
Add Signal to rtic-sync

AdinAck:
Add the Signal structure to rtic-sync.

A Signal has unlimited writers and one reader, many writes can occur before a read, all overwriting the previous.

A reader can asynchronously wait for a write, requiring the write to occur *during* the lifetime of the wait future, or immediately resolve from previous writes (up to the user).

That's it, thanks!

### DIFF
--- a/rtic-sync/src/signal.rs
+++ b/rtic-sync/src/signal.rs
@@ -1,0 +1,148 @@
+//! A "latest only" value store with unlimited writers and async waiting.
+//!
+//! Example usage:
+//!
+//! ```rust
+//! fn init(ctx: init::Context) -> (Shared, Local) {
+//!     // mono set up among other things...
+//!
+//!     let (writer, reader) = make_signal!(u8);
+//!
+//!     writer_task::spawn(writer);
+//!     reader_task::spawn(reader);
+//! }
+//!
+//! #[task]
+//! async fn writer_task(_ctx: writer_task::Context, mut writer: SignalWriter<u8>) {
+//!     for i in 0..10 {
+//!         writer.write(i);
+//!         Mono::delay(500.millis()).await;
+//!     }
+//! }
+//!
+//! #[task]
+//! async fn reader_task(_ctx: reader_task::Context, mut reader: SignalReader<u8>) {
+//!     loop {
+//!         let value = reader.wait().await;
+//!         defmt::info("received value: {}", value);
+//!     }
+//! }
+//! ```
+
+use core::{
+    cell::UnsafeCell,
+    future::poll_fn,
+    sync::atomic::{fence, Ordering},
+    task::Poll,
+};
+use rtic_common::waker_registration::CriticalSectionWakerRegistration;
+
+/// Basically an Option but for indicating
+/// whether the store has been set or not
+#[derive(Clone, Copy)]
+enum Store<T> {
+    Set(T),
+    Unset,
+}
+
+/// An async message passing structure with unlimited writers and one reader.
+pub struct Signal<T: Copy> {
+    waker: CriticalSectionWakerRegistration,
+    store: UnsafeCell<Store<T>>,
+}
+
+unsafe impl<T: Copy> Send for Signal<T> {}
+unsafe impl<T: Copy> Sync for Signal<T> {}
+
+impl<T: Copy> Signal<T> {
+    /// Create a new signal.
+    pub const fn new() -> Self {
+        Self {
+            waker: CriticalSectionWakerRegistration::new(),
+            store: UnsafeCell::new(Store::Unset),
+        }
+    }
+
+    /// Split the signal into a writer and reader.
+    pub fn split(&'static self) -> (SignalWriter<T>, SignalReader<T>) {
+        (SignalWriter { parent: self }, SignalReader { parent: self })
+    }
+}
+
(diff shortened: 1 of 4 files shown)


rtic-sync/src/signal.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// Create a new signal.
+    pub const fn new() -> Self {
+        Self {
+            waker: CriticalSectionWakerRegistration::new(),
+            store: UnsafeCell::new(Store::Unset),
+        }
+    }
+
+    /// Split the signal into a writer and reader.
+    pub fn split(&'static self) -> (SignalWriter<T>, SignalReader<T>) {
```
korken89: Is there a reason to force `'static` and not let the lifetime system figure it out?

rtic-sync/src/signal.rs
```diff
@@ -0,0 +1,119 @@
...
+/// Fascilitates the writing of values to a Signal.
+#[derive(Clone)]
+pub struct SignalWriter<T: 'static + Copy> {
+    parent: &'static Signal<T>,
+}
+
+impl<T: Copy> SignalWriter<T> {
+    /// Write a raw Store value to the Signal.
+    fn write_inner(&mut self, value: Store<T>) {
+        fence(Ordering::SeqCst);
```
korken89: This fence looks like it's not needed, what's the reason for it?

rtic-sync/src/signal.rs
```diff
@@ -0,0 +1,119 @@
...
+        })
+    }
+
+    /// Wait for a new value to be written and read it.
+    ///
+    /// If a value is already pending it will be returned immediately.
+    pub async fn wait(&mut self) -> T {
+        poll_fn(|ctx| match self.take() {
+            Store::Unset => {
+                self.parent.waker.register(ctx.waker());
```
korken89: This is a race-condition.

The registration must happen before `self.take()` on line 91 as the value can be set (and fail to be woken) between the `self.take()` but before the waker registration.

korken89 (changes requested):
Thanks for adding! I have some comments that need looking into.

rtic-sync/src/signal.rs
AdinAck: Of course, I've updated it to be generic.

rtic-sync/src/signal.rs
AdinAck: To be honest I don't understand compiler fencing and simply included it because during my research looking through the other rtic-sync structures I saw it used just before critical sections often. I'll remove them.

rtic-sync/src/signal.rs
AdinAck: Ah very interesting. I was trying to avoid registration if a value is already there, but it seems unavoidable.

By the way, I noticed the WakerRegistration never *clears* the registered waker, and during my experimentation I noticed that calling wake on a Waker belonging to an expired Future didn't do anything. I assume this is a feature of the rtic executor but was curious if this was truly the case?

rtic-sync/src/signal.rs
korken89: In this case the critical section does everything with the shared state so there should not be any need for extra synchronization :rocket:

rtic-sync/src/signal.rs
korken89: :rocket:

rtic-sync/src/signal.rs
korken89: Yeah it's generally unavoidable - but the async executor and tasks are robust to spurious wakes. So in case an extra wake does happen it's no problem. 

One can add an API to the waker registration to clear a registration if it was not needed, but I've never measured the impact of it.
And if we don't measure we don't know if it will be better or worse :)

korken89 (changes requested):
Thanks for the updates! Can you have a look at adding tests?
You can copy the tests from channel if you need inspiration.

rtic-sync/src/signal.rs
AdinAck: I actually did add that at one point but removed it because it did incur a slight performance hit, and of course I observed no issue with *not* clearing it.

AdinAck:
Sure! I also planned on adding an entry to `examples` later, I can include that in this PR along with the tests if you like.

AdinAck:
Ok I think I added all the necessary tests.

I'm not quite sure about the soundness of the last test though, it works on my machine and the GitHub action, basically I yield the test thread to get tokio to poll the reader future waiting for a "fresh" write *before the write occurs*. What are your thoughts?

```rust
#[tokio::test]
async fn waiting() {
    static READER: StaticCell<SignalReader<u32>> = StaticCell::new();
    let (mut writer, reader) = make_signal!(u32);

    writer.write(0xaa);

    let reader = READER.init(reader);
    let handle = tokio::spawn(reader.wait_fresh());

    tokio::task::yield_now().await; // encourage tokio executor to poll reader future
    assert!(!handle.is_finished()); // verify reader future did not resolve after poll

    writer.write(0xab);

    assert!(handle.await.is_ok_and(|value| value == 0xab));
}
```

korken89 (approved):
Thanks for the effort!

Scutua (approved):

PR #937, opened 2024-05-09
updated esp32c3 dependency to version 0.22.0

broke:
This PR updates RTIC esp32c3 dependency to version 0.22.0 as well as version used for the corresponding example


AfoHT (approved):
👍

PR #943, opened 2024-05-27
Improve error output for prios > dispatchers

hegza:
I ran into this `UNREACHABLE` while porting RTIC for an experimental RISC-V platform. I decided to improve the error for others. Platform-specific analysis may catch this as well but I would've found this implementation helpful, were it to exist prior to my port.


korken89 (approved):
Thanks!

PR #960, opened 2024-07-05
Fix STM32 Monotonic for peripherals with only two Clock Compare modules

### LINKED ISSUE #956: rtic-monotonic panics

Finomnis:
Fixes #956.
Replaces #958.

Rationale: Some timer peripherals seem to have only two clock compare modules. Therefore the half period interrupt does not fire and the monotonic sanity check triggers.

Solution: Use CC1 instead of CC3. This was more or less a typo anyway; we originally intended to use CC1 and CC2, we just ended up using CC2 and CC3 because we didn't realize that the enumeration starts with 0.

Review: @korken89

### DIFF
--- a/examples/stm32g030f6_periodic_prints/.cargo/config.toml
+++ b/examples/stm32g030f6_periodic_prints/.cargo/config.toml
@@ -1,6 +1,6 @@
 [target.'cfg(all(target_arch = "arm", target_os = "none"))']
 # TODO(2) replace `$CHIP` with your chip's name (see `probe-run --list-chips` output)
-runner = "probe-run --chip STM32G030F6Px"
+runner = "probe-rs run --chip STM32G030F6Px"
 rustflags = [
   "-C", "linker=flip-link",
   "-C", "link-arg=-Tlink.x",
--- a/examples/stm32g030f6_periodic_prints/Cargo.lock
+++ b/examples/stm32g030f6_periodic_prints/Cargo.lock
@@ -76,9 +76,9 @@ checksum = "7059fff8937831a9ae6f0fe4d658ffabf58f2ca96aa9dec1c889f936f705f216"
 
 [[package]]
 name = "defmt"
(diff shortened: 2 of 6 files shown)


Finomnis:
Seems to still run fine on my stm32g030, indicating that the CC indices indeed start at 0.

@great-houk, would you mind testing if this fixes your problem?

EDIT: Ran for 3800 seconds now, still fine.

great-houk:
I can confirm it works, thanks for your help.

korken89 (approved):
Thanks for the fix!

Finomnis:
> Thanks for the fix!

@korken89 Might want to yank the previous two releases if such a serious bug existed.

PR #967, opened 2024-08-07
Fix codegen emitting unqualified `Result`

### LINKED ISSUE #965: rtic::app macro "should be producing references to ::core::result::Result, not unqualified Result"
[Rust forum discussion](https://users.rust-lang.org/t/rtic-app-macro-problem/114576/5?u=anichang)

korken89:
Closes https://github.com/rtic-rs/rtic/issues/965


perlindgren (approved):
now refers to qualified path

PR #970, opened 2024-08-31
Feature/rp235x

taunusflieger:
Add support for RP235x (Raspberry Pico 2)

PR #971, opened 2024-08-31
Create rp235x monotonic backend

pta2002:
Create a monotonic backend for the RP235x family of chips, based off the RP2040 backend, but adjusted to take into account the new timer peripherals.

I've tested this in a Pi Pico 2 and it works identically to the RP2040 backend on a Pi Pico 1

WIP: For now, this will always use the `TIMER0`, but I still want to make this configurable.


pta2002:
Looks like https://github.com/rtic-rs/rtic/pull/970 got ahead of me with an identical implementation, so I'll close this one in favor of that PR.

PR #972, opened 2024-09-02
Add SYSTIMER based ESP32-C3 monotonic

onsdagens:
This also cleans up the ESP32-C3 examples a little bit, moving the crate to the stable channel and clarifying the README.


onsdagens:
I've added expected output from a QEMU run of the ESP32-C3 monotonic example. 

I'm not sure if that's a good idea, since the order of task spawn and output has the potential for not being very deterministic, so i think i'd rather avoid running that example in QEMU altogether, but i don't know how to do that nicely, maybe an xtask pro could chip in?

korken89 (approved):
Very nice, thank you!

PR #975, opened 2024-09-05
Update support/example for ESP32-C3 to use latest versions of dependencies

jessebraham:
This updates the `rtic` and `rtic-macros` packages to handle some changes which have been made since support for the ESP32-C3 has been updated last. It additionally updates the `sw_and_hw` example for this device, which I have tested and verified using a `ESP32-C3-DevKit-RUST-1 v1.2a` board.

If there are any questions or concerns please let me know, I'm happy to make any requested changes.

This supersedes #973 as well, so that can be closed when/if this is merged.

### DIFF
--- a/examples/esp32c3/examples/sw_and_hw.rs
+++ b/examples/esp32c3/examples/sw_and_hw.rs
@@ -4,17 +4,17 @@
 mod app {
     use esp_backtrace as _;
     use esp_hal::{
-        gpio::{Event, Gpio9, Input, PullDown, IO},
+        gpio::{Event, GpioPin, Input, Io, Pull},
         peripherals::Peripherals,
-        prelude::*,
     };
     use esp_println::println;
+
     #[shared]
     struct Shared {}
 
     #[local]
     struct Local {
-        button: Gpio9<Input<PullDown>>,
+        button: Input<'static, GpioPin<9>>,
     }
 
     // do nothing in init
@@ -22,14 +22,14 @@ mod app {
     fn init(_: init::Context) -> (Shared, Local) {
         println!("init");
         let peripherals = Peripherals::take();
-        let io = IO::new(peripherals.GPIO, peripherals.IO_MUX);
-        let mut button = io.pins.gpio9.into_pull_down_input();
+        let io = Io::new_no_bind_interrupt(peripherals.GPIO, peripherals.IO_MUX);
+        let mut button = Input::new(io.pins.gpio9, Pull::Up);
         button.listen(Event::FallingEdge);
         foo::spawn().unwrap();
         (Shared {}, Local { button })
     }
 
-    #[idle()]
+    #[idle]
     fn idle(_: idle::Context) -> ! {
         println!("idle");
         loop {}
@@ -46,6 +46,7 @@ mod app {
         }
         println!("Leaving high prio task.");
     }
+
     #[task(priority = 2)]
     async fn bar(_: bar::Context) {
         println!("Inside low prio task, press button now!");
--- a/rtic/src/export/riscv_esp32c3.rs
+++ b/rtic/src/export/riscv_esp32c3.rs
@@ -1,7 +1,6 @@
-use esp32c3::INTERRUPT_CORE0; //priority threshold control
+use esp32c3::INTERRUPT_CORE0;
 pub use esp32c3::{Interrupt, Peripherals};
-pub use riscv::interrupt;
-pub use riscv::register::mcause; //low level interrupt enable/disable
+pub use riscv::{interrupt, register::mcause};
 
 #[cfg(all(feature = "riscv-esp32c3", not(feature = "riscv-esp32c3-backend")))]
 compile_error!("Building for the esp32c3, but 'riscv-esp32c3-backend not selected'");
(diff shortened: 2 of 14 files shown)


examples/esp32c3/examples/sw_and_hw.rs
```diff
@@ -5,32 +5,32 @@
...
     }
 
     // do nothing in init
     #[init]
     fn init(_: init::Context) -> (Shared, Local) {
         println!("init");
         let peripherals = Peripherals::take();
-        let io = IO::new(peripherals.GPIO, peripherals.IO_MUX);
-        let mut button = io.pins.gpio9.into_pull_down_input();
+        let io = Io::new_no_bind_interrupt(peripherals.GPIO, peripherals.IO_MUX);
```
jessebraham, line 25: I will note that this is a bit of a rough edge for which we don't have a good solution yet. Essentially the issue is that `Io::new()` binds a handler to the `GPIO` interrupt for async functionality, so if we want to use this interrupt for other purposes (like in the `sw_and_hw` example, when creating the hardware task) then we need to _not_ bind the interrupt in the constructor.

This is potentially a bit of a footgun, so I can add a comment here if necessary.

jessebraham:
I'm not sure what you'd like me to do about the QEMU failure, I would not expect this output to stay consistent (as we're seeing with this failure).

rtic/src/export/riscv_esp32c3.rs
```diff
@@ -138,28 +138,30 @@ pub fn unpend(int: Interrupt) {
...
-                                                //should maybe figure out a workaround that
-                                                //doesnt involve raw pointers.
-                                                //Again, this is how they do it in the HAL
-                                                //but i'm really not a fan.
-    let interrupt_number = int as isize;
-    let cpu_interrupt_number = cpu_int_id as isize;
+    const INTERRUPT_MAP_BASE: u32 = 0x600C_2000;
+    const RESERVED_INTERRUPTS: &[u8] = &[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
+
+    if RESERVED_INTERRUPTS.contains(&cpu_int_id) {
```
onsdagens: The line [here](https://github.com/rtic-rs/rtic/pull/975/files#diff-3fe7541514399a332f826c2d8567f0708a53cd71bcadd777705979177f0b2beeR229) already ensures this doesn't happen. 

This function is not meant to be called directly by the user, and is side-effectful, maybe it should be marked as `unsafe` to reflect that?

rtic/src/export/riscv_esp32c3.rs
```diff
@@ -138,28 +138,30 @@ pub fn unpend(int: Interrupt) {
...
+        panic!("interrupt {cpu_int_id} is reserved!");
+    }
+    if prio == 0 {
```
onsdagens: Similarly, no interrupt will ever actually be configured at priority 0 by RTIC. We can add an additional compile-time check to really make sure.

onsdagens:
Tested this locally on the same board, and it seems to work as expected.

I've fixed up the QEMU expected output, applied my suggestions and opened a PR on your fork.

onsdagens:
i see my review is set as pending, so i'm not sure if you can see my comments. in any case, i see now what you were saying about QEMU, the output does not match what is expected, and what actually happens when running on a board

i think we should keep QEMU around, adjust the expected output to the unexpected results and use it as a marker for when something is potentially wrong and needs to be run on a board.

jessebraham:
I agree, it's probably worth keeping the checks around, however perhaps being a bit less strict with regards to the bootloader output would be best.

Anyway, with your changes merged CI is green here now at least, so thanks for that!

examples/esp32c3/examples/sw_and_hw.rs
jessebraham: There is a tracking issue for this in the `esp-hal` repo, FYI: https://github.com/esp-rs/esp-hal/issues/2009

onsdagens:
i guess i just forgot to press submit, the comments are now outdated but i will leave them here for the record.

korken89:
Great work, thank you!

I just merged some changes from @onsdagens , could you double check this PR is still OK?
E.g. there was a reference to `esp32c3` v0.22 in `rtic-montonics` now as well.

jessebraham:
Thanks for taking a look. I have rebased my branch locally but am running into some issues with the monotonic example; will investigate further.

Might take me a few days to get to this, lots going on right now for me :)

korken89:
Thank you, no rush!

jessebraham:
I've resolved the issues I had encountered with the monotonic example, seems to be working correctly now.

Unfortunately, again we are having issues with the checks on the QEMU output. Happy to make whatever changes here, but I am not very familiar with these checks so am not sure how best to resolve this long-term. I think that the bootloader output can probably be ignored entirely in the QEMU checks.

AfoHT:
I agree z

> I've resolved the issues I had encountered with the monotonic example, seems to be working correctly now.
> 
Awesome! :rocket: 
> Unfortunately, again we are having issues with the checks on the QEMU output. Happy to make whatever changes here, but I am not very familiar with these checks so am not sure how best to resolve this long-term. I think that the bootloader output can probably be ignored entirely in the QEMU checks.

I agree, bootloader is not the main part. Do you want to attempt implement the ignoring of bootloader output, or should we join forces on that one?

jessebraham:
> Do you want to attempt implement the ignoring of bootloader output, or should we join forces on that one?

Would be great if I could get a hand with that, please. New project at work will be consuming a fair bit of my brainpower and time for at least the next couple weeks here.

AfoHT (approved):
:rocket:

jessebraham:
Thank you for wrapping this up!

PR #992, opened 2024-11-04
Remove unused crate proc-macro-error

epontan:
The `proc-macro-error` crate is unmaintained[1] and is causing `cargo audit` failures[1]. Since the create is not even used by `rtic`, just remove it as a dependency.

[1]: https://rustsec.org/advisories/RUSTSEC-2024-0370.html

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -7,6 +7,9 @@ fn main() {
         println!("cargo:rustc-cfg=rustc_is_nightly");
     }
 
+    // Make `have_basepri` an expected cfg.
+    println!("cargo::rustc-check-cfg=cfg(have_basepri)");
+
     // These targets all have know support for the BASEPRI register.
     if target.starts_with("thumbv7m")
         | target.starts_with("thumbv7em")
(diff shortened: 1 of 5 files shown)


epontan:
I realize that there is still a bunch of compilation issues left for this release branch. Is this change something that you are willing to take in and release for the v1 track? I just want to get confirmation of that before I spend too much time trying to fix things.

AfoHT:
I think it's a good thing, and as there should be no functional change I don't see why not. The work you're doing is appreciated!

build.rs
```diff
@@ -7,6 +7,9 @@ fn main() {
         println!("cargo:rustc-cfg=rustc_is_nightly");
     }
 
+    // Make `have_basepri` an expected cfg.
+    println!("cargo::rustc-check-cfg=cfg(have_basepri)");
```
elmjag, line 11: This config have been added via different PR now:
https://github.com/rtic-rs/rtic/commit/ba3af7e3fb4bdd78de44e1ee2e58ace3c639f57f

elmjag:
> I realize that there is still a bunch of compilation issues left for this release branch.

I think all compilation issues are now fixed on `release/v1`  branch. 

The `proc-macro-error` is still listed as dependency, it would be good to drop it. I suggest rebasing this PR on latests `release/v1` and merging it.

AfoHT (approved):
Looking good!

PR #994, opened 2024-11-11
atsamd21: initial implementation usign TC4/5

maximeborges:
Initial implementation of a monotonic timer using half-period counting with TC4/TC5.

Tested on a custom device using an ATSAMD21G16B.

### DIFF
--- a/rtic-monotonics/CHANGELOG.md
+++ b/rtic-monotonics/CHANGELOG.md
@@ -7,6 +7,10 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ## Unreleased
 
+### Added
+
+- ATSAMD21G support
+
 ## v2.0.3 - 2024-10-23
 
 ### Added
--- a/rtic-monotonics/Cargo.toml
+++ b/rtic-monotonics/Cargo.toml
@@ -32,6 +32,7 @@ features = [
     "stm32_tim5",
     "stm32_tim15",
     "esp32c3-systimer",
+    "atsamd21g",
 ]
 rustdoc-flags = ["--cfg", "docsrs"]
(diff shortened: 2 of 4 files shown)


Sympatron:
Does this implementation run the timer at 48MHz? If so, why not prescale it to 1MHz? 1µs resolution is plenty for almost any use case and it will generate less overflows and allows for greater intervals to be usable.

maximeborges:
> Does this implementation run the timer at 48MHz? If so, why not prescale it to 1MHz? 1µs resolution is plenty for almost any use case and it will generate less overflows and allows for greater intervals to be usable.

As a first step for simplicity, since there are currently no RTIC v2 monotonic implementation for the SAMD family (now the [RTC implementation](https://github.com/rtic-rs/rtic/issues/995) seems to be ready to use also).
I would definitely accept PR to add support for other chips in the family to improve this one. I'm a bit constrained by time until the end of the month to polish things myself here, but could help other initiatives.

kyp44:
Just wanted to note that it was [decided](https://hackmd.io/c_mFUZL-Q2C6614MlrrxOg?view#Agenda-2024-12-18) along with the HAL team that [ATSAMD RTC-based monotonic drivers](https://github.com/atsamd-rs/atsamd/pull/804) should be released as part of the HAL instead of in `rtic-monotonics`. I would assume this should apply to this timer-based driver as well.

AfoHT:
I want to thank @maximeborges for your efforts, but as discussed and agreed with atsamd-hal devs it brings more value to have it "in house" in relation to atsamd.

Thus I will close this PR, but I hope the actual code will live and prosper over at https://github.com/atsamd-rs/atsamd

PR #996, opened 2024-11-12
Do not limit async priority with `NVIC_PRIO_BITS` when targeting ESP32-C3

jessebraham:
Since RISC-V devices do not have an NVIC, we set the `nvicPrioBits` field in our SVDs simply to 0. As such, this value should not be used to determine the maximum async priority, and I have aligned the codegen bindings for this target with the `riscv_slic` version.


jessebraham:
Sorry, forgot I had a PR open with this branch and force pushed 😅

korken89 (approved):

PR #1007, opened 2024-12-23
Add blinky example for STM32F1 bluepill board

rxwb:
Code taken from stm32f3_blinky example with LED port adapted to bluepill board.
Port initialization and README from v1 rtic-examples with slight edits.


AfoHT (approved):
Nice! Always good with some different boards :) Thank you

PR #1010, opened 2025-01-15
rtic-monotonics: rp2040 extra parenthesis

AfoHT:


perlindgren (approved):
Just a clippy issue, should be fine.

PR #1015, opened 2025-01-27
Add standalone waker constructor for executor and expose it

Ddystopia:
Currently we can spawn a task with `task::spawn()`, but there is no way to get a waker. This PR adds const function `task::waker()`, that returns a waker that will wake that task. 

My use case for this is smoltcp, they expose a lot of functions like that one: [Socket::register_waker](https://docs.rs/smoltcp/latest/smoltcp/socket/dhcpv4/struct.Socket.html#method.register_waker)

### DIFF
--- a/rtic-macros/CHANGELOG.md
+++ b/rtic-macros/CHANGELOG.md
@@ -7,6 +7,10 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ## [Unreleased]
 
+### Added
+
+- Added `waker` getter to software tasks
+
 ## [v2.1.3] - 2025-06-08
 
 ### Changed 
--- a/rtic-macros/src/codegen/module.rs
+++ b/rtic-macros/src/codegen/module.rs
@@ -192,7 +192,7 @@ pub fn codegen(ctxt: Context, app: &App, analysis: &Analysis) -> TokenStream2 {
             #[allow(non_snake_case)]
             #[doc(hidden)]
             pub fn #internal_waker_ident() -> ::core::task::Waker {
-                // SAFETY: If `try_allocate` succeeds one must call `spawn`, which we do.
+                // SAFETY: #exec_name is a valid pointer to an executor.
                 unsafe {
                     let exec = rtic::export::executor::AsyncTaskExecutor::#from_ptr_n_args(#name, &#exec_name);
                     exec.waker(|| {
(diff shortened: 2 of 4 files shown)


korken89:
Hi! Thanks for the PR.

Can you motivate the need for this a bit more? As I see it, and e.g. what `embassy-net` does internally (https://github.com/embassy-rs/embassy/blob/main/embassy-net/src/lib.rs#L797) is the following to not need to have a way to get a waker:

```rust
async fn do_something() {
    poll_fn(|cx| {
        smoltcp.register_waker(cx.waker());

        if smoltcp.api_that_later_will_wake().is_done_or_something() {
            Poll::Ready(())
        } else {
            Poll::Pending
        }
    }).await
}
```

Ddystopia:
> Hi! Thanks for the PR.
> 
> Can you motivate the need for this a bit more? As I see it, and e.g. what `embassy-net` does internally (https://github.com/embassy-rs/embassy/blob/main/embassy-net/src/lib.rs#L797) is the following to not need to have a way to get a waker:
> 
> ```rust
> async fn do_something() {
>     poll_fn(|cx| {
>         smoltcp.register_waker(cx.waker());
> 
>         if smoltcp.api_that_later_will_wake().is_done_or_something() {
>             Poll::Ready(())
>         } else {
>             Poll::Pending
>         }
>     }).await
> }
> ```

Sometimes I need to wake other task than the one setting the waker, or with some other logic. So I would need to create the waker with poll_fn and store it somewhere, then retrieve with other methods. And it can't be done in const contexts. With that change code is very straightforward by directly using the waker for a task. 

So yes, it is possible, for non-const use cases, to use `poll_fn` to extract the waker, but when use case is more complicated then a basic one, user of the api would suffer, while it is really trivial for rtic to add that getter, as presented in pr.

Ddystopia:
@korken89 is this motivation enough? This allows to not require a poll and not store a waker somewhere if you are not using it directly inside, as `embassy-net` is doing. This PR is really small while ergomic gains are larger than PR.

korken89 (dismissed):
It's no big addition, I'll add it. :+1:

korken89:
@Ddystopia can you fix the conflicts and I'll merge?

AfoHT (approved):
Rebased and moved changelog entries to Unreleased (releases happened) :)

Thanks!

PR #1017, opened 2025-01-28
Add MECALL backend for RISC-V targets

### LINKED ISSUE #998: docs: Be more vocal about backends different than NVIC
Hi,

I only found out that RTIC already supports some RISC-V targets when I revisited an idea to port it, it would be super cool if documentation could be more vocal about it.

Also, since there are already plenty of backends already, I think it would be very helpful for a broader audience if there was a porting guide for new targets

romancardenas:
This PR adds a new backend for every RISC-V target. It uses the SLIC philosophy of using a software interrupt controller that multiplexes an interrupt source for every task. However, instead of relying on a CLINT peripheral with software interrupts, it now triggers machine environment call exceptions to do the trick.

**Pros of this approach:** It is valid for every RISC-V target, regardless of its peripherals. Thus, it is very versatile. It also supports virtually an unlimited number of tasks, as long as they fit in memory. Also, users do not need to bind tasks to interrupt sources manually, as they are synthetically generated when needed.

**Cons of this approach:** It has more software overhead than using a dedicated physical interrupt controller. For example, for ESP32C3 targets, it may make more sense to use its dedicated ESP32C3 backend.


romancardenas:
CI tweaked and passes. I will publish the new version of `riscv-slic` as soon as you are happy with this new backend.

rtic/CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ Example:
 - Updated esp32c3 dependency to v0.22.0
 - Use `riscv-slic` from `crates.io`
 - Remove unused dependency `rtic-monotonics`
+- Updated esp32c3 dependency to v0.25.0
```
AfoHT: Nitpick: Please put these at the top of changelog (like a stack)

romancardenas:
I addressed all the required actions except for documentation. Where should I write about using `riscv-clint-backend` and `riscv-mecall-backend`?

AfoHT:
> Where should I write about using `riscv-clint-backend` and `riscv-mecall-backend`?

I think we need something in the main RTIC readme, something in the book and in the specific crates that now support it.  (for starters in that priority order) 

There's this ticket #998 we can close by at least doing the main readme stating what RTIC currently supports. 

Suggestions: 

* For the book a small note in 1 "Starting a new project" about the two options, up to "now" there was only one option... 

* Add 8.2 RISC-V with some more in depth explanation, perhaps comparing the alternatives. Presenting MECALL. 

Do note that parts of the readme gets included in the book, so something to watch out for. (see preface.md as example)

romancardenas:
I added a few docs in the main `README.md` and in the "Starting a new project" section of the book. Let me know what you think and then I will write a draft of  Section 8.2

AfoHT:
> Let me know what you think and then I will write a draft of Section 8.2

I think it looks really good! :rocket:

romancardenas:
Sorry for the delay, I've been quite busy these weeks. Let me know if the docs are OK.

AfoHT:
> Let me know if the docs are OK.

Docs look awesome! I've fixed the mdbook title (super hidden in SUMMARY.md) and a potential markdown issue.

@korken89 any input on this? I vote :rocket: :clock1:  time to merge :D

AfoHT (approved):
Nice work!!! :rocket:

PR #1019, opened 2025-02-09
Update esp32-c3 and esp-hal crates

rxwb:
Update the ESP32-C3 examples to the latest versions of the HAL crates.
Tested to be working on a real board.


AfoHT (approved):
Awesome! :rocket: Thank you :)

PR #1027, opened 2025-03-09
Support `loom` in `rtic-common` and `rtic-sync`. Fix `rtic-sync` bug.

### LINKED ISSUE #780: assertion failed: !self.0.access(cs).freeq.is_empty() in rtic-sync 1.0.1 when trying to send data to channel
> panicked at 'assertion failed: !self.0.access(cs).freeq.is_empty()', /home/serg/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rtic-sync-1.0.1/src/channel.rs:301:17

I've tried checking the code myself, but it is not trivial, so any help would be appreciated

datdenkikniet:
Adds `--cfg loom` support to `rtic-common` and `rtic-sync`.

Additionally, adds a test for an `rtic-sync` bug & fixes it (I cannot find the bug report itself here on github).

Perhaps splitting out the bug fix into its own PR ([d8f8a15](https://github.com/rtic-rs/rtic/pull/1027/commits/d8f8a155e0b4b3dedc1b5386fa3d6dced23ce830)) is desirable. I can do that if we deem that it makes sense.

Fixes #780 

A description of the path leading to the bug, which also proves that the fix actually solves the problem:

1. `try_recv` pushes into `freeq`.
2. Before popping `wait_queue`, an interrupt occurs.
3. `send` no. 1 eats up the `freeq` slot.
4. `send` no. 2 is called, pushes its link to `wait_queue` and goes to sleep (because `freeq` is now empty)
5. Control returns to `try_recv`, which pops `wait_queue` (from the newly waiting `send` call) _without_ touching the free queue.
6. `send` is awoken at some point. Its link is popped, but the `freeq` is empty.


datdenkikniet:
Closing in favor of #1038

PR #1033, opened 2025-03-13
rtic-sync: fix another instance of #780 (panic because `freeq` is empty when it should not be)

datdenkikniet:
This time:

1. `send` registers a waker, goes to sleep.
2. `send` is awoken, validates that `!self.is_closed()`.
3. An interrupt occurs.
4. `Receiver` is dropped, pops waker
5. `send` is awoken, calls `poll_fn` with potentially empty `freeq` (because `drop`-ing `Receiver` does not add freeq items).

`loom` test that explicitly drops `Receiver` before `join`-ing a `send` thread confirms that this bug exists, and that this code change fixes it.

### DIFF
--- a/rtic-sync/src/channel.rs
+++ b/rtic-sync/src/channel.rs
@@ -285,12 +285,12 @@ impl<T, const N: usize> Sender<'_, T, N> {
         });
 
         let idx = poll_fn(|cx| {
-            if self.is_closed() {
-                return Poll::Ready(Err(()));
-            }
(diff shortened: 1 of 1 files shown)


datdenkikniet:
There is a better, more holistic fix for the class of problem we're trying to prevent here (namely: prevent the waker from being woken with a popped link, _without_ a freeq slot available, by construction). Will open a different PR with that.

PR #1034, opened 2025-03-13
rtic-sync: explicitly send an awoken Sender the slot it can use to avoid TOCTU bugs

datdenkikniet:
This PR ensures that `send` behaves consistently in the face of concurrent sending and receiving to and from a channel.

Since we now no longer rely on a slot being available in `freeq`, but rather on direct message passing by the `try_recv` function that cannot be re-ordered or interrupted in the same way as before, it should prevent all of the TOCTU problems we have observed when using `send` so far.

This also makes the send queue fairer: a `try_send` and `send` can now no longer "steal" a `freeq` slot from a longer-waiting `send` future.

The unsafe parts of this function are a bit more scary, though.

### DIFF
--- a/rtic-sync/CHANGELOG.md
+++ b/rtic-sync/CHANGELOG.md
@@ -7,6 +7,9 @@ For each category, _Added_, _Changed_, _Fixed_ add new entries at the top!
 
 ## [Unreleased]
 
+### Fixed
+- Improve handling of free slots for `send` by explicitly writing the free slot to the awoken future.
+
 ## v1.3.1 - 2025-03-12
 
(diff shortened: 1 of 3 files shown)


datdenkikniet:
Validated that all of the bugs we've found so far are indeed fixed here with `loom` (which is & will be the next item on the agenda to merge):

1. `send` interrupting `try_recv` and "stealing" a `freeq` slot from the now-waked `send`.
2. `send` being interrupted by `Receiver::drop` after `is_closed` & panicking that way.

korken89 (approved):
Thank you!

PR #1035, opened 2025-03-15
rtic-sync: reorganize channel code for improved maintainability.

datdenkikniet:
Supersedes #1038 

Also reorganizes the `channel` files into more bite-sized chunks, and gives `struct Channel` all of the responsibilities.


rtic-sync/src/channel/sender.rs
```diff
@@ -0,0 +1,249 @@
...
+                if let Some(freed_slot) = unsafe { free_slot_ptr2.take(cs) } {
+                    // SAFETY: `freed_slot` is a free slot in our referenced channel.
+                    unsafe { self.0.return_free_slot(freed_slot) };
+                }
+            });
+        });
+
+        let idx = poll_fn(|cx| {
+            //  Do all this in one critical section, else there can be race conditions
+            critical_section::with(|cs| {
```
datdenkikniet, line 157: Logically, I have reason to believe that this critical section can now, indeed, be removed: all of the critical operations (obtaining a free slot, pushing to the wait queue) are handled atomically by the channel, and are now dependent on each other (a free queue slot is sent to a waiting future directly, so it no longer needs to "snipe" the slot out of the `freeq` independently).

For now, I think sticking them all in a single `critical-section` is the way to go. Once I get `loom` support up and running more, I will definitely investigate whether things do or do not break down if this critical section is removed.

PR #1037, opened 2025-03-16
Update esp32c3 to 0.28.0

rxwb:
The examples still indirectly depend on the old version via esp-hal. But the HAL version depending on the new esp32c3 version is still in beta for version 1.0.
Should esp-hal also be updated in the examples?


AfoHT (approved):
Nice! Thank you :)

AfoHT:
> Should esp-hal also be updated in the examples?

It would be awesome if you would like to do that :) I think that can be a follow up PR then

PR #1039, opened 2025-03-16
Don't force `Signal` import when using `make_signal` macro

cyril-marpaud:
The current source code of `make_signal` forces one to `use rtic_sync::signal::Signal;`.
This commit fixes that.

Also, I updated the documentation to mention that the created signal has a `'static` lifetime, just like for the `make_channel` macro.

### DIFF
--- a/rtic-sync/CHANGELOG.md
+++ b/rtic-sync/CHANGELOG.md
@@ -7,6 +7,9 @@ For each category, _Added_, _Changed_, _Fixed_ add new entries at the top!
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


korken89 (approved):
Thanks for fixing!

cyril-marpaud:
My pleasure, thx for merging so swiftly !
Would you merge the addition of a usage example to the documentation ?

AfoHT:
> Would you merge the addition of a usage example to the documentation ?

Usage examples and documentation are always welcome and appreciated!

PR #1040, opened 2025-03-17
rtic-sync: deal with dropping & re-splitting channels.

datdenkikniet:
Currently `split`-ing a channel is technically allowed, but there is no non-`panic` way of actually doing so.

Currently it will either `panic` if `freeq` is full (and `readyq` is empty), or succesfully re-split and always produce `Err(NoReceiver)` when sending, and/or `panic` eventually when returning a free slot while `recv`-ing. 

Additionally, items in the ready queue are never dropped if the channel is dropped.

This is not a particularly often-used use-case, but that doesn't mean we cannot try fixing it :)

An alternative approach is to just have a `split: bool` in the `Channel` and always `panic` if that bool is already set, or always calling `Channel::clear` when `split`-ing.

To see concrete use-cases that this PR fixes, please check out the introduced tests.

### DIFF
--- a/rtic-sync/src/channel.rs
+++ b/rtic-sync/src/channel.rs
@@ -102,62 +102,10 @@ impl<T, const N: usize> Channel<T, N> {
         }
     }
 
-    /// Clear any remaining items from this `Channel`.
-    pub fn clear(&mut self) {
-        for item in self.queued_items() {
-            drop(item);
-        }
-    }
-
-    /// Return an iterator over the still-queued items, removing them
-    /// from this channel.
-    pub fn queued_items(&mut self) -> impl Iterator<Item = T> + '_ {
-        struct Iter<'a, T, const N: usize> {
-            inner: &'a mut Channel<T, N>,
-        }
-
-        impl<T, const N: usize> Iterator for Iter<'_, T, N> {
-            type Item = T;
-
-            fn next(&mut self) -> Option<Self::Item> {
-                let slot = self.inner.readyq.as_mut().pop_back()?;
-
-                let value = unsafe {
-                    // SAFETY: `ready` is a valid slot.
-                    let first_element = self.inner.slots.get_unchecked(slot as usize).get_mut();
-                    let ptr = first_element.deref().as_ptr();
-                    // SAFETY: `ptr` points to an initialized `T`.
-                    core::ptr::read(ptr)
-                };
-
-                // NOTE: do not `return_free_slot`, as we have mutable
-                // access to this `Channel` and no `Receiver` or `Sender`
-                // exist.
-                debug_assert!(!self.inner.freeq.as_mut().is_full());
-                unsafe {
-                    // SAFETY: `freeq` is not ful.
-                    self.inner.freeq.as_mut().push_back_unchecked(slot);
-                }
-
-                Some(value)
-            }
-        }
-
-        Iter { inner: self }
-    }
-
     /// Split the queue into a `Sender`/`Receiver` pair.
-    ///
-    /// # Panics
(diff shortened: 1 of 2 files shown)


rtic-sync/src/channel.rs
```diff
@@ -77,8 +77,61 @@ impl<T, const N: usize> Channel<T, N> {
...
+                    // SAFETY: `ready` is a valid slot.
+                    let ptr = self.inner.slots.get_unchecked(slot as usize).get() as *const T;
+                    // SAFETY: `ptr` points to an initialized `T`.
+                    core::ptr::read(ptr)
+                };
+
+                // NOTE: do not `return_free_slot`, as we have mutable
+                // access to this `Channel` and no `Receiver` or `Sender`
+                // exist.
+                assert!(!self.inner.freeq.get_mut().is_full());
```
korken89: Should this be `debug_assert`?

rtic-sync/src/channel.rs
```diff
@@ -77,8 +77,61 @@ impl<T, const N: usize> Channel<T, N> {
         }
     }
 
+    /// Clear any remaining items from this `Channel`.
+    pub fn clear(&mut self) {
+        for _ in self.queued_items() {}
```
korken89: To help readability, lets do `drop(item)` in the loop.

rtic-sync/src/channel.rs
```diff
@@ -77,8 +77,61 @@ impl<T, const N: usize> Channel<T, N> {
...
     /// Split the queue into a `Sender`/`Receiver` pair.
+    ///
+    /// # Panics
+    /// This function panics if there are items in this channel while splitting.
+    ///
+    /// Call [`Channel::clear`] to clear all items from it, or [`Channel::queued_items`] to retrieve
+    /// an iterator that yields the values.
     pub fn split(&mut self) -> (Sender<'_, T, N>, Receiver<'_, T, N>) {
+        assert!(
+            self.readyq.get_mut().is_empty(),
```
korken89: Since we have `&mut channel` here, can't we do `self.clear()`?

korken89 (changes requested):
Very nice, thank you! Just some minor comments.

rtic-sync/src/channel.rs
datdenkikniet: Yes, we could. When writing this felt weird to drop the items in the queue on `split`, but not any earlier. Since most callers won't ever re-`split()` their queue I figured we could get away with never calling `clear` if they do not.

If you think we should `clear()`, should we do it in `Receiver::drop` too? It just feels kind of arbitrary to `drop` leftover items on `split()` and not earlier (i.e. `Receiver::drop`), but I have no clue what the usual thing to do is, here.

rtic-sync/src/channel.rs
datdenkikniet: Yes :) Done

rtic-sync/src/channel.rs
datdenkikniet: Done

rtic-sync/src/channel.rs
datdenkikniet: Have rebased on the latest `master`. Any opinions on the above?

korken89: Yeah, true. Sounds like `Receiver::drop` is where it should happen.

korken89:
@datdenkikniet was this ready for review again?

datdenkikniet:
Have rebased, and if I remember correctly I addressed all of the concerns, so should be good for another round.

korken89 (approved):
♥️ Thank you for fixing!

PR #1038, opened 2025-03-16
rtic-sync: add loom support & tests

datdenkikniet:
Adding `loom` support to `rtic-sync`.

Supersedes #1027

This PR, unlike the others, fixes no bugs and makes no end-user-functionality changes. It just adds `loom` support.

PR #1042, opened 2025-03-18
rtic-sync: No need for c-s to check if link is in wait queue if link is popped

datdenkikniet:
Minor improvement: if we insert our link into the wait queue, we always take a critical section (in the `OnDrop`) to ensure that it is no longer in the list on drop. 

However, if our link is popped, we know for a fact that it is not in the list. `take()`-ing it in the block that gets us into that state saves us an unnecessary critical section, which is probably the 2nd-most common scenario (the 1st being finding a free slot immediately).

Additionally, we take a critical-section to check whether we still have an entry in our `SlotPtr`. However, once we are certain that our `link` is removed from the list, we have exclusive access to it again: hence, we can skip that critical section too, on the happy path.

### DIFF
--- a/rtic-sync/CHANGELOG.md
+++ b/rtic-sync/CHANGELOG.md
@@ -7,8 +7,9 @@ For each category, _Added_, _Changed_, _Fixed_ add new entries at the top!
 
 ## [Unreleased]
 
+- Avoid a critical section when a `send`-link is popped and when returning `free_slot`.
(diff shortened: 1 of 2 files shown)


korken89 (approved):
LGTM! Thank you!

datdenkikniet:
This will block merging of #1038, so will wait until that is in :)

ETA: I am entirely lying. Will merge this

PR #1044, opened 2025-03-22
Simplify CI & expand coverage

datdenkikniet:
Instead of having a bunch of duplicate tasks, we can simply use/abuse the github matrices a little.

There's a bunch of repetition for platform/backend/toolchain combinations, but at least it's obvious what is and isn't being tested. We should be able to replace them with a YAML anchor [once that's supported][0]

Also add coverage for RISCV platforms/backends in clippy and check.

[0]: https://github.com/actions/runner/issues/1182

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,33 +69,32 @@ jobs:
   # Compilation check (hifive1) TODO
   # checkhifive1:
 
-  # Clippy (lm3s6965)
-  clippylm3s6965:
-    name: clippy (lm3s6965)
+  # Clippy
+  # TODO: clippy hifive1, esp32-c3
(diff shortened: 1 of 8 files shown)


korken89 (approved):
Nice improvement! The CI files were getting unwieldy. The only thing I think we should do in the future is to move as much as possible of all this into `xtask` or similar so one can do this locally and not need to have CI do it every time.

datdenkikniet:
Yeah, this is initial clean up. I plan to move the book scripts into actual script files, and have asked @AfoHT to look over the QEMU setup (since I'm not sure the prebuilt binaries are really being used).

ETA: ah, and yes, moving this matrix into `xtask` would be good.

PR #1046, opened 2025-03-24
Rebased: Added esp32c6 support

tshakah:
This is _mostly_ a rebase of the [previous](https://github.com/rtic-rs/rtic/pull/917) ESP32-C6 PR, updated to fix merge conflicts and implement the same structural changes that the ESP32-C3 implementation had.

A big difference is due to changes in the esp-hal and esp32c6 crates, as the interrupt handling was broken.

I haven't been able to get the qemu tests working locally, but have added the expected output and tested on hardware


examples/esp32c6/README.md
```diff
@@ -0,0 +1,39 @@
+### ESP32-C6 RTIC template
+This crate showcases a simple RTIC application for the ESP32-C6.
+
+## Prerequisites
+
+# Nightly Rust
+The ESP32-C6 HAL requires a nightly build of Rust.
```
jessebraham: FYI this is no longer the case, and the HAL can be built using `stable` Rust.

(I imagine this is just an artifact from the previous PR on which this one is based, given its age)

examples/esp32c6/rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "nightly-2023-11-14"
```
AfoHT: Perhaps this can be something a bit more modern, as indicated by previous comment `stable` should work nowadays?

examples/esp32c6/runner.sh
```diff
@@ -0,0 +1,37 @@
...
+qemuexec=qemu-system-riscv32
+
+# Building ESP32-C6 image
+espflash save-image --chip esp32c6 --merge "$outputfilenamecargo" "$outputfilename" 1>&2
+
+# Get stats
+esptool.py image_info --version 2 "$outputfilename" 1>&2
+
+# Run in QEMU
+$qemuexec -nographic -monitor tcp:127.0.0.1:55555,server,nowait -icount 3 -machine esp32c6 -drive file="$outputfilename",if=mtd,format=raw -serial file:"$logfile" &
```
AfoHT: At present there is no C6 support in [ESP-QEMU](https://github.com/espressif/qemu/releases), and QEMU reports there is no such machine.

We can keep this file, in case they ever add support, but considering it is a copy from C3 it might as well just be copied over if support is added.

Thus I propose you default runner to "the real chip runner", as we won't have QEMU in CI anyhow :)

examples/esp32c6/.cargo/config.toml
```diff
@@ -1,5 +1,9 @@
 [target.riscv32imac-unknown-none-elf]
-runner = "espflash flash --monitor"
+# Real hardware
+# runner = "espflash flash --monitor"
+
+# QEMU emulator
```
AfoHT: See other comment, sadly no QEMU support as of now

rtic-macros/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 ## [Unreleased]
 
 - Adapt `slic` backends to new version with `mecall`
+- Unstable support for ESP32-C6
```
AfoHT, line 10: Please put new entries above, so above "- Adapt ..."

Same for all changelog files

tshakah:
Thanks for the reviews, I am very busy with work this week but will pick this up again after tomorrow

AfoHT (approved):
Well done!

tshakah:
Looks like your merge queue is stuck?

PR #1048, opened 2025-03-27
examples/esp32c3: Update esp-hal to 1.0.0-beta.0

rxwb:
As wished for in #1037 here comes the follow up.
No big changes, only the input pin configuration needed a slight update.


AfoHT (approved):
Nice! Thank you :)

PR #1049, opened 2025-04-01
fix(doc): fix typo in file names to match with docs

ckrenslehner:
https://rtic.rs/2/book/en/by-example/tips/static_lifetimes.html

The book wants to include `static-resources-in-init` but the file name was `static-recources-in-init`

grafik


AfoHT (approved):
Thank you!

PR #1050, opened 2025-04-02
fix(doc): add safety notices to `Sender::send` and `Arbiter::access`

Ddystopia:
Hello, those methods are technically unsound, because you can create the future, `poll` it, it will register the link to the queue, than you can `core::mem::forget` that future and return from the function, that will deallocate stack frame. But other links will still have those pointers, causing use after free and memory corruption.


Ddystopia:
Whoops, it's a mistake on my side, that `link` is not borrowed but owned by the future, thus pin guarantees do apply

PR #1051, opened 2025-04-02
feat: do not cause memory corruption if `remove_from_list` (safe function) is called multiple times

Ddystopia:
`Link::remove_from_list` is safe, but if you will call it several times, it will again change the list. If it was changed in between those calls, state would be corrupted. This can be fixed either by setting `self.next` and `self.prev` to `null_mut()` after removing the node, or by using already present `is_popped` boolean.

### DIFF
--- a/rtic-common/src/wait_queue.rs
+++ b/rtic-common/src/wait_queue.rs
@@ -141,11 +141,12 @@ impl<T: Clone> Link<T> {
     /// Remove this link from a linked list.
     pub fn remove_from_list(&self, list: &DoublyLinkedList<T>) {
         cs::with(|_| {
+            // Make sure all previous writes are visible
+            core::sync::atomic::fence(Ordering::SeqCst);
+
             if self.is_popped() {
                 return;
             }
-            // Make sure all previous writes are visible
-            core::sync::atomic::fence(Ordering::SeqCst);
 
             let prev = self.prev.load(Self::R);
             let next = self.next.load(Self::R);
--- a/rtic-common/src/wait_queue.rs
+++ b/rtic-common/src/wait_queue.rs
@@ -140,6 +140,10 @@ impl<T: Clone> Link<T> {
 
     /// Remove this link from a linked list.
     pub fn remove_from_list(&self, list: &DoublyLinkedList<T>) {
+        if self.is_popped() {
+            return;
+        }
+
         cs::with(|_| {
             // Make sure all previous writes are visible
             core::sync::atomic::fence(Ordering::SeqCst);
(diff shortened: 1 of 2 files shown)


Ddystopia:
I am not sure why CI fails, it doesn't look related to the PR

korken89 (dismissed):
Thanks for fixing!

rtic-common/src/wait_queue.rs
```diff
@@ -140,6 +140,10 @@ impl<T: Clone> Link<T> {
 
     /// Remove this link from a linked list.
     pub fn remove_from_list(&self, list: &DoublyLinkedList<T>) {
+        if self.is_popped() {
```
datdenkikniet: I believe this is a good change, however, for this to be truly sound, this check has to be within the critical section.

Imagine the scenario where:

1. Thread 1 calls `remove_from_list`, checks `is_popped`, sees a `false` and is interrupted before entering the critical section.
2. Thread 2 starts, calls `remove_from_list`, checks `is_popped` and also sees a `false`.
3. The same problem as before happens (if I have understood the problem correctly).

datdenkikniet (changes requested):

rtic-common/src/wait_queue.rs
datdenkikniet: Something like this would also greatly benefit from a `loom` test, I think. Is that something you'd be willing to add? If not, I can do it in a different PR.

It would also require moving `loom_cs` into `rtic-common`, so maybe more work than this PR should have for now.

Ddystopia: Good catch! I'll move it into the critical section. As for the loom test, I don't have experience with it so it would be great if you did it.

rtic-common/src/wait_queue.rs
```diff
@@ -141,6 +141,9 @@ impl<T: Clone> Link<T> {
     /// Remove this link from a linked list.
     pub fn remove_from_list(&self, list: &DoublyLinkedList<T>) {
         cs::with(|_| {
+            if self.is_popped() {
```
datdenkikniet, line 147: Sorry to double-ask, will need @korken89 confirmation/insight here: I think it'll also need to be after the compiler fence, to ensure that the atomic write is observed.

This is probably not a problem on ARM in practice, but for good measure :)

`atomic::Ordering::Relaxed` is weird, so in theory having a critical section does not guarantee ordering when it comes to atomic reads/writes.

datdenkikniet (changes requested):

datdenkikniet (approved):

PR #1052, opened 2025-04-03
feat: `wait_until` method for the wait_queue

Ddystopia:
Diatomic Waker has a [method like this](https://docs.rs/diatomic-waker/latest/diatomic_waker/struct.WakeSinkRef.html#method.wait_until) and it is very handy. I've been implementing async wrapper over [heapless::mpmc::MpMcQueue](https://docs.rs/heapless/latest/heapless/mpmc/struct.MpMcQueue.html) and wait_queue is great, but a method like this is really benefitial. You can *safely* create async versions of `enqueue` and `dequeue`, and don't need `unsafe` to interact with the `wait_queue`.

Note that this PR has only 1 additional `unsafe` block.

To give a perspective, this method is used like that :

```rust
impl<T, const N: usize> MpMcChannel<T, N> {
    pub const fn new() -> Self {
        Self {
            tx_wait_queue: WaitQueue::new(),
            rx_wait_queue: WaitQueue::new(),
            queue: MpMcQueue::new(),
        }
    }

    pub fn try_send(&self, value: T) -> Result<(), T> {
        while let Some(waker) = self.rx_wait_queue.pop() {
            waker.wake();
        }

        self.queue.enqueue(value)
    }

    pub fn try_recv(&self) -> Option<T> {
        while let Some(waker) = self.rx_wait_queue.pop() {
            waker.wake();
        }

        self.queue.dequeue()
    }

    pub async fn send(&self, value: T) {
        let mut value = Some(value);
        self.tx_wait_queue
            .wait_until(|| {
                self.try_send(value.take().expect("Future is polled after completion"))
                    .map_err(|old_val| value = Some(old_val))
                    .ok()
            })
            .await
    }

    pub async fn recv(&self) -> T {
        self.rx_wait_queue.wait_until(|| self.try_recv()).await
    }
}
```

### DIFF
--- a/rtic-common/src/wait_queue.rs
+++ b/rtic-common/src/wait_queue.rs
@@ -141,11 +141,12 @@ impl<T: Clone> Link<T> {
     /// Remove this link from a linked list.
     pub fn remove_from_list(&self, list: &DoublyLinkedList<T>) {
         cs::with(|_| {
+            // Make sure all previous writes are visible
+            core::sync::atomic::fence(Ordering::SeqCst);
+
             if self.is_popped() {
                 return;
             }
-            // Make sure all previous writes are visible
-            core::sync::atomic::fence(Ordering::SeqCst);
 
             let prev = self.prev.load(Self::R);
             let next = self.next.load(Self::R);


Ddystopia:
> error: called `Iterator::last` on a `DoubleEndedIterator`; this will needlessly iterate the entire iterator
  --> rtic-macros/src/lib.rs:91:54
   |
91 |                 if let Some(dir) = path.components().last() {
   |                                                      ^^^^^^ help: try: `next_back()`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#double_ended_iterator_last
   = note: `-D clippy::double-ended-iterator-last` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::double_ended_iterator_last)]`

That one is unrelated to this PR

AfoHT:
Really interesting! Nice with the example in the PR, is that something you could add somewhere in the CI/examples that are run too?

And something about the administrative, instead of merging in the `master` branch in the PR, could you rebase on top of the master branch instead?

PR #1055, opened 2025-04-13
fix edition 2024 unsafe link section

Ddystopia:
In edition 2024, we must use `#[unsafe(link_section = ...)]` instead of `#[link_section = ...]`. But `rtic_macros` is only checking for `#[link_section = ...]`.

- **fix: in edition 2024 `link_section` is used as `unsafe(link_section = ...)`**
- **chore: changelog entry**


rtic-macros/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 - Adapt `slic` backends to new version with `mecall`
 - Allow software tasks to be diverging (return `!`) and give them `'static` context.
+- Fix: supoort edition 2024 `unsafe(link_section)` attribute
```
AfoHT:
```suggestion
- Fix: supoort edition 2024 `unsafe(link_section)` attribute
- Adapt `slic` backends to new version with `mecall`
- Allow software tasks to be diverging (return `!`) and give them `'static` context.
```

AfoHT:
Really nice you spotted it and got it sorted! Thank you :rocket: 

Would be amazing if this could be caught in the future with some UI test or similar, something you could look into perhaps?

Ddystopia:
Oh, I am really now sure how to use existing framework to do this kind of test. When developing this, I used `objdump` and `cargo expand` to see there the symbols were put. `unsafe(link_section)` was ignored by rtic and my symbols ended up  in the wrong section

Ddystopia:
Oh, I probably found a way. Rust gives warning if rtic is not using it

Ddystopia:
Is this possible to make a ui test in `2024` edition while not changing the edition of `rtic` itself? Because it seems to me, that to have a ui test in a new edition, I must migrate `rtic` to the new edition.

Ddystopia:
I honestly didn't find a better way to tests this. Rust would issue additional warning if attribute would get ignored, and that will cause the test to fail.

Ddystopia:
@AfoHT I think this is some flaky test, that test failes and fixes randomly

AfoHT:
I forgot to update this ticket, but previous week I spent some time on trying to get CI to pass, there's a cargo clippy PR in the pipe and then a few additional places that popped up needing handling. 

Should be similar to what this PR already does though. 

It's not forgotten, just slow 😅

korken89 (approved):

PR #1057, opened 2025-05-15
feat: `wait_until` method for the wait_queue 

Ddystopia:
For some reason github cannot understand that this is the same branch, so I forced to make the second PR after #1052

Diatomic Waker has a [method like this](https://docs.rs/diatomic-waker/latest/diatomic_waker/struct.WakeSinkRef.html#method.wait_until) and it is very handy. I've been implementing async wrapper over [heapless::mpmc::MpMcQueue](https://docs.rs/heapless/latest/heapless/mpmc/struct.MpMcQueue.html) and wait_queue is great, but a method like this is really benefitial. You can *safely* create async versions of `enqueue` and `dequeue`, and don't need `unsafe` to interact with the `wait_queue`.

Note that this PR has only 1 additional `unsafe` block.

To give a perspective, this method is used like that :

```rust
impl<T, const N: usize> MpMcChannel<T, N> {
    pub const fn new() -> Self {
        Self {
            tx_wait_queue: WaitQueue::new(),
            rx_wait_queue: WaitQueue::new(),
            queue: MpMcQueue::new(),
        }
    }

    pub fn try_send(&self, value: T) -> Result<(), T> {
        // wakes should actually be after successful enqueue
        while let Some(waker) = self.rx_wait_queue.pop() {
            waker.wake();
        }

        self.queue.enqueue(value)
    }

    pub fn try_recv(&self) -> Option<T> {
        // wakes should actually be after successful dequeue
        while let Some(waker) = self.rx_wait_queue.pop() {
            waker.wake();
        }

        self.queue.dequeue()
    }

    pub async fn send(&self, value: T) {
        let mut value = Some(value);
        self.tx_wait_queue
            .wait_until(|| {
                self.try_send(value.take().expect("Future is polled after completion"))
                    .map_err(|old_val| value = Some(old_val))
                    .ok()
            })
            .await
    }

    pub async fn recv(&self) -> T {
        self.rx_wait_queue.wait_until(|| self.try_recv()).await
    }
}
```

### DIFF
--- a/ci/expected/lm3s6965/wait-queue.run
+++ b/ci/expected/lm3s6965/wait-queue.run
@@ -1,0 +1,4 @@
+Inc
+Inc
+Inc
+Got 3
--- a/examples/lm3s6965/Cargo.toml
+++ b/examples/lm3s6965/Cargo.toml
@@ -16,6 +16,7 @@ bare-metal = "1.0.0"
 cortex-m-semihosting = "0.5.0"
 rtic-time = { path = "../../rtic-time" }
 rtic-sync = { path = "../../rtic-sync" }
+rtic-common = { path = "../../rtic-common" }
 rtic-monotonics = { path = "../../rtic-monotonics", features = ["cortex-m-systick"] }
 rtic = { path = "../../rtic" }
 cfg-if = "1.0"
--- a/examples/lm3s6965/examples/wait-queue.rs
+++ b/examples/lm3s6965/examples/wait-queue.rs
@@ -1,0 +1,64 @@
+//! examples/wait-queue.rs
+
+#![no_main]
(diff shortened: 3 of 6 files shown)


Ddystopia:
@AfoHT #1052 moved there

korken89 (dismissed):
This was a really nice way to handle the dropper, I'll look into using this in `Channel` and any other place I'll find.

AfoHT:
All is good, just Cargo.lock built in the past (at the time correct) made git rebasing problematic. Rebased locally and removed Cargo.lock, that can be separate/we'll handle it. Thank you for your work!!! :rocket:

AfoHT (approved):

PR #1074, opened 2025-06-10
docs: rtic-time, monotonic, monotonics

### LINKED ISSUE #1061: rtic-monotonic vs rtic-time & rtic-monotonics
The current state of the vital trait `Monotonic`'s definitions between RTIC versions is incredibly confusing. 

For RTIC version 1 you have [rtic-monotonic](https://lib.rs/crates/rtic-monotonic), which itself has stayed at version 1.0.0. Then, you have the similarly named [rtic-monotonic**s**](https://lib.rs/crates/rtic-monotonics) (with an s) at version 2.0.3 which does not redefine the `Monotonic` trait or give more traits like it as one might expect from the name, but instead gives actual PAC-based implementations. You instead have to look at [rtic-time](https://lib.rs/crates/rtic-time) to see the current (2.0.0) `Monotonic` definition that is actually used for the up-to-date RTIC version. This is needlessly confusing.

My proposed solutions in order of preference:
1. Update [rtic-monotonic](https://lib.rs/crates/rtic-monotonic) to 2.0.0 with the content of [rtic-time](https://lib.rs/crates/rtic-time) then yank [rtic-time](https://lib.rs/crates/rtic-time) from crates.io as it is no longer needed.
2. Rename [rtic-monotonic**s**](https://lib.rs/crates/rtic-monotonics) to something less silly and stick with [rtic-time](https://lib.rs/crates/rtic-time) as the successor to [rtic-monotonic](https://lib.rs/crates/rtic-monotonic).
3. Just update the READMEs of all the crates to be even more clear about which version of RTIC they concern and what their purposes are.
 
The first solution is in my eyes the most desirable as it'd prompt HAL-maintainers to look into implementing support for RTIC 2.0 and thus render the need for a separate crate ([rtic-monotonic**s**](https://lib.rs/crates/rtic-monotonics)) moot, while avoiding the problem of introducing even more crates to maintain.

AfoHT:
- **docs: Monotonics crates: Fix old copy-paste**
- **docs: Monotonics: Highlight rtic-time is used**

The newcomer to this space usually fails to relate the term "monotonic" with their needs of dealing with "time".

Thus the crate `rtic-time`:

> Basic definitions and utilities that can be used to keep track of time.

bridges the gap towards the `Monotonic` trait inside.

It may not have been intentional to name the reference implementations `rtic-monotonic*s*`,
but considering the crate contains a whole set of different implmentations targeting various producs it might be the better choice. And that it distinctively does not provide the `Monotonic` trait sets it apart from the old RTIC v1 `rtic-monotonic`.

Ideally, with time whatever was in here can be absorbed into each products relevant HAL crate(s).

Related PR: https://github.com/rtic-rs/rtic-monotonic/pull/12

Closes #1061

### DIFF
--- a/rtic-monotonics/README.md
+++ b/rtic-monotonics/README.md
@@ -3,7 +3,7 @@
 
 # `rtic-monotonics`
 
-> Core abstractions of the Real-Time Interrupt-driven Concurrency (RTIC) Monotonics timers
+> Reference implementations of the Real-Time Interrupt-driven Concurrency (RTIC) Monotonics timers
 
 `rtic-monotonics` is for RTIC v2.
 
--- a/rtic-monotonics/README.md
+++ b/rtic-monotonics/README.md
@@ -5,6 +5,8 @@
 
 > Reference implementations of the Real-Time Interrupt-driven Concurrency (RTIC) Monotonics timers
 
+Uses [`rtic-time`](https://github.com/rtic-rs/rtic/tree/master/rtic-time) defined [`Monotonic`](https://docs.rs/rtic-time/latest/rtic_time/trait.Monotonic.html) trait.
+
 `rtic-monotonics` is for RTIC v2.
 
 For RTIC v1 see [`rtic-monotonic`](https://github.com/rtic-rs/rtic-monotonic)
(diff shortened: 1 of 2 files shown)


korken89 (approved):
🚀

PR #1075, opened 2025-06-14
Correct timer type in "Delay and Timeout using Monotonics"

thejpster:
The example in `lm3s6965/examples/async-timeout.rs` uses `Mono` as the monotonic timer type, so it's confusing that the second example switches to using `Systick`.

### DIFF
--- a/book/en/src/by-example/delay.md
+++ b/book/en/src/by-example/delay.md
@@ -21,7 +21,7 @@ A _software_ task can `await` the delay to expire:
 #[task]
 async fn foo(_cx: foo::Context) {
     ...
-    Systick::delay(100.millis()).await;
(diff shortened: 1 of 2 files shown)


thejpster:
I was still confused when reading this chapter, so I've had a go a re-wording some of it.

thejpster:
The systick! macro docs were also not very helpful, to avoid people having to read the source or do a `cargo expand`, I've tried to add some more explanatory text here too.

korken89 (dismissed):
Nice updates, thank you! Can you check the `fmt` failure?

thejpster:
Force pushed with fixed formatting

AfoHT (approved):

PR #1079, opened 2025-06-15
Require exclusive access to the RESET peripheral for RP2040/RP2350

thejpster:
If you don't have exclusive access to the RESET peripheral, someone else could race you on the read/modify/write cycle on the update of the RESETS.reset register.

This is a breaking change.


thejpster:
rebased on master

thejpster:
Hmmm. RESETS is [Send but !Sync](https://docs.rs/rp2040-pac/latest/rp2040_pac/struct.RESETS.html) so this change might be moot.

korken89:
Yeah, it's an oversight that it's not `&mut` but the bounds you point out makes it not horrible at least.

PR #1085, opened 2025-06-24
[rtic-monotonics] Panic if STM32 clock prescaler value overflows

antonok-edm:
I recently had a confusing issue on my STM32 board where I used a TIMx driver for `Monotonic` and the timer was progressing ~5.5x faster than it should have been. The system clock was 80MHz and I used 1000 ticks per second for the timer. Turns out there was an overflow in the prescaler configuration. This PR adds an assertion in that case, to make debugging this problem easier for other users in the future.

### DIFF
--- a/rtic-monotonics/CHANGELOG.md
+++ b/rtic-monotonics/CHANGELOG.md
@@ -7,6 +7,9 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ## Unreleased
 
+### Changed
+- Panic if STM32 prescaler value would overflow
+
 ## v2.1.0 - 2025-06-22
 
 ### Changed
--- a/rtic-monotonics/src/stm32.rs
+++ b/rtic-monotonics/src/stm32.rs
@@ -238,8 +238,10 @@ macro_rules! make_timer {
                 $timer.cr1().modify(|r| r.set_cen(false));
 
                 assert!((tim_clock_hz % timer_hz) == 0, "Unable to find suitable timer prescaler value!");
-                let psc = tim_clock_hz / timer_hz - 1;
-                $timer.psc().write(|r| r.set_psc(psc as u16));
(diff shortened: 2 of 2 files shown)


rtic-monotonics/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 - Updated esp32c3 dependency to v0.28.0
 - Updated esp32c3 dependency to v0.27.0
+- Panic if STM32 prescaler value would overflow
```
AfoHT: This changelog entry does not belong to this release. 

You should create a new "changed" header under "unreleased" header and place it there instead :)

antonok-edm: ah, I did it too quickly :sweat_smile: 

should be fixed now

rtic-monotonics/src/stm32.rs
```diff
@@ -238,8 +238,8 @@ macro_rules! make_timer {
                 $timer.cr1().modify(|r| r.set_cen(false));
 
                 assert!((tim_clock_hz % timer_hz) == 0, "Unable to find suitable timer prescaler value!");
-                let psc = tim_clock_hz / timer_hz - 1;
-                $timer.psc().write(|r| r.set_psc(psc as u16));
+                let psc = u16::try_from(tim_clock_hz / timer_hz - 1).expect("Clock prescaler overflowed!");
```
korken89: To not have formatting bloat unconditionally added.

```suggestion
                let Ok(psc) = u16::try_from(tim_clock_hz / timer_hz - 1) else {
                    panic!("Clock prescaler overflowed!");
                };
```

korken89 (dismissed):

rtic-monotonics/src/stm32.rs
antonok-edm: thanks, updated

AfoHT (approved):

PR #1094, opened 2025-08-19
cortex-m-source-masking: use global critical section when resource is shared with exception handler

### LINKED ISSUE #1088: Unsoundness when using shared resources in exception handlers on ARMv6-M
According to the RTIC documentation ([source](https://rtic.rs/2/book/en/internals/targets.html)):
> [...] ARMv6-M/ARMv8-M-base cannot have tasks with shared resources bound to exception handlers, as these cannot be masked in hardware.

This sentence seems to indicate that on ARMv6-M something like this won't work (and indeed on RTIC v1 this produced a compile-time error):
```rust
#[task(binds = SysTick, shared = [peripheral])]
fn watchdog(mut ctx: watchdog::Context) {
    ctx.shared.peripheral.lock(|p| p.whatever());
}
```

However, with RTIC v2 this code compiles just fine on `thumbv6m-none-eabi` with the `thumbv6-backend` feature enabled.

What's going on here? Is there a bug in RTIC? Is there outdated info in the documentation? Am I just completely misunderstanding something?

FeldrinH:
Fixes https://github.com/rtic-rs/rtic/issues/1088.

This implements a simple but somewhat heavy-handed fix for the issue described in https://github.com/rtic-rs/rtic/issues/1088:
If at least one of the tasks accessing a shared resource is an exception handler, then the access is protected by a global critical section, regardless of the priorities of tasks involved.

Unfortunately, this blocks all tasks, even those that have higher priority and do not access the shared resource. In the future it might be worthwhile to look into some way to mask the exception handlers (maybe by dynamically lowering exception priorities).

TODO:
- [x] ~~Update documentation to note the new behavior.~~
- [x] ~~Add note to changelog.~~
- [ ] Add some kind of QEMU test for the locking.

PS: I know that the CI is failing, but as far as I can tell this is caused by automatic updates to the Rust toolchain used in the CI, not by my changes.


perlindgren:
Thanks for working on this, I'll leave some general comments in the open issue.

/Per

PR #1098, opened 2025-09-03
Forward attributes applied to app module

datdenkikniet:
Instead of ignoring additional attributes applied to the `#[app]` module, we can forward them to the generated code to make the `mod app` act more like a normal `app`.

Also fixing some unrelated changes tests that have come as a result of updates to rustc stable, in separate commits. I'm trying my best to figure out in which version they stopped compiling, but can't seem to get TryBuild to use a different version of `cargo`.

### DIFF
--- a/rtic-macros/CHANGELOG.md
+++ b/rtic-macros/CHANGELOG.md
@@ -7,6 +7,10 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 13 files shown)


datdenkikniet:
QEMU seems to be quite broken, once again... I don't think I can fix that for this PR, but am certain that my changes do not affect how well QEMU (doesn't) function.

AfoHT (approved):
Nice!

I've tried the failed QEMU step (affected by the same thing as what you fixed in rtic-sync, thank you!) locally and it passed.

Merging :)

PR #1100, opened 2025-09-13
Update esp32c{3,6} support to esp-hal-1.0.0-rc.0.

pebender:
Thank you for RTIC.

I don't know whether or not anyone is interested, but I made an attempt at updating RTIC's esp32c3 and esp32c6 support to esp-hal-1.0.0-rc.0. I did it for a driver I am working on that includes an example using the driver with RTIC and and example using the driver with embassy.
[rtic.patch](https://github.com/user-attachments/files/22316168/rtic.patch)

As far as I could tell, there was nothing much to change and the changes are confined to esp32c3 and esp32c6 specific code.

The esp32c3 examples build and run on qemu. The esp32c6 examples build and run on target (there is no qemu support for the esp32c6).

There is one ci failure and one clippy failure that both exist in the unmodified code. They are both likely due to difference in compiler versions (I am running the 1.89.0). The ci failure is in ui/task-priority-too-high.rs. The test that is intended to fail to compile. It has the expected failure code and detailed information but the formatting of detailed information is somewhat different. The clippy failure is in rtic-sync/src/signal.rs. The compiler now wants lifetimes to be given. Compilation of xtask/src/cargo_command.rs has the same warning but it does not result in a clippy error because xtask source is not being checked. I have attached a diff of the changes I needed to make so that ci and clippy pass.

### DIFF
--- a/examples/esp32c3/.cargo/config.toml
+++ b/examples/esp32c3/.cargo/config.toml
@@ -1,16 +1,20 @@
 [target.riscv32imc-unknown-none-elf]
 # Real hardware
-# runner = "espflash flash --monitor"
+# runner = "espflash flash --monitor --chip esp32c3"
 
 # QEMU emulator
 runner = "./runner.sh"
 
 [build]
 rustflags = [
-  "-C", "link-arg=-Tlinkall.x",
   # Required to obtain backtraces (e.g. when using the "esp-backtrace" crate.)
   # NOTE: May negatively impact performance of produced code
   "-C", "force-frame-pointers",
+  # This should be last.
+  "-C", "link-arg=-Tlinkall.x",
 ]
 
 target = "riscv32imc-unknown-none-elf"
+
+[unstable]
+build-std = ["alloc", "core"]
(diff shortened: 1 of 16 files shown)


pebender:
A brief look at the CI failures indicate they are due to the same clippy problem I ran into that is a problem with the existing codebase and the current compiler.

AfoHT:
> A brief look at the CI failures indicate they are due to the same clippy problem I ran into that is a problem with the existing codebase and the current compiler.

Thank you for these, the reformatting and rtic-sync lifetime thing was also part of https://github.com/rtic-rs/rtic/pull/1098 that I just merged.

I did not see you had already sorted the xtask issue, so perhaps my tiny CI fix PR was not needed... https://github.com/rtic-rs/rtic/pull/1102

But it should be easy enough thanks to your power-move of keeping the CI fixups in separate commits :) Will try to get the ESP32 changes merged tonight :rocket:

AfoHT (approved):
Thank you! :)

PR #1102, opened 2025-09-17
CI: Fix xtask lifetime warning

AfoHT:
- **CI: esptool warns about deprecating esptool.py**
- **CI: xtask: Resolve warning about mixed lifetimes**


AfoHT:
Running `cargo xtask --deny-warnings --platform esp32-c3 --backend riscv-esp32-c3 qemu` locally passes all tests, this ESP QEMU seems to need some stern talking to, again :weary:

PR #1104, opened 2025-09-17
Add links to changelogs

datdenkikniet:
Instead of having bare, link-less references in our markdown files, we can have all of them! (And perhaps become better at linking new releases?)

Also, please squash this PR, I've just added each commit separately so it's not as messy.

Script:

<details>
<summary>Awful shell script</summary>

```bash
#!/bin/bash

# set -x

get_version() {
    revision="$1"
    file="$2"

    git show $revision:"$file" | grep "version =" | head -n 1 | cut -d ' ' -f3 | cut -d '"' -f2
}

get_commit() {
    revision="$1"
    file="$2"

    git blame --abbrev=40 $revision -- "$file" | grep "version = " | head -n 1 | cut -d ' ' -f1
}

file="$1"
changelog="$2"
current_commit=$(get_commit HEAD "$file")
current_version=$(get_version $current_commit "$file")

echo "Commit for v$current_version: $current_commit"

while [ $(git show $current_commit~1:"$file" > /dev/null 2>&1; echo $?) = "0" ]; do
    current_commit=$(get_commit $current_commit~1 "$file")
    current_version=$(get_version $current_commit "$file")

    tag=$(git describe $current_commit)
    echo "Commit for v$current_version: $current_commit."
    escaped_version=$(echo -n "$current_version" | sed "s/\\./\\\\./g")
    url="https://github.com/rtic-rs/rtic/commit/$current_commit"
    # I hate it, but it works :)
    sed -E -i "s@## (\\[v$escaped_version\\]|v$escaped_version)(.*)\$@## [v$current_version]($url)\\2@g" "$changelog"
done

```

### DIFF
--- a/rtic-common/CHANGELOG.md
+++ b/rtic-common/CHANGELOG.md
@@ -25,8 +25,8 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ### Fixed
 
-## [v1.0.1]
+## [v1.0.1](https://github.com/rtic-rs/rtic/commit/2b2208e217a96086696bd6f36cff2a6cd4c4ac9f)
 
 - `portable-atomic` used as a drop in replacement for `core::sync::atomic` in code and macros. `portable-atomic` imported with `default-features = false`, as we do not require CAS.
 
-## [v1.0.0] - 2023-05-31
+## [v1.0.0](https://github.com/rtic-rs/rtic/commit/e65e532c2a342f77080ac6fc8e5be11aa7d82575)(https://github.com/rtic-rs/rtic/commit/c3884e212c36d2a9cf260b1d9ae37c92b91ea73d) - 2023-05-31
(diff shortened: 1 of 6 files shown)


datdenkikniet:
OK the script is doing some bad stuff, may fix it later (but not right now).

AfoHT:
Interesting approach, if we disregard that the script at this moment isn't doing what you wanted it to do:

Isn't the git tags meant to make the unsightly hashes a bit more kind on those who need to edit the changelogs (currently imposed on all contributors, not just maintainers)?

It does happen that we forget to tag releases, but it's easy enough to rectify, even back-dating them is not that hard. Perhaps this step even helps us, and encourage proper tagged releases :)

I like this approach though, especially in contrast to that diff-thingy in the footer going back to the RTIC stone ages that was yanked not that long ago was too much bother for little gain xD

Added tags now for [v2.1.2](https://github.com/rtic-rs/rtic/releases/tag/v2.1.2), [v2.1.3](https://github.com/rtic-rs/rtic/releases/tag/v2.1.3) and [v2.2.0](https://github.com/rtic-rs/rtic/releases/tag/v2.2.0) :smile:

AfoHT:
I propose we close this, with the following reasoning:

* Links should be to the tags, not the commit hashes 
* Try to keep the changelog simple
  * These links add some effort for upkeep with or without scripting/tools to help
* Avoid adding one additional "anchor" to GitHub

PR #1106, opened 2025-10-03
trampoline functionalty

Spooky-Firefox:
### Trampoline Feature for Hardware Tasks

This pull request introduces a new **trampoline** feature for hardware tasks in RTIC.
It allows an interrupt handler to pend another interrupt (the *trampoline*), which then executes the actual task code.
This approach improves interrupt handling flexibility and source masking safety (as some exceptions cannot be masked through the NVIC).

The implementation includes changes to parsing, code generation, and error checking for correct usage.
It also includes a small CI fix and a minor clippy compliance improvement to the `rtic-monotonics` crate.

---

### Overview

**Trampoline Behavior**

Normally, when using the source masking backend, we expect that a lower-priority exception waits until the higher-priority task finishes:

```text
  lock     |     _______________
           |     |             |
  exception|     |             | =========
  task     | ____| _  _  _  _  |_
           |_______________________________
```

However, because we artificially raise priority by disabling interrupts (not exceptions), the effective task priority remains unchanged.
This means that lower-priority exceptions can still preempt the task:

```text
  lock     |     ______          ______
           |     |                    |
  exception|     |     ==========     | 
  task     | ____| _  _          _  _ |_
           |_______________________________
```

The trampoline mechanism solves this by pended interrupts that are properly masked by the source masking backend, preserving the Stack Resource Policy (SRP):

```text
  lock     |     ______      ______
           |     |                |
  exception|     |     =pend=     | ~~trampoline irq~~
  task     | ____| _  _      _  _ |_
           |___________________________________________
```

As illustrated, there is a brief priority inversion when pended, but since no critical section is executed during this window, it is not problematic.
We also do not enforce a trampoline when shared resources are not used — these cases still involve a minor priority inversion, but since shared data is untouched, this trade-off was accepted.

For more discussion, see [issue #1088](https://github.com/rtic-rs/rtic/issues/1088).

---

### Implementation Details

* Added support for the `trampoline` argument in hardware task definitions, allowing an interrupt handler to pend another interrupt that executes the task logic.
  [[1]](https://github.com/rtic-rs/rtic/pull/1106/files#diff-b6dc308038f110374a277cf42acfb20d19f30bca1cf30e7bd41528e465030e93R227-R242)
  [[2]](https://github.com/rtic-rs/rtic/pull/1106/files#diff-34953ff60f9179b03ac164a501e3396172b7e3c7c3831c58b8b895beca7925f1R301-R303)
* Updated code generation: when a trampoline is specified, the original interrupt handler only pends the trampoline, which then runs the user’s task code.
  [[1]](https://github.com/rtic-rs/rtic/pull/1106/files#diff-20d738b7b62f50c8c036ade301eaa17e9c66da51fccca21d539bcb444fb24c64R28-R50)
  [[2]](https://github.com/rtic-rs/rtic/pull/1106/files#diff-20d738b7b62f50c8c036ade301eaa17e9c66da51fccca21d539bcb444fb24c64R61-L54)
* Improved source masking logic to account for trampolines and added checks preventing illegal combinations of exceptions, trampolines, and shared resources.
  [[1]](https://github.com/rtic-rs/rtic/pull/1106/files#diff-d5902ea97b19f8ce617060fd709d3e5bce97f124ab11c4c86dc1ab6869c66354L105-R118)
  [[2]](https://github.com/rtic-rs/rtic/pull/1106/files#diff-d5902ea97b19f8ce617060fd709d3e5bce97f124ab11c4c86dc1ab6869c66354R257-R268)
  [[3]](https://github.com/rtic-rs/rtic/pull/1106/files#diff-d5902ea97b19f8ce617060fd709d3e5bce97f124ab11c4c86dc1ab6869c66354R378-R402)

---

### Documentation and Changelog Updates

* Updated changelogs to include the trampoline feature and its interaction with source masking.
  [[1]](https://github.com/rtic-rs/rtic/pull/1106/files#diff-2687262b53a875baa24bac87b2b1551239b8726f3e871a591c61c64b5e139a36R13-R15)
  [[2]](https://github.com/rtic-rs/rtic/pull/1106/files#diff-eb5b7735d327102d329bd71086282b67b4f503c11238bfa3e92e22118b7a76f9R26)

---

### Miscellaneous Improvements

* Changed modulo operations to `is_multiple_of()` in `rtic-monotonics` for clippy compliance.
  [[1]](https://github.com/rtic-rs/rtic/pull/1106/files#diff-b6cbf6dc4cb3d0b823e54ddeb94539639d57a6bcbea7863d36d7c074c7d89cbbL8-R18)
  [[2]](https://github.com/rtic-rs/rtic/pull/1106/files#diff-1ea01758c49cc52a84867079df6484f9af8308985863b4c3a2038cd7fba92ad9L77-R77)
* Fixed a CI issue caused by a leading space in a test output file.

---

### Summary

These changes collectively enhance the flexibility and safety of interrupt handling in RTIC while improving CI reliability and code linting compliance.

### DIFF
--- a/examples/lm3s6965/examples/trampoline.rs
+++ b/examples/lm3s6965/examples/trampoline.rs
@@ -1,0 +1,65 @@
+//! examples/bouncy_trampoline.rs
+
+#![no_std]
+#![no_main]
+#![deny(warnings)]
+#![deny(unsafe_code)]
+#![deny(missing_docs)]
+
+use panic_semihosting as _;
+
+// `examples/bouncy_trampoline.rs` testing trampoline feature 
+#[rtic::app(device = lm3s6965)]
+mod app {
+    use cortex_m_semihosting::{debug, hprintln};
+    use lm3s6965::Interrupt;
+
+    #[shared]
+    struct Shared {
+        h_priority: u8,
+    }
+
+    #[local]
+    struct Local {}
+
+    #[init]
+    fn init(_: init::Context) -> (Shared, Local) {
+        // Trigger SW0 interrupt to test NMI preemption
+        rtic::pend(Interrupt::GPIOA);
+        (Shared { h_priority: 0 }, Local {})
+    }
+
+    #[idle]
+    fn idle(_: idle::Context) -> ! {
+        hprintln!("idle");
+
+        // pend another interrupt to test trampoline
+        cortex_m::peripheral::SCB::set_pendst();   
+        hprintln!("idle end");
+        
+        debug::exit(debug::EXIT_SUCCESS);
+        loop {
+            cortex_m::asm::wfi();
+        }
+    }
+
+    #[task(binds = SysTick, trampoline = GPIOC, priority = 2, shared = [h_priority])]
+    fn sys_tick(_: sys_tick::Context) {
+        hprintln!("SysTick start");
+    }
+
+    #[task(binds = GPIOA, priority = 1, shared = [h_priority])]
+    fn gpioa(mut ctx: gpioa::Context) {
+        ctx.shared.h_priority.lock(|_| {
+            hprintln!("gpioa lock");
+            cortex_m::peripheral::SCB::set_pendst();   
+            cortex_m::asm::delay(100_000_000);
+            hprintln!("gpioa unlock");
+        });
+    }
+
+    #[task(binds = GPIOB, priority = 3, shared = [h_priority])]
+    fn high_priority_task(_: high_priority_task::Context) {
+        hprintln!("High priority task");
+    }
+}
--- a/rtic-macros/src/codegen/bindings/cortex.rs
+++ b/rtic-macros/src/codegen/bindings/cortex.rs
@@ -319,13 +319,13 @@ pub fn architecture_specific_analysis(app: &App, _: &SyntaxAnalysis) -> parse::R
         .filter(|prio| *prio > 0)
         .collect::<HashSet<_>>();
 
-    let need = priorities.len();
+    let need_software = priorities.len();
     let given = app.args.dispatchers.len();
-    if need > given {
+    if need_software > given {
         let s = {
             format!(
                 "not enough interrupts to dispatch \
-                    all software tasks (need: {need}; given: {given})"
+                    all software tasks (need: {need_software}; given: {given})"
             )
         };
 
--- a/rtic-macros/src/codegen/bindings/cortex.rs
+++ b/rtic-macros/src/codegen/bindings/cortex.rs
@@ -376,7 +376,10 @@ pub fn architecture_specific_analysis(app: &App, _: &SyntaxAnalysis) -> parse::R
     }
 
     // Check that a exception is not used with shared resources
-    // TODO provide warning when a exception is used with source masking since it then ignores locks in priority ceilings
+    // TODO This does not stop priority inversion in source masking if the
+    // exception does not use shared resources. We do currently allow this,
+    // since it does not corrupt data, but it may be worth reconsidering this
+    // decision.
     #[cfg(feature = "cortex-m-source-masking")]
     for (name, task) in &app.hardware_tasks {
         if is_exception(name) && !app.shared_resources.is_empty(){
@@ -390,7 +393,8 @@ pub fn architecture_specific_analysis(app: &App, _: &SyntaxAnalysis) -> parse::R
             } else {
                 return Err(parse::Error::new(
                     name.span(),
-                    "cannot use exceptions with shared resources as hardware tasks when using source masking, consider adding the trampoline attribute",
+                    "cannot use exceptions with shared resources as hardware tasks when using \
+                    source masking, consider adding the trampoline attribute",
                 ));
             }
         }
--- a/rtic-macros/src/codegen/bindings/cortex.rs
+++ b/rtic-macros/src/codegen/bindings/cortex.rs
@@ -102,7 +102,20 @@ mod source_masking {
             } else {
                 None
             }
-        })) {
+        })
+        // Add trampoline tasks if present and not exceptions
+        .chain(app.hardware_tasks.values().filter_map(|task| {
+            if let Some(trampoline) = &task.args.trampoline {
+                if !is_exception(trampoline) {
+                    Some((&task.args.priority, trampoline))
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
+        }))
+    ) {
             let v: &mut Vec<_> = prio_to_masks.entry(priority - 1).or_default();
             v.push(quote!(#device::Interrupt::#name as u32));
             mask_ids.push(quote!(#device::Interrupt::#name as u32));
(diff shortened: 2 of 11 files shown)


examples/lm3s6965/examples/trampoline.rs
```diff
@@ -0,0 +1,65 @@
...
+
+#![no_std]
+#![no_main]
+#![deny(warnings)]
+#![deny(unsafe_code)]
+#![deny(missing_docs)]
+
+use panic_semihosting as _;
+
+// `examples/bouncy_trampoline.rs` testing trampoline feature
```
AfoHT, line 11: Nitpick: Filename and comment are disagreeing :)

rtic-monotonics/CHANGELOG.md
```diff
@@ -5,10 +5,17 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
-## Unreleased
+## [Unreleased]
 
 ### Changed
+
+- changed (sysclk % timer_hz) == 0 to sysclk.is_multiple_of(timer_hz) to make clippy happy as required to CI.
```
AfoHT, line 12: CI always uses latest stable Rust, so new versions bring in new lints/defaults all the time :) 

Thank you for dealing with this, could you split out the 'rtic-monotonics' changes into a separate commit? If you want you could even create a separate PR for this, to help your colleague, that can be fast-tracked into master branch that way :)

Great you added to the CHANGELOG, but please add new entries above, building like a stack. Please cleanup the commentary and the double `### Changed` headers while you're at it :P

AfoHT (changes requested):
As mentioned on Matrix nitpicks so far, generally good stuff :)

FeldrinH:
Is there some rationale somewhere for why this approach was chosen to solve #1088? Out of the alternatives mentioned in the discussion, trampolines seem to me the most cumbersome. They introduce a new public API which has to be documented and maintained indefinitely, they require a free interrupt to be usable, and they add the extra complexity that the actual code no longer runs in the exception handler, which introduces a number of new edge cases for users (e.g. when reasoning about pending interrupts/exceptions).

Spooky-Firefox:
> Is there some rationale somewhere for why this approach was chosen to solve #1088? Out of the alternatives mentioned in the discussion, trampolines seem to me the most cumbersome. They introduce a new public API which has to be documented and maintained indefinitely, they require a free interrupt to be usable, and they add the extra complexity that the actual code no longer runs in the exception handler, which introduces a number of new edge cases for users (e.g. when reasoning about pending interrupts/exceptions).

The trampoline approach was chosen because it avoids using a global critical section, which would block higher-priority tasks. I agree that the solution is not ideal—it introduces an additional public API that must be maintained and documented, and it currently requires a free interrupt (something that could be improved later by reusing existing dispatchers).

However, I don’t fully agree that it adds complexity for the user. With trampolines, all tasks now adhere to the Stack Resource Policy (SRP), so users no longer need to handle exceptions that violate SRP manually. This makes timing and priority behavior easier to reason about overall.

FeldrinH:
> However, I don’t fully agree that it adds complexity for the user. With trampolines, all tasks now adhere to the Stack Resource Policy (SRP), so users no longer need to handle exceptions that violate SRP manually. This makes timing and priority behavior easier to reason about overall.

Perhaps. I suppose it depends on the specific use case. I'm going to trust the judgement of developers more familiar with the RTIC codebase on this.

I wonder what the performance overhead of the trampoline is, but I'm not sure how to best measure it.

FeldrinH:
> I wonder what the performance overhead of the trampoline is, but I'm not sure how to best measure it.

I made some measurements to try and answer this question. My general methodology is that at the start of hardware task I read the current SysTick timer count and then compare it with the value at which SysTick should pend an exeption. SysTick counter clock is configured to match the processor clock.

Basic measurement code:
```rust
#[task(binds = SysTick, trampoline = EXTI0_1, priority = 2)]
fn watchdog(mut ctx: watchdog::Context) {
    let counter = SYST::get_current();
    rprintln!("Counter: {}", 1_000_000 - counter); // 1_000_000 is SysTick reload value
}
```

The results are as follows (tested on STM32C071): Without the trampoline it takes 28 clock cycles from the SysTick counter reaching 0 to the start of the hardware task (including the time taken to read the SysTick counter value). With the trampoline this time increases to 52 clock cycles, which is an extra overhead of 24 clock cycles.

This is suprisingly high. I could not find any clear documentation for what the interrupt latency should be on Cortex-M0+, but some forum posts claimed 16 clock cycles without tail chaining. For the trampoline tail chaining should apply so I would expect the overhead to be less than 16 clock cycles, but it is 24. If somebody knows what is going on here or where the overhead comes from I would appreciate any insight.

korken89:
It's most likely interrupt latency + register stacking that's the 24 clocks.

perlindgren:
Regarding added overhead, the best approach is to look into the generated assembly. 

Right now the exceptions are not masked at on the M0, so the priority inversion is already present in today's implementation. This is not great. For maskable exceptions one could consider generating code that disable the particular exception (SysTick comes to mind). Alternatively, the same effect might be achievable by lowering the priority of the said exception, but I'm not 100% sure fiddling with priority changes at run-time always have defined behavior (so that might be un-sound), or rising the running priority (I have a gut feeling that changing the running priority is UB in some cases.)

The trampoline approach allows (maskable) exceptions access to shared state and should be sound (if we haven't missed anything). It does not fully solve the priority inversion problem though, which is kind of bad.

FeldrinH:
> For maskable exceptions one could consider generating code that disable the particular exception (SysTick comes to mind).

I think that doesn't work correctly because IIRC disabling SysTick exceptions doesn't postpone handling SysTick exceptions, it discards them alltogether. So if a SysTick exception would have happened while a lock is held, instead of getting pended it just gets discarded. That's probably not what we want.

> Alternatively, the same effect might be achievable by lowering the priority of the said exception, but I'm not 100% sure fiddling with priority changes at run-time always have defined behavior (so that might be un-sound)

I tried looking into this twice since the first time you mentioned this and all I found was vague mentions of potential issues in forums. I would hope that masking all interrupts with a critical section while changing the priority would avoid any issues, but I have found nothing in the official documentation to confirm or deny this.

> or rising the running priority.

AFAIK this can't be done for the idle task, because the idle task does not have a configureable priority (at least not on Cortex M0/M0+).

From what I can tell there is only one alternative to trampolines that preserves SRP and that is to dynamically lower priorities of exceptions. Unfortunately it is also the most complicated of the listed alternatives, but I think it can be made to work (whether it can be proven correct I'm not sure; that depends on finding something in the official documentation about the soundness of dynamically changing priorities).

rtic-macros/src/codegen/bindings/cortex.rs
```diff
@@ -350,6 +375,31 @@ pub fn architecture_specific_analysis(app: &App, _: &SyntaxAnalysis) -> parse::R
...
+                    return Err(parse::Error::new(
+                        trampoline.span(),
+                        "cannot use exceptions as trampoline tasks when using source masking",
+                    ));
+                }
+            } else {
+                return Err(parse::Error::new(
+                    name.span(),
+                    "cannot use exceptions with shared resources as hardware tasks when using \
+                    source masking, consider adding the trampoline attribute",
```
FeldrinH, line 397: I tried to get this "cannot use exceptions with shared resources as hardware tasks" error to trigger in my own code, but it does not show up. I'm reasonably sure that a task like this should produce the error, but it doesn't:
```rust
#[task(binds = SysTick, priority = 2, shared = [resource])]
fn systick(mut ctx: systick::Context) {
    ...
```
(Tested on commit d422879f.)

PR #1108, opened 2025-10-06
Add local tasks

usbalbin:
This PR adds the concept of "local tasks". Local tasks are tasks that may only be spawned from other tasks with the same priority, thus running on the same executor. Thanks to this, there is no requirement on the arguments to local tasks to be `Send`/`Sync`. Local tasks will therefore not have the usual `my_task::spawn` method available.

All tasks, local or not, that may spawn other local tasks will have a public field `local_spawner` in their `Context`. This type will then have methods for all local tasks that may be spawned. Tasks are made into local tasks by setting the task attribute `local_task = true` (default false).

```rust
#[app(device = pac, peripherals = false, dispatchers = [SPI1])]
mod app {
    use super::*;

    #[shared]
    struct Shared {}

    #[local]
    struct Local {}

    #[init]
    fn init(_cx: init::Context) -> (Shared, Local) {
        task1::spawn().ok();
        //task2::spawn(Default::default()).ok(); <--- This is rejected since it is a local task
        (Shared {}, Local {})
    }

    #[task(priority = 1)]
    async fn task1(cx: task1::Context) {
        defmt::info!("Hello from task1!");

        // This is the only way to spawn task2, by using the local spawner
        cx.local_spawner.task2(Default::default()).ok();
    }

    //    Note the same prio as above
    //     vvvvvvvv
    #[task(priority = 1, local_task = true)] // <-- This is a local task
    async fn task2(_cx: task2::Context, _nsns: NotSendNotSync) {
        defmt::info!("Hello from task1!");
    }
}

#[derive(Default)]
struct NotSendNotSync(PhantomData<*mut u8>);
```

### DIFF
--- a/rtic-macros/CHANGELOG.md
+++ b/rtic-macros/CHANGELOG.md
@@ -14,6 +14,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 ### Added
 
 - Outer attributes applied to RTIC app module are now forwarded to the generated code.
+- Add attribute `local_task` for tasks that may take args that are !Send/!Sync and can only be spawned from same executor
 
 ## [v2.2.0] - 2025-06-22
 
--- a/rtic-macros/src/codegen/module.rs
+++ b/rtic-macros/src/codegen/module.rs
@@ -1,5 +1,6 @@
 use crate::syntax::{ast::App, Context};
 use crate::{analyze::Analysis, codegen::bindings::interrupt_mod, codegen::util};
+
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
 
@@ -112,37 +113,7 @@ pub fn codegen(ctxt: Context, app: &App, analysis: &Analysis) -> TokenStream2 {
     let internal_context_name = util::internal_task_ident(name, "Context");
     let exec_name = util::internal_task_ident(name, "EXEC");
 
-    items.push(quote!(
-        #(#cfgs)*
-        /// Execution context
-        #[allow(non_snake_case)]
-        #[allow(non_camel_case_types)]
-        pub struct #internal_context_name<'a> {
-            #[doc(hidden)]
-            __rtic_internal_p: ::core::marker::PhantomData<&'a ()>,
-            #(#fields,)*
-        }
-
-        #(#cfgs)*
-        impl<'a> #internal_context_name<'a> {
-            #[inline(always)]
-            #[allow(missing_docs)]
-            pub unsafe fn new(#core) -> Self {
-                #internal_context_name {
-                    __rtic_internal_p: ::core::marker::PhantomData,
-                    #(#values,)*
-                }
-            }
-        }
-    ));
-
-    module_items.push(quote!(
-        #(#cfgs)*
-        #[doc(inline)]
-        pub use super::#internal_context_name as Context;
-    ));
-
-    if let Context::SoftwareTask(..) = ctxt {
+    if let Context::SoftwareTask(t) = ctxt {
         let spawnee = &app.software_tasks[name];
         let priority = spawnee.args.priority;
         let cfgs = &spawnee.cfgs;
@@ -163,13 +134,21 @@ pub fn codegen(ctxt: Context, app: &App, analysis: &Analysis) -> TokenStream2 {
         let (input_args, input_tupled, input_untupled, input_ty) =
             util::regroup_inputs(&spawnee.inputs);
 
+        let local_task = app.software_tasks[t].args.local_task;
+        let unsafety = if local_task {
+            // local tasks are only safe to call from the same executor
+            quote! { unsafe }
+        } else {
+            quote! {}
+        };
+
         // Spawn caller
         items.push(quote!(
             #(#cfgs)*
             /// Spawns the task directly
             #[allow(non_snake_case)]
             #[doc(hidden)]
-            pub fn #internal_spawn_ident(#(#input_args,)*) -> ::core::result::Result<(), #input_ty> {
+            pub #unsafety fn #internal_spawn_ident(#(#input_args,)*) -> ::core::result::Result<(), #input_ty> {
                 // SAFETY: If `try_allocate` succeeds one must call `spawn`, which we do.
                 unsafe {
                     let exec = rtic::export::executor::AsyncTaskExecutor::#from_ptr_n_args(#name, &#exec_name);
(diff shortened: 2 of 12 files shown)


rtic-macros/src/codegen/module.rs
```diff
@@ -204,11 +184,74 @@ pub fn codegen(ctxt: Context, app: &App, analysis: &Analysis) -> TokenStream2 {
...
+                #(#cfgs)*
+                #[doc(inline)]
+                pub use super::#internal_spawn_ident as spawn;
+            ));
+        }
+
+        let local_tasks_on_same_executor: Vec<_> = app
+            .software_tasks
+            .iter()
+            //.filter(|(_, t)| t.args.is_local_task && t.args.priority == priority)
```
usbalbin: Not quite sure why, but this line filters out everything..

usbalbin:
CI seems quite angry, however I am not so sure if it is all my fault...

rtic-macros/ui/spawn-local-different-exec.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    #[local]
+    struct Local {}
+
+    #[init]
+    fn init(_cx: init::Context) -> (Shared, Local) {
+        (Shared {}, Local {})
+    }
+
+    #[task(priority = 1, is_local_task = true)]
```
datdenkikniet: Is there a specific need/reason _not_ to default to all tasks at the same priority to be locally spawnable? If the executor/performance characteristics are expected to be the same, and all of the non-local spawning is unaffected, is there any need for being able to turn off local tasks? 

If one of the (few) reasons is that introducing new functionality like this introduces lots of unknowns, or there is some fixed-size overhead for this, could it make more sense to put this functionality behind a cargo feature in `rtic`/`rtic-macros` instead?

rtic-macros/src/codegen/software_tasks.rs
```diff
@@ -37,8 +37,16 @@ pub fn codegen(app: &App, analysis: &Analysis) -> TokenStream2 {
             let cfgs = &task.cfgs;
             let stmts = &task.stmts;
             let inputs = &task.inputs;
-            let lifetime = if task.is_bottom { quote!('static) } else { quote!('a) };
-            let generics = if task.is_bottom { quote!() } else { quote!(<'a>) };
+            let lifetime = if task.is_bottom {
```
datdenkikniet: Is this an unrelated formatting change, or am I missing a functionality change here?

rtic-macros/ui/spawn-local-different-exec.rs
datdenkikniet: Woops, I see now that I left this as a single comment: I meant to put it up as part of a review.

I think the name change still applies, but the new question (why have it be a configuration field at all) is more important.

rtic-macros/src/codegen/software_tasks.rs
usbalbin: > Is this an unrelated formatting change

I think so, yes

rtic-macros/ui/spawn-local-different-exec.rs
datdenkikniet: Ah OK, I've thought about this a little more and now realize that this isn't an additive change, making the task local-only also _removes_ the global spawning mechanism. That doesn't really fit behind a feature, so I guess the name change (`is_local_task` -> `local_task`) and extra allowed form are actually relevant.

To save you from clicking through the history, this was the original comment:

> To be consistent, I think we should rename this to `local_task`, and accept both of these forms:
> 
> ```rust
> #[task(local_task)]
> #[task(local_task = <bool>)]
> ```

rtic-macros/src/codegen/software_tasks.rs
usbalbin: Would you like me to revert this?

datdenkikniet: Ideally, yes. If you submit it as a separate PR I can force-merge it, while getting rid of this thing bugging you

rtic-macros/ui/spawn-local-different-exec.rs
usbalbin: > [...] this isn't an additive change [...]

Yeah exactly. Marking a task as local_task is quite a fundamental change with large limitations.

usbalbin:
I cherry-picked your commits from #1115 to see if the CI is happy. Let me know if you want me to clean up the history and/or rebase.

usbalbin:
Not sure what is wrong but CI seems to be stuck running forever :/

datdenkikniet:
It seems that the examples in `examples/lm3s6965` expect a `debug::exit` call to indicate that the example has finished when running in QEMU, which the example you added does not have.

I've cancelled the two jobs for now

korken89:
A thought, why is the `is_local_task = true` needed? All tasks which are at the same prio should get the extended API in their context so that they can spawn other tasks at the same prio.
To me `is_local_task = true` is the same as checking if two tasks have the same priority.
To get the effect of supporting `!Send` spawns, we only need an API that enforces that the `spawner` and `spawnee` have the same priority.

Or is there something wrong in my conclusion?

usbalbin:
My understanding is that we can not know in the macro if the type used in the arg to a task is Send/Sync or limit the task to only be spawnable on the same prio. Yet we have to decide in the macro whether to produce the assert Send/Sync code. The `is_local_task`(recently renamed to `local_task`) is to tell the macro that some args are not Send/Sync.

However if there is some clever way to get rid of it then I am happy to do so :)

usbalbin:
[One hack](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=efcc222984d07552fe0f48b29dec34c8) that would avoid the attribute would be to produce something like this:

```rust
mod task1 {
    use super::*;
    pub fn spawn<T0, T1>(x: T0, y: T1)
        where T0: Dummy<T=u8> + Send + Sync, T1: Dummy<T=NotSendSync> + Send + Sync
    {
        let x = x.to();
        let y = y.to();
        // the_real_spawn_here(x, y)
    }
}
```

```rust
pub trait Dummy {
    type T;
    fn to(self) -> Self::T;
}

impl<T> Dummy for T {
    type T = T;
    fn to(self) -> T {
        self
    }
}
```

This defers the errors to the call site:

```
   Compiling playground v0.0.1 (/playground)
error[E0277]: `*mut ()` cannot be sent between threads safely
   --> src/main.rs:2:23
    |
  2 |     task1::spawn(1u8, NotSendSync(core::marker::PhantomData));
    |     ------------      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `*mut ()` cannot be sent between threads safely
    |     |
    |     required by a bound introduced by this call
    |
    = help: within `NotSendSync`, the trait `Send` is not implemented for `*mut ()`
```

Maybe there is a better way?

korken89:
As I see it, the only optimization this does is that you can only generate the local spawn when `is_local_task = true`, and not generate the global one.
My thought is that instead: for all tasks generate local spawn to all other tasks of the same prio AND keep the current global spawn. Then you should not need the `is_local_task` parameter.

They  fill different use-cases. Global spawn for example is so a driver can start something from the depths of a driver given only an function pointer, where local spawns are to support better interoperability with drivers that assume you run at the same priority.

rtic-macros/src/syntax/analyze.rs
```diff
@@ -285,13 +285,16 @@ pub(crate) fn app(app: &App) -> Result<Analysis, syn::Error> {
...
-        // All inputs are send as we do not know from where they may be spawned.
-        spawnee.inputs.iter().for_each(|input| {
-            send_types.insert(input.ty.clone());
-        });
+        if !spawnee.args.local_task {
+            // All inputs are send as we do not know from where they may be spawned.
+            spawnee.inputs.iter().for_each(|input| {
+                send_types.insert(input.ty.clone());
+            });
+        }
```
usbalbin, line 297: @korken89 
> As I see it, the only optimization this does is that you can only generate the local spawn when is_local_task = true, and not generate the global one.
My thought is that instead: for all tasks generate local spawn to all other tasks of the same prio AND keep the current global spawn. Then you should not need the is_local_task parameter.
>
> They fill different use-cases. Global spawn for example is so a driver can start something from the depths of a driver given only an function pointer, where local spawns are to support better interoperability with drivers that assume you run at the same priority.

As is now, we need the attribute to know if it is safe to skip the check for Send/Sync. The global spawn would be unsafe to call with !Send/!Sync args. But I get the feeling that I do not quite understand :)

usbalbin: As for always populating the local spawner with tasks of same prio, sure absolutely! Should the task also have itself in its local spawner?

korken89: The local spawn will have `!Send`, while the global spawn will always require `Send` - as global spawn can never know which one is calling it.
So the only ones that can alleviate the `Send` requirements will be the local spawns.

usbalbin: > [...] the global spawn will always require Send [...]

Yes, and right now this is enforced by [this assert](https://github.com/rtic-rs/rtic/blob/master/rtic-macros/src/codegen/assertions.rs) which is emitted for all `send_types` from above, right?

This PR removes the assert and the global spawn only for the tasks marked with the attribute. If we remove the attribute then how will the macro know when it should skip generating the assert and global spawn?

We could emit where clauses on the global spawn function which checks that all the arguments impl Send + Sync like this:

```rust
fn spawn(x: u8, y: NotSendSync)
    where u8: Send + Sync, NotSendSync: Send + Sync { ... }
```

However that fails to compile with !Send/Sync even if not calling the types since we are using concrete types.

When creating dummy trait as done in my comment. This defers the compile error to when the function is called. This would, I think, do the same thing as the assert but only at the call site. This way we can replace the asserts with where clauses and then just always emit all the global spawn functions since it will be impossible to call them with !Send/!Sync types.

Or do you have something different in mind?

rtic-macros/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 ### Added
 
 - Outer attributes applied to RTIC app module are now forwarded to the generated code.
+- Add attribute `local_task` for tasks that may take args that are !Send/!Sync and can only be spawned from same executor
```
AfoHT: Please put new entries at the top, like a stack.

usbalbin:
Is it ok if I rebase this over the latest changes on master?

AfoHT:
Please go ahead!

korken89 (approved):

PR #1111, opened 2025-10-12
Change stm32 TIM5 from u16 to u32

boondocklabs:
I noticed that the first overflow ISR was corrupting the time on a G474 when using TIM5 as a monotonic.

All stm32 with TIM5 are 32bit timers to the best of my knowledge. This changes the type from u16 to u32 which has resolved the issue.

### DIFF
--- a/rtic-monotonics/src/stm32.rs
+++ b/rtic-monotonics/src/stm32.rs
@@ -365,7 +365,7 @@ make_timer!(Tim3Backend, TIM3, u16, TIMER3_OVERFLOWS, TIMER3_TQ);
 make_timer!(Tim4Backend, TIM4, u16, TIMER4_OVERFLOWS, TIMER4_TQ);
 
 #[cfg(feature = "stm32_tim5")]
-make_timer!(Tim5Backend, TIM5, u16, TIMER5_OVERFLOWS, TIMER5_TQ);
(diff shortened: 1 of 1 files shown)


burrbull:
> All stm32 with TIM5 are 32bit timers

All except F1, L0, L1 series and F410.

boondocklabs:
> > All stm32 with TIM5 are 32bit timers
> 
> All except F1, L0, L1 series and F410.

Well that complicates things. Any clean way to do this without listing each device in #[cfg] directives? It would be nice if stm32-metapac had associated types on the timers or from the metadata module.

Finomnis:
Are you 100% sure they are 32 bits? The datasheet says stm32g474 only has 16 bit timers ... I suspect it's a bug on our side, based on stm32-metapac. The metapac returns the timer value as `u32`, which the monontonic then interprets as being a 32-bit counter.

boondocklabs:
Yes, TIM2 and TIM5 on G474 are 32 bit

Finomnis:
Ok yes I see. Sorry my mistake.

Finomnis:
@boondocklabs Would you mind trying #1171 again? It now derives the counter size from the PAC definitions.

Finomnis:
Can be closed as a duplicate of #1171.

Issue #1107, opened 2025-10-06
Add spawn_local


usbalbin:
(From the matrix discussion, I hope it's ok if a create an issue for this)

Add some way of spawning tasks on the same executor without requiring the args to be `Send` and `Sync`.

PR #1116, opened 2025-10-20
Spawn local without attribute

usbalbin:
This is an alternative to #1108 for solving #1107 that does not require the `local_task` attribute.

### Breaking change

Note this is a breaking change.

Inference will no longer accept

```rust
task1::spawn(Default::default()).unwrap();
```

So that would have to be changed to 

```rust
task1::spawn(MyArgType::default()).unwrap();

// or

ctx.local_spawner.task1(Default::default()).unwrap();
```

Similarly with `&mut T` for task argument no longer auto converted into `&T`.

```rust
let x: &mut u8 = ctx.local.x;
task1_taking_shared_ref::spawn(x).unwrap();
```
instead do

```rust
let x: &mut u8 = ctx.local.x;
task1_taking_shared_ref::spawn(&*x).unwrap();

// or 

ctx.local_spawner.task1_taking_shared_ref(x).unwrap();
```


usbalbin:
Well this breaks a lot of inference stuff, so probably not a great solution...

examples/lm3s6965/examples/wait-queue.rs
```diff
@@ -28,8 +28,8 @@ mod app {
     fn init(cx: init::Context) -> (Shared, Local) {
         Mono::start(cx.core.SYST, 12_000_000);
 
-        incrementer::spawn(cx.local.wait_queue).ok().unwrap();
-        waiter::spawn(cx.local.wait_queue).ok().unwrap();
+        incrementer::spawn(&*cx.local.wait_queue).ok().unwrap();
+        waiter::spawn(&*cx.local.wait_queue).ok().unwrap();
```
usbalbin, line 32: This is a breaking change. For some reason the compiler can not auto demote `&mut T` to `&T` when generics are involved (not sure of the rules here). 

Without the above fix, we get

```
error[E0271]: type mismatch resolving `<&mut DoublyLinkedList<Waker> as Dummy>::T == &DoublyLinkedList<Waker>`
  --> examples/wait-queue.rs:31:28
   |
31 |         incrementer::spawn(cx.local.wait_queue).ok().unwrap();
   |         ------------------ ^^^^^^^^^^^^^^^^^^^ type mismatch resolving `<&mut DoublyLinkedList<Waker> as Dummy>::T == &DoublyLinkedList<Waker>`
   |         |
   |         required by a bound introduced by this call
   |
note: expected this to be `&'static DoublyLinkedList<Waker>`
  --> examples/wait-queue.rs:11:1
```

rtic/src/export/dummy.rs
```diff
@@ -0,0 +1,15 @@
+// TODO: What should we name this?
+
+/// Dummy trait which will only ever be implemented where type T is Self
+pub trait Dummy {
```
usbalbin: Any suggestions for what to call this?

usbalbin: This is used because we can not have a function with a where clause which requires a concrete type to be Send if that type is not Send. The compiler will error out on us.
However hiding that behind a dummy trait which is only implemented for that same type then we defer the error to when the user tries to call the function

usbalbin: I went with `TaskArg` for now

examples/lm3s6965/examples/wait-queue.rs
usbalbin: Copying AfoHT's link on matrix: https://doc.rust-lang.org/reference/type-coercions.html#r-coerce.types.mut-reborrow

usbalbin: From the [urlo thread](https://users.rust-lang.org/t/auto-reborrow-not-working-for-generic-arguments/136707?u=usbalbin) it seems that the issue is a variant of this https://github.com/rust-lang/rust/issues/85161

PR #1120, opened 2025-10-28
chore: replace deprecated `bare-metal` with `critical-section`

Ddystopia:
`bare_metal` was last time updated 3 years ago with a [deprecation].

image

Yet, rtic still uses `bare_metal` and gives `ctx.cs` as `bare_metal::CriticalSection`. This is a problem because most libraries use `critical_section::CriticalSection`. And for some reason no one published `bare_metal 1.1.0`, thus on crates.io the critical section of the `bare_metal` is a different type from that of `critical_section`.

[deprecation]: https://github.com/rust-embedded/bare-metal/blob/master/src/lib.rs


AfoHT (approved):
Thank you!

PR #1121, opened 2025-10-28
Fix mutex send

Ddystopia:
Each mutex is generated uniquely for each task, it is unsound to send them between tasks. But they are `Send`. Before, it wasn't an issue, because you couldn't share non-`'static` data between them, but with #1043 you can make the mutex `'static`. Thus we need to use actual tools that Rust provides and out out from `Send`.

Currently, mutexes are simple ZSTs with `PhantomData<&'a ()>`, which is `Send`. We replace it with `PhantomData<(&'a (), *const u8)>`, and return `Sync` back via `unsafe` implementation. It is trivially sound, because mutexes have
no method methods that accept `&self`. See https://doc.rust-lang.org/std/sync/struct.Exclusive.html for details.

### DIFF
--- a/rtic-macros/src/codegen/shared_resources.rs
+++ b/rtic-macros/src/codegen/shared_resources.rs
@@ -62,8 +62,13 @@ pub fn codegen(app: &App, analysis: &Analysis) -> TokenStream2 {
                 #[allow(non_camel_case_types)]
                 #(#cfgs)*
                 pub struct #shared_name<'a> {
-                    __rtic_internal_p: ::core::marker::PhantomData<&'a ()>,
+                    __rtic_internal_p: ::core::marker::PhantomData<(&'a (), *const u8)>,
                 }
+
+                // Opt out from `Send`.
+                // `#shared_name` is trivially `Sync` because there are no `&self` methods.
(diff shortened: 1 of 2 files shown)


rtic-macros/src/codegen/shared_resources.rs
```diff
@@ -62,9 +62,14 @@ pub fn codegen(app: &App, analysis: &Analysis) -> TokenStream2 {
...
                 #(#cfgs)*
                 pub struct #shared_name<'a> {
-                    __rtic_internal_p: ::core::marker::PhantomData<&'a ()>,
+                    __rtic_internal_p: ::core::marker::PhantomData<(&'a (), *const u8)>,
                 }
 
+                // Opt out from `Send`.
+                // `#shared_name` is trivially `Sync` because there are no `&self` methods.
+                // See https://doc.rust-lang.org/std/sync/struct.Exclusive.html .
+                unsafe impl<'a> Sync for #shared_name<'a> {}
```
korken89, line 72:
```suggestion
                #(#cfgs)*
                unsafe impl<'a> Sync for #shared_name<'a> {}
```

korken89 (approved):

PR #1124, opened 2025-11-08
Support external SYST clock source

mciantyre:
Give users the option to configure SYST with an external clock source. By default, the clock source is the core, which keeps us backwards compatible.

`_start` has a new input, but since users don't call that function directly, it doesn't seem like a problem to break that public API.


korken89 (approved):
Thanks!

PR #1125, opened 2025-11-18
fix the const assertion for the queue size

robamu:
The const check was not evaluated, see https://doc.rust-lang.org/reference/items/associated-items.html#r-items.associated.const.eval

### DIFF
--- a/book/en/book.toml
+++ b/book/en/book.toml
@@ -6,7 +6,6 @@ authors = [
   "Jorge Aparicio <jorge@japaric.io>",
   "Per Lindgren <per.lindgren@ltu.se>",
 ]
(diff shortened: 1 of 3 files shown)


AfoHT (approved):
Thank you! Hope you don't mind me hijacking this PR a bit to fix CI (to get this merged) :D

korken89:
In my experiments this sometimes puts the `assert!` in runtime code if not used as a `const new`. You need to place the calls in const blocks to get rid of it:

```rust 
const { Self::size_check() };
```

robamu:
See https://github.com/rtic-rs/rtic/pull/1127

PR #1134, opened 2025-12-17
Remove `portable-atomic` from rtic-monotonics

### LINKED ISSUE #1117: `systick-64bit` on armv7 can deadlock
It appears that using a 64-bit backed rtic-monotonics counter, on armv7 portable-atomics will polyfill with a AtomicU64 implementation that can deadlock.

To be fair this is [expected behavior](https://docs.rs/portable-atomic/latest/portable_atomic/struct.AtomicU64.html#method.is_lock_free) of portable-atomics, it's just that because the polyfill is automatic it's easy to not realise that you've introduced a deadlock.

Could this at least be outlined in the docs? Or even better, prevent armv7 targets from using 64bit systick at all? In either case I am happy to open the PR if someone is able to point me in the right direction.

Here's a stack trace from when using `systick-64bit` locks up:
```
#0  0x08017168 in core::iter::range::{impl#5}::spec_next<i32> (self=<optimized out>) at /Users/liam/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/iter/range.rs:764
#1  core::iter::range::{impl#6}::next<i32> (self=<optimized out>) at /Users/liam/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/iter/range.rs:849
#2  portable_atomic::imp::fallback::utils::Backoff::snooze (self=<optimized out>) at /Users/liam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/portable-atomic-1.11.1/src/imp/fallback/utils.rs:138
#3  portable_atomic::imp::fallback::seq_lock::SeqLock::write (self=0x20007480 <portable_atomic::imp::fallback::lock::LOCKS+768>)
    at /Users/liam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/portable-atomic-1.11.1/src/imp/fallback/seq_lock_wide.rs:97
#4  0x080170c4 in portable_atomic::imp::fallback::AtomicU64::fetch_add (self=0x200079e0 <rtic_monotonics::systick::SYSTICK_CNT>, val=1, _order=core::sync::atomic::Ordering::AcqRel)
    at /Users/liam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/portable-atomic-1.11.1/src/imp/fallback/mod.rs:322
#5  portable_atomic::AtomicU64::fetch_add (self=0x200079e0 <rtic_monotonics::systick::SYSTICK_CNT>, val=1, order=core::sync::atomic::Ordering::AcqRel)
    at /Users/liam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/portable-atomic-1.11.1/src/lib.rs:3241
#6  0x0800c720 in rtic_monotonics::systick::{impl#1}::on_interrupt () at /Users/liam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rtic-monotonics-2.1.0/src/systick.rs:130
#7  rtic_time::timer_queue::TimerQueue<rtic_monotonics::systick::SystickBackend>::on_monotonic_interrupt<rtic_monotonics::systick::SystickBackend> (self=<optimized out>)
    at /Users/liam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rtic-time-2.0.1/src/timer_queue.rs:102
#8  pdm_36::{impl#0}::start::SysTick () at /Users/liam/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rtic-monotonics-2.1.0/src/systick.rs:181
```

korken89:
Closes #1117

### DIFF
--- a/rtic-monotonics/CHANGELOG.md
+++ b/rtic-monotonics/CHANGELOG.md
@@ -8,7 +8,9 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 ## Unreleased
 
 ### Changed
+
 - Panic if STM32 prescaler value would overflow
+- Moved away from `portable-atomic` due to it's spinlock design
 
 ### Added
 
--- a/rtic-monotonics/Cargo.toml
+++ b/rtic-monotonics/Cargo.toml
@@ -39,10 +39,9 @@ rustdoc-flags = ["--cfg", "docsrs"]
 [dependencies]
 rtic-time = { version = "2.0.0", path = "../rtic-time" }
 fugit = { version = "0.3.6" }
-portable-atomic = { version = "1" }
 cfg-if = "1.0.0"
 cortex-m = { version = "0.7.6", optional = true }
-critical-section = { version = "1", optional = true }
+critical-section = { version = "1" }
 
 # RP2040
 rp2040-pac = { version = "0.6", optional = true }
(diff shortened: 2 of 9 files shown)


FeldrinH:
It would be nice to keep portable-atomics around. Outside of this one deadlock issue, portable-atomics generally has ways to emulate atomics that are more efficient than a critical section around every operation, especially on multicore systems.

korken89:
If they get our base case fixed, then we can keep it. I'm not that positive to keep it when especially if it can cause dead locks.

Could you @FeldrinH champion the fixing of that create?

FeldrinH:
I can try to if I can reproduce the deadlock on hardware I own. Unfortunately I do not own any armv7m devices.

PR #1142, opened 2026-01-20
Add watch sync primitive

syltkaka:
`watch` does not evict latest value as `signal` do. Once initialized the watch reader always return a value.

### DIFF
--- a/rtic-sync/src/watch.rs
+++ b/rtic-sync/src/watch.rs
@@ -1,6 +1,8 @@
 //! A "latest only" value store with unlimited writers and async waiting. Value is always available once initialized.
 
 use crate::signal::{Signal, SignalWriter, Store};
+use core::{future::poll_fn, task::Poll};
+use portable_atomic::Ordering::{Acquire, Release};
 
 /// A "latest only" value store with unlimited writers and async waiting. Value is always available once initialized.
 pub struct Watch<T: Copy>(Signal<T>);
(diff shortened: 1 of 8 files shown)


korken89:
Thanks for the PR! Can you add an example here so we run it through QEMU and check intended behavior? https://github.com/rtic-rs/rtic/tree/master/examples/lm3s6965/examples
You can use the `async-channel` examples are inspiration.

rtic-sync/src/watch.rs
```diff
@@ -0,0 +1,190 @@
...
+
+/// Creates a split watch with `'static` lifetime.
+#[macro_export]
+macro_rules! make_watch {
+    ( $T:ty ) => {{
+        static WATCH: $crate::watch::Watch<$T> = $crate::watch::Watch::new();
+
+        WATCH.split()
+    }};
+}
```
glaeqen, line 57: We were lazy to fix unsoundness of `Signal` but we should have an atomic bool in the Watch (and Signal) that checks if we already split the Watch

glaeqen: Something like this https://docs.rs/bbqueue/0.5.0/src/bbqueue/bbbuffer.rs.html#90

glaeqen: Maybe not necessarily a soundness issue, but it is not correct, for example with two SignalReaders wakers get overwritten etc.

syltkaka: I see. There seems to be a slightly different approach used with `channel`, it panics rather than returns an error. Should the handling be aligned across `channel`, `signal`, and `watch`? And secondly, if yes to previous question, how would you like to introduce it?

glaeqen: If it was me I'd do it in the same PR because I'm lazy and GitHub sucks. However, it is not critical-critical to have it together, so consider making a separate change.

rtic-sync/src/watch.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+    }
+
+    /// Returns the latest value once Watch is initialized.
+    pub async fn get(&mut self) -> T {
+        poll_fn(|ctx| {
+            self.parent.waker.register(ctx.waker());
+            match self.get_inner() {
+                Store::Unset => Poll::Pending,
+                Store::Set(value) => Poll::Ready(value),
```
korken89: I get the feeling this in unintentional, this will return directly if there is data in the watch - but not wait for *new* data. This is odd for an `async` method, where you expect it to wait for something (even though the documentation says it will return directly after initialization).

We add an API similar to `tokio` if we want, or something of our own but with similar semantics:

- https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.changed
- https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.has_changed

Thoughts? So there is a way for the user to wait for new data.

glaeqen: We talked about this with Nils one time and my impression was that we roll with something simpler than embassy_sync::Watch for now but I don't have an opinion. It'd be good not drag this review for too long as well though.

korken89: We can roll with something simpler as long as it will not entail a major release of `rtic-sync` to fix it.

korken89 (approved):
Looks good, thanks for fixing!

PR #1144, opened 2026-02-02
fix unhygienic panic use in macros and codegen

nilfit:
In some places in codegen, `::core::panic` was used, but in others places plain `panic` was used. This meant that a macro caller could shadow the definition for some panics (for example with `defmt::panic`), but not others. Fix this inconsistency by using `::core::panic` everywhere.

This affects not only the `rtic::app` proc macro, but also `macro_rules` [1]. I searched for and changed the following in both:

- assert_eq
- assert_ne
- assert
- debug_assert_eq
- debug_assert_ne
- debug_assert
- todo
- unimplemented
- unreachable
- panic

[1]: https://doc.rust-lang.org/reference/macros-by-example.html#hygiene

### DIFF
--- a/rtic-macros/CHANGELOG.md
+++ b/rtic-macros/CHANGELOG.md
@@ -9,6 +9,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ### Fixed
 
+- Generated `panic`s can no longer be shadowed. Previously, they were inconsistently shadowed.
 - Generated mutexes for shared resourses no longer implement `Send`. They are meant to be used inside a task locally.
 
 ### Added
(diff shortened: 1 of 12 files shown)


rtic-macros/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 ### Fixed
 
 - Generated mutexes for shared resourses no longer implement `Send`. They are meant to be used inside a task locally.
+- Generated `panic`s can no longer be shadowed. Previously, they were inconsistently shadowed.
```
AfoHT, line 12: Please put the entry above existing entries 

Same "nit" for all Changelog files

AfoHT (changes requested):
Awesome!

rtic-macros/CHANGELOG.md
nilfit: Changed

AfoHT (approved):
Lovely!

PR #1156, opened 2026-03-20
bump lm3s6965 crate version on release/v1 branch

elmjag:
Require version 0.2.0 of the `lm3s6965` crate, when building examples.

The older version does not compile anymore, it fails due to missing `Copy` trait for `lm3s6965::Interrupt` type.


elmjag:
I checked failing CI tests. They are failing on release/v1 branch already, without my change.

korken89 (approved):

PR #1166, opened 2026-04-22
Clippy: Fix clippy::question-mark

AfoHT:
- ****
- **rtic-macros: Clippy: Fix clippy::question-mark**


perlindgren (approved):
Clippy fixes only.

PR #1169, opened 2026-05-04
examples: fix typo in sleeponexit

Finomnis:
The documentation said `SLEEPONEXIT` (which seems correct), but the code contained `set_sleepdeep`.

Also, the doc links were broken.


AfoHT (approved):
Thanks!

PR #1171, opened 2026-05-05
monotonics: stm32: derive counter size from pac

### LINKED ISSUE #1175: stm32_tim5_monotonic! doesn't work on g474
Depending on the value of `tick_rate_hz` and `timer_clock_hz` awaiting `Mono::delay` _will work some number of times_. But eventually the delay stops working and it delays forever.

I believe the issue is that the code assumes timer 5 is 16 bit but on g4 it's actually 32 bit.

Image

If I use `stm32_tim4_monotonic` it works.

Finomnis:
Fixes #1111.

`.0` reads the full 32-bit register as `u32`, while `.cnt()` reads only the 16-bit counter value as `u16`.

Further, this removes all hard-coded instances of the counter sizes, deriving them from the stm32 pac.

### DIFF
--- a/rtic-monotonics/CHANGELOG.md
+++ b/rtic-monotonics/CHANGELOG.md
@@ -20,6 +20,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 ### Fixed
 
 - `assert`s in macro expansions can no longer be shadowed.
+- Incorrect bit width of stm32's timer2 based peripherals
 
 ## v2.1.0 - 2025-06-22
 
--- a/rtic-monotonics/CHANGELOG.md
+++ b/rtic-monotonics/CHANGELOG.md
@@ -9,6 +9,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ### Changed
 
+- STM32's monotonics now derive the timer counter size from the pac register definition
 - Updated rp235x-pac dependency to v0.2.0
 - Panic if STM32 prescaler value would overflow
 
@@ -20,7 +21,6 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 ### Fixed
 
 - `assert`s in macro expansions can no longer be shadowed.
-- Incorrect bit width of stm32's timer2 based peripherals
 
 ## v2.1.0 - 2025-06-22
 
--- a/rtic-monotonics/src/stm32.rs
+++ b/rtic-monotonics/src/stm32.rs
@@ -287,7 +287,7 @@ macro_rules! make_timer {
             type Ticks = u64;
 
             fn now() -> Self::Ticks {
-                calculate_now(
+                calculate_now::<_, $bits, _, _, _>(
                     || $overflow.load(Ordering::Relaxed),
                     || $timer.cnt().read()
                 )
@@ -431,9 +431,9 @@ macro_rules! make_timer2 {
             type Ticks = u64;
 
             fn now() -> Self::Ticks {
-                calculate_now(
+                calculate_now::<_, $bits, _, _, _>(
                     || $overflow.load(Ordering::Relaxed),
-                    || $timer.cnt().read().0
+                    || $timer.cnt().read().cnt()
                 )
             }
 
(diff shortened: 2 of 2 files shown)


rtic-monotonics/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 - Updated rp235x-pac dependency to v0.2.0
 - Panic if STM32 prescaler value would overflow
+- STM32's monotonics now derive the timer counter size from the pac register definition
```
AfoHT: I think this might need rebasing as we brought in other things into the monotonics changelog

When you do (or if I do), put the changelog entry "at the top" of the stack, ie before "updated rp235x..."

Finomnis: Should not prevent target tests though

Finomnis: @AfoHT Rebased, but I think it was already up to date

rtic-monotonics/src/stm32.rs
```diff
@@ -211,8 +211,61 @@ macro_rules! stm32_tim15_monotonic {
...
+    fn trunc_from_u64(val: u64) -> Self {
+        val as Self
+    }
+}
+
+fn half_period_value<T: TimerWord>(
+    _: impl FnOnce() -> T, /* as a sanity check that the CCR width is identical to the CNT width */
+) -> T {
+    T::HALF_PERIOD_VALUE
+}
```
andresv, line 259: I propose a doc change here:

```rust
/// Returns the half-period value for a timer.
///
/// The closure argument is never invoked; it forces the CNT and CCR word
/// types to be inferred as the same `T`, causing a width mismatch between
/// the two registers to fail type-checking.
///
/// # Example
///
/// ```ignore
/// timer.ccr(0).write(|r| *r = half_period_value(|| timer.cnt().read()));
/// ```
fn half_period_value<T: TimerWord>(_: impl FnOnce() -> T) -> T {
    T::HALF_PERIOD_VALUE
}
```

rtic-monotonics/src/stm32.rs
```diff
@@ -211,8 +211,61 @@ macro_rules! stm32_tim15_monotonic {
...
+    let val = if T::try_from(instant.wrapping_sub(now)).is_ok() {
+        instant
+    } else {
+        // In the past or will overflow
+        0
+    };
+
+    T::trunc_from_u64(val)
+}
+
```
andresv, line 295: I propose a doc change here:

```rust
/// Computes the CCR compare value for a target `instant`, given the current
/// counter value `now`.
///
/// If the target is in the past or would overflow the timer's range before
/// being reached, returns `0` so the compare match fires on the next
/// wrap-around rather than at a stale value.
///
/// The closure argument is never invoked; it forces the CNT and CCR word
/// types to be inferred as the same `T`, causing a width mismatch between
/// the two registers to fail type-checking.
///
/// # Example
///
/// ```ignore
/// timer.ccr(1).write(|r| {
///     *r = compute_compare_value(instant, now, || timer.cnt().read());
/// });
/// ```
fn compute_compare_value<T: TimerWord>(
    instant: u64,
    now: u64,
    _: impl FnOnce() -> T,
) -> T {
    // Since the timer may or may not overflow based on the requested compare val, we check how many ticks are left.
    // `wrapping_sub` takes care of the u64 integer overflow special case.
    let val = if T::try_from(instant.wrapping_sub(now)).is_ok() {
        instant
    } else {
        // In the past or will overflow
        0
    };
    T::trunc_from_u64(val)
}
```

Finomnis: Awesome, thanks :)

Added.

rtic-monotonics/src/stm32.rs
Finomnis: Added :)

pietgeursen:
I just tested this branch on stm32g474.
| Timer | Version | Works? |
|--------|--------|--------|
| TIM2 | v2.1.0 | ✅  |
| TIM2 | PR | ✅  | 
| TIM5 | v2.1.0 | 🔴  |
| TIM5 | PR | ✅  |

AfoHT (approved):
Great job!

AfoHT:
@korken89 Could you check this too? Just to make sure :)

We'll merge this now to increase exposure, but before cutting any releases I mean.

PR #1182, opened 2026-05-30
Release rtic sync v1.5.0

datdenkikniet:
Some handy fixes, and the very nice `Watch` channel have been laying around for a while. It felt reasonable to do a release.

Also added "yanked" notices to all versions of `rtic-sync` that are actually yanked. I don't believe 1.4.0 needs yanking: it has some correctness (i.e. "can we use the crate correctly") bugs, but no soundness issues.


AfoHT (approved):
Nice! Yes this is a great call

PR #1183, opened 2026-05-30
Upgrade rtic-sync version of examples to non-yanked versions

datdenkikniet:
Some of our examples have dependencies on yanked versions of `rtic-sync`. That's probably bad, so I've updated them everywhere. Have also ran `cargo update -p rtic-sync` for all examples to ensure they're at a good version.


AfoHT (approved):
Thanks!

PR #1184, opened 2026-05-31
monotonics: implement RTCC and LETimer monotonic for the SiLabs EFR32

Bnyro:
Hey, this PR adds two monotonic implementations for the EFR32.

I've based this on https://github.com/rtic-rs/rtic/blob/master/rtic-monotonics/src/rp2040.rs and tested using a small blinky demo app on the EFR32xG22E dev kit.

This adds two monotonic impls:
- rtcc: uses the RTCC peripheral (32-bit), only supported on the EFR32MG22
- letimer: uses the LETimer peripheral (24-bit)

Cheers

PS: Here's the reference manual: https://www.silabs.com/documents/public/reference-manuals/efr32xg22-rm.pdf


rtic-monotonics/src/silabs/rtcc.rs
```diff
@@ -0,0 +1,160 @@
...
+            .cc0_ocvalue()
+            .write(|w| unsafe { w.oc().bits(instant) });
+    }
+
+    fn clear_compare_flag() {
+        // unfortunately the RTCC_IF_CLR flag is not part of the SVD/PAC
+        let cc0_clear_reg = (RtccNs::ptr() as u32 + 0x2014) as *mut u32;
+        unsafe {
+            cc0_clear_reg.write_volatile(0b10_000 /* CC0 clear is at bit 4 */)
+        };
```
Bnyro: Unfortunately this is something I've also had to fight with in the past few months: The SVDs/PACs don't contain any `SET` or `CLR` registers for interrupt flags, so we have to manually write to the required address.

andresv:
What a time to be alive. I did something similar for embassy just 2 weeks ago or so: https://github.com/embassy-rs/embassy/pull/6080.

Look `PAC generation pipeline` in this PR. I added EFR32/EFM32 PAC generation from scratch and fixed exactly this kind of clear flag problem in my PAC version.

# IF Clear
https://github.com/andresv/silabs-data-generated/blob/main/silabs-metapac/src/peripherals/timer_v1.rs#L367-L371
https://github.com/andresv/silabs-data-generated/blob/main/silabs-metapac/src/peripherals/burtc_v1.rs#L147-L151

rtic-monotonics/src/efr32.rs
```diff
@@ -0,0 +1,160 @@
...
+//!          let timestamp = Mono::now();
+//!          // ...and you can use it to add a delay to this async function
+//!          Mono::delay(100.millis()).await;
+//!     }
+//! }
+//! ```
+
+/// Common definitions and traits for using the efr32 rtc monotonic
+pub mod prelude {
+    pub use crate::efr32;
```
andresv: Better to name it `silabs` instead of `efr32`. Silicon Labs has `efr32`, `efm32` and there was also some other series that I do not remember anymore. I think it is better to make it more generic to support more MCUs in the future.

Bnyro: You're right, thanks!

EFM32 should definitely work as well (as the EFR32 are basically the same but with an additional radio interface). I don't know if there are any other boards either, but EFR32 and EFM32 are already enough of a reason to name it `silabs`.

Bnyro:
> What a time to be alive. I did something similar for embassy just 2 weeks ago or so: [embassy-rs/embassy#6080](https://github.com/embassy-rs/embassy/pull/6080).
> 
> Look `PAC generation pipeline` in this PR. I added EFR32/EFM32 PAC generation from scratch and fixed exactly this kind of clear flag problem in my PAC version.
> # IF Clear
>
Thank you, very cool, I'll look into that!

The metapac looks very interesting because that would also allow us to support all EFRs/EFMs by depending on just one single crate (so that you only have to enable the correct EFR/EFM feature and then can use it on all silabs boards).

Unfortunately, it seems like the metapac is not on `crates.io` yet though?  I'm not sure if its okay for the maintainer if we add a dependency on a crate that is only available via Git?

Bnyro:
I've pushed https://github.com/rtic-rs/rtic/pull/1184/changes/75bc40ca246b921a0e5cae5de8431feab778e77b which switches to using your `silabs-metapac` (https://github.com/andresv/silabs-data-generated/) and added support for the EFR32MG24, EFR32MG26 and EFR32FG25 (only tested if it compiles - I don't have these boards. The only MCU I tested this on is the EFR32MG22).

andresv:
Just published https://crates.io/crates/silabs-metapac

Bnyro:
> Just published https://crates.io/crates/silabs-metapac
>
Thanks! Switched to using it the crates.io variant now.

Finomnis:
Do the timers have two compare interrupts and an overflow interrupt? If so, please consider widening them to `u64` using [`rtic_time::half_period_counter`](https://docs.rs/rtic-time/latest/rtic_time/half_period_counter/).

Note that the rp2040 monotonic is not a good example to base your code off, as it is already native 64bit and does not require widening.

Finomnis:
The main point of a monotonic is to be, well, monotonic :) That means overflows after minutes/hours/days are not allowed, and they must be at least u64.

I'm the author of most of the other monotonics and the half-period-counter mechanism, if you have questions, feel free to contact me on the matrix chat :)

andresv:
I am preparing a PR on top of @Bnyro work to fix this problem. Should be ready quite soon. I will let you know.

andresv:
PR https://github.com/rtic-rs/rtic/pull/1186 fixes issue that @Finomnis mentioned.

Finomnis:
Damn, that was fast

Bnyro:
> I am preparing a PR on top of @Bnyro work to fix this problem. Should be ready quite soon. I will let you know.
>
Hey, thanks for looking into this.

However, I would have really loved to work on this myself to understand [rtic_time::half_period_counter](https://docs.rs/rtic-time/latest/rtic_time/half_period_counter/) and get a bit more familiar with the EFR.

I also made all my changes without using AI help, so I'm not really excited that you continued vibe-coding the rest of the changes, which in my eyes usually always reduces the code quality. We can also see that here when looking at the code-to-comments ratio, no human would have written so exhaustive comments and commit messages (e.g. https://github.com/rtic-rs/rtic/pull/1186/changes/fa386bf5aa7e330d3511d1433a32ce768a289a92) that are actually not making the code better in any way.

Finomnis:
Fair point. If you want it as an exercise that's totally fine, I suggest @andresv to close the pr and review your code instead :)

Finomnis:
@andresv you can keep your change in your fork and once @Bnyro is done, we can compare the two to get the best of both worlds.

andresv:
@Bnyro fair point, I indeed used claude to do it. We could ofcourse go with your PR. My proof of concept PR can be used as reference if something from there is useful.

In the mean time I polish it a bit. In the end goal is to get silabs monotonics that are usable and maintainable.

Finomnis:
@Bnyro in case you don't quickly find it yourself: the asserts in the interrupt handlers should check for even and odd, both of yours check for odd. I'm sure you'll find it yourself, just wanted to save you time :)

Bnyro:
> @Bnyro in case you don't quickly find it yourself: the asserts in the interrupt handlers should check for even and odd, both of yours check for odd. I'm sure you'll find it yourself, just wanted to save you time :)
>
Good catch, thank you! I've only tested the wrap around the full period (FFFFFFFF -> 0) so far, so I hadn't noticed this copy-paste whoopsie yet :)

Finomnis:
FYI you can use a 32 bit atomic for the overflow counter if your base counter is already 32 bit. This is a total of 63 bit which is big enough.

For the 24 bit counter we would have to do the math how long it would take to overflow with a 32 bit counter.

Bnyro:
> FYI you can use a 32 bit atomic for the overflow counter if your base counter is already 32 bit. This is a total of 63 bit which is big enough.
>
Good to know, thanks! I was already wondering as I saw another monotonic implementation (I think one of the NRF monotonis? idk) that used a 32 bit atomic and hence are also only 63 bit, but that makes sense when doing the math. The SoC probably won't survive (2^63 / 2^15 seconds ~= 8,9 * 10^6 years :) )

Bnyro:
If I didn't fail the math, a 32bit atomic should suffice for the LETimer as well:
```

  ((2^40 / 3600) / 24) / 365

    = 34865.3
```
i.e. 34k years

Bnyro:
If I did the math right, the LETimer will overflow after 8.5 minutes. So it should be possible to actually test this without modifying the timer's counter or doing other weird hacks as I did for the RTCC. I'll do that later.

Finomnis:
> If I didn't fail the math, a 32bit atomic should suffice for the LETimer as well:
> 
> ```
> 
>   ((2^40 / 3600) / 24) / 365
> 
>     = 34865.3
> ```
> 
> i.e. 34k years

I think you forgot to include the tick rate. This assumes 1hz. If it ticks at 32khz, it would be one year, which isn't enough

Bnyro:
I think it should be included?
>I think you forgot to include the tick rate. This assumes 1hz. If it ticks at 32khz, it would be one year, which isn't enough
>
The LETimer ticks 2^15 times per second and has a width of 2^24. That means we get 2^9 seconds per period.

Then, from the AtomicU32, we get 2^32 / 2 = 2^31.

Summa summarum, that makes 2^9 * 2^31 seconds. Which is equivalent to 34865.3 years.

Finomnis:
EDIT: I was wrong, you are of course right :) 

Your timer has 24 bits and when you have a overflow counter of 32 bits, with one bit of overlap, that's 55 bits. With 32kHz (15 bits) that's indeed 2^40 seconds.

Ah that's where you got your 2^40 from. Yeah that should definitely be enough.

rtic-monotonics/src/silabs/letimer.rs
```diff
@@ -0,0 +1,270 @@
...
+        // COMP0: Dynamic wakeup, COMP1: Half-period interrupt, UF: Underflow
+        timer.ien().modify(|w| {
+            w.set_comp0(true);
+            w.set_comp1(true);
+            w.set_uf(true)
+        });
+
+        // configure half-period compare register to be triggered exactly
+        // when have of the period's time has passed
+        timer.comp1().write(|w| w.set_comp1(0x80_000));
```
andresv: I think it should be like:

```rust
const U24_MAX: u32 = 0x00FF_FFFF;
const HALF_PERIOD_UP: u32 = 0x0080_0000;
// COMP1 match value: fires when `U24_MAX - cnt` crosses HALF_PERIOD_UP.
const HALF_PERIOD_CNT: u32 = U24_MAX - HALF_PERIOD_UP;
...

timer.en().write(|w| w.set_en(true));
timer.comp1().write(|w| w.set_comp1(HALF_PERIOD_CNT));
timer.cmd().write(|w| w.set_start(true));
```

Finomnis: Good catch

Bnyro: So essentially we should use `0x7F_FFFF` instead `0x80_0000`, right? I agree that this is technically correct, but just for my understanding regarding https://docs.rs/rtic-time/latest/rtic_time/half_period_counter/, would this make an actual difference in practice?
> Both the overflow interrupt and the half-way interrupt must be processed within half a timer period
>
As I understand it doesn't really matter when exactly the interrupt is processed / the counter incremented, as long as its within half a period.

Or are you also referring to the order of the enable and start instructions here?

Finomnis: It's absolutely detrimental that the interrupt gets processed between the half way point and the overflow. If it gets processed between 0x7ffff and 0x80000, the returned monotonics value could be off by an entire timer wrap for a short period. At 32khz, that's a time period of around 40 us, it's well possible that the interrupt gets processed there

Bnyro: Thanks for clarifying :+1:

Finomnis: You can try, just query the time in an endless loop and check whether the time is newer or equal to the last measurement. You should be able to detect the short period where time jumped forward and back again

Bnyro: I tried it with the code below, but unfortunately it didn't fire :/ (yes, I made sure that the `comp1` interrupt arrives properly)

```rs
    #[task]
    async fn main_task(_cx: main_task::Context) {
        let mut prev_now = Mono::now();
        loop {
            let now = Mono::now();
            if now < prev_now {
                defmt::info!("back to the future");
            }

            prev_now = now;
        }
    }
```

Anyways, I've changed it now to set the proper `comp1` value.

Bnyro:
Looking at the implementation now it looks very similar to your implementation @andresv - I might have been a bit too skeptical about the code you've written.

rtic-monotonics/src/silabs/letimer.rs
Finomnis: I mean I guess that's good :)

If you are curious how critical it is you can try it a few comp1 values to see at what point it does become a problem :)

Also, did you wait long enough? 24 bit takes a while to overflow at 32khz. If the error happens it will happen after 4 minutes 16 seconds, I think.

Bnyro: I wrote to the `cnt` register by hand to jump forward in time (something like `cnt().write(|w| w.set_cnt(HALF_PERIOD_DOWN + 30000))`) - I think that should work? But I can also test it from 0 later :)

Finomnis: Nah, I think that should work

andresv:
No problem, now at least 3 people have looked into it and we can be quite sure it is decent for initial release.
One option would be to merge yours first and then I will do my changes on top of it.

Bnyro:
> No problem, now at least 3 people have looked into it and we can be quite sure it is decent for initial release. One option would be to merge yours first and then I will do my changes on top of it.
>
I think that's a good idea, ready from my side :+1: 

Fyi, I've just squashed all the changes here into one commit to keep the commit history clean.

andresv:
@AfoHT take a look if you have any objections. Idea is to merge @Bnyro PR here first and then I'll rework my PR https://github.com/rtic-rs/rtic/pull/1186 on top of it. My PR adds additional timers support and examples.

korken89 (approved):

Bnyro:
There we go, thank you for merging @korken89!

Thank you a lot for providing the PAC and helping here @andresv.

Thank you for all the explanations regarding the `half-periodic-counter` and math lessons @Finomnis :)

PR #1190, opened 2026-06-17
Add monotonic support for nrf9151/61, support secure mode for nrf91

korken89:


korken89 (commented):
Tested on HW, working as expected

PR #1194, opened 2026-06-29
Assert task local external resources are Send

mciantyre:
The RTIC docs claim

https://github.com/rtic-rs/rtic/blob/28652962d3c46bab79d5fa466815fd68d5cf7cb3/book/en/src/by-example/resources.md?plain=1#L22

However, RTIC doesn't assert this. We see this by undo-ing the changes in analyze.rs and observing a successful build of the new test.

The analysis tries to assert that its resources are Send using the loop shown below. Until now, we never filled the ownerships collection with local resources, so we never took the branch to update send_types.

https://github.com/rtic-rs/rtic/blob/28652962d3c46bab79d5fa466815fd68d5cf7cb3/rtic-macros/src/syntax/analyze.rs#L268-L281

With this commit, RTIC rejects programs it previously accepted. I think users might appreciate the extra check, even if it comes unexpectedly from a non-breaking update.


korken89 (approved):
Thanks for fixing!

PR #1196, opened 2026-07-03
Switch to Rust Edition 2024 on release/v1

elmjag:
Update this repository to use Rust edition 2024.

The major change in Rust 2024, compared to 2021, is how the `unsafe` is handled.

[`unsafe` attributes](https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-attributes.html) need to be explicitly marked. The https://github.com/rtic-rs/rtic/commit/58b2f5b56489913e051eb72c944ae9bf9cfc7fe3 commit deals with that.
[`unsafe fn` bodies](https://rust-lang.github.io/rfcs/2585-unsafe-block-in-unsafe-fn.html) are no longer considered `unsafe` blocks. The https://github.com/rtic-rs/rtic/commit/8f45cf87a58b4a4a2e73c4e41c7ae69f94985ef2 commit adds `unsafe {}` where needed.

There is also some stylistic changes are needed to satisfy [cargo fmt](https://github.com/rtic-rs/rtic/commit/859ace41ff6a54d6b176f3847d7f35c6fc7a7792) and [clippy](https://github.com/rtic-rs/rtic/commit/616545a5791dbc3c915bd4f330828a19d86ba161).


kscz:
This looks like it will require Rust edition 2024 for both building RTICv1 and for consumers using RTICv1 - is that correct? Is there any way to pass through whether or not the `unsafe` attributes are present and allow consumers to be on edition 2021?

jonathanpallant:
Please note that Ferrocene is currently only qualified for use with Edition 2021, so this change means Ferrocene users using RTIC v1 will need to fork it to revert these changes.

https://public-docs.ferrocene.dev/main/safety-manual/rustc/options.html

elmjag:
> This looks like it will require Rust edition 2024 for both building RTICv1 and for consumers using RTICv1 - is that correct? Is there any way to pass through whether or not the `unsafe` attributes are present and allow consumers to be on edition 2021?

No, consumers can still be on edition 2021. I did a quick test, and it seems it works fine using these patches with edition 2021 consumer.

The `unsafe` wrapper is valid edition 2021 syntax, but it's optional. It becomes mandatory with edition 2024.

elmjag:
Not sure anymore if it's a good idea to merge this.

I have now checked that it's possible to use `rtic` on `release/v1` as dependency with edition 2024 code base as is. If switching to edition 2024 creates problems for others, then it's not worth it right now.

However, I suggest that we merge the `unsafe` wrapper bits, without actually changing to edition 2024. These commis:
https://github.com/rtic-rs/rtic/pull/1196/commits/58b2f5b56489913e051eb72c944ae9bf9cfc7fe3
https://github.com/rtic-rs/rtic/pull/1196/commits/c58c3a12a2210eecfe737f7bfd245a3de106be23
https://github.com/rtic-rs/rtic/pull/1196/commits/24ad86a9c16269da1d385c4d8ead35ea23dce532

That will make easier to migrate to later edition easier in the future, if a need arises.

Putting this PR into draft for now.

ps. I'm leaving for vacation, I'm back in the beginning of August.

kscz:
My experience trying to use RTIC v1 with a Rust 2024 codebase was that everything worked until I tried to run a task from ram ( https://rtic.rs/1/book/en/by-example/tips_from_ram.html ). The lack of the unsafe keyword in the tags failed to compile, and if I added the unsafe keyword the RTIC v1 macros failed to compile.

elmjag:
> My experience trying to use RTIC v1 with a Rust 2024 codebase was that everything worked until I tried to run a task from ram ( https://rtic.rs/1/book/en/by-example/tips_from_ram.html ). The lack of the unsafe keyword in the tags failed to compile, and if I added the unsafe keyword the RTIC v1 macros failed to compile.

I'm not sure specifically about running tasks from ram. Maybe you should try using latest version of [rtic-syntax](https://github.com/rtic-rs/rtic-syntax). The latest version of rtic-syntax on github have been upgraded to use [`syn 2` crate](https://github.com/rtic-rs/rtic-syntax/commit/3a7e983861382838ed0b69804092a2b788b2c7bf). It should in principle include `unsafe` in all macro generated code.

On our side, we did manage to successfully build and run our Rust 2024 codebase, using RTIC v1. We used latest source code of RTIC v1 and rtic-syntax crates, directly from github.

elmjag:
Turns out there is no need to switch this repository to Rust 2024, in order to use this crate from Rust 2024 consumer code.

As switching to Rust 2024 creates problems, for Rust 2021 consumer code, let's skip these changes for the time being.

kscz:
Just to clarify my comment - Rust 2024 codebases which would like to use RTIC v1 will not be able to if they attempt to tag a function with a linker section (for example, when trying to run code from RAM), because the unsafe keyword in the linker section tag does not propagate successfully.

The same question applies to your change - I was asking if you had tested it with a Rust 2021 consumer which attempted to run code from RAM or otherwise tag a function with an unsafe keyword required.

PR #1199, opened 2026-07-15
docs(book): explain monotonics core concept and mention `half_period_counter`

Bnyro:
In my opinion the current documentation leaves it unclear what the "core" ideas of monotonics are, and it doesn't mention the `half_period_counter` that probably most peripheral timer implementations need.

This is inspired by the explanations of @Finomnis in #1184.

I'm not 100% happy with it yet, feedback appreciated.

### DIFF
--- a/book/en/src/monotonic_impl.md
+++ b/book/en/src/monotonic_impl.md
@@ -8,11 +8,13 @@ The [`rtic-time`] framework is flexible because it can use any timer which has c
 
 For RTIC 2.0, we assume that the user has a time library, e.g. [`fugit`], as the basis for all time-based operations when implementing [`Monotonic`]. These libraries make it much easier to correctly implement the [`Monotonic`] trait, allowing the use of almost any timer in the system for scheduling.
 
-The trait documents the requirements for each method. There are reference implementations available in [`rtic-monotonics`] that can be used for inspriation.
+The trait documents the requirements for each method. There are reference implementations available in [`rtic-monotonics`] that can be used for inspiration.
 
-- [`Systick based`], runs at a fixed interrupt (tick) rate - with some overhead but simple and provides support for large time spans
+- [`Systick based`], runs at a fixed interrupt (core tick) rate - simple, but comes with some overhead and can only be used while the core is running
 - [`RP2040 Timer`], a "proper" implementation with support for waiting for long periods without interrupts. Clearly demonstrates how to use the [`TimerQueue`] to handle scheduling.
 - [`nRF52 timers`] implements monotonic & Timer Queue for the RTC and normal timers in nRF52's
+
+Often, hardware counters only have a width of only 16 or 32 bit, meaning that they would overflow after some minutes to years, depending on the frequency of the counter peripheral. To overcome this issue, monotonic implementations for such counters must use an atomic overflow counter (e.g. [`portable_atomic::AtomicU32`]) to correctly calculate the current time instant. This leads to a race condition between the overflow counter and the time register, which isn't trivial to solve. Hence, a half period counter must be used. Please read and refer to [`rtic_time::half_period_counter`], which provides helper methods and examples for implementing this logic.
 
 ## Contributing
 
@@ -30,9 +32,11 @@ Contributing new implementations of `Monotonic` can be done in multiple ways:
 [`rtic-time::Monotonic`]: https://docs.rs/rtic-time/latest/rtic_time/trait.Monotonic.html
 [`Monotonic`]: https://docs.rs/rtic-time/latest/rtic_time/trait.Monotonic.html
 [`TimerQueue`]: https://docs.rs/rtic-time/latest/rtic_time/timer_queue/struct.TimerQueue.html
+[`rtic_time::half_period_counter`]: https://docs.rs/rtic-time/latest/rtic_time/half_period_counter/
+[`portable_atomic::AtomicU32`]: https://docs.rs/portable-atomic/latest/portable_atomic/struct.AtomicU32.html
 
 ## The timer queue
 
-The timer queue is implemented as a list based priority queue, where list-nodes are statically allocated as part of the  `Future` created when `await`-ing a Future created when waiting for the monotonic. Thus, the timer queue is infallible at run-time (its size and allocation are determined at compile time).
+The timer queue is implemented as a list based priority queue, where list-nodes are statically allocated as part of the `Future` created when `await`-ing a Future created when waiting for the monotonic. Thus, the timer queue is infallible at run-time (its size and allocation are determined at compile time).
 
 Similarly the channels implementation, the timer-queue implementation relies on a global *Critical Section* (CS) for race protection. For the examples a CS implementation is provided by adding `--features test-critical-section` to the build options.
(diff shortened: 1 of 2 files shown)


book/en/src/monotonic_impl.md
```diff
@@ -1,5 +1,9 @@
 # The magic behind Monotonics
 
+Monotonics can be thought of like timekeepers, they measure the amount of time elapsed since the start of the monotonic (usually the start of the application).
+In RTIC, monotonics can be used for delaying the execution of code, see the [Delay and Timeout using Monotonics] chapter.
```
Finomnis: And for time measurement. Also it could be mentioned that their resolution is very high for very precise scheduling and time measurements, with a resolution down to single digit nanoseconds. (for example stm32h7 has a 240mhz monotonic, which can measure with a resolution of 2.5ns. Of course both time measurement and scheduling still have an overhead though)

book/en/src/monotonic_impl.md
```diff
@@ -1,5 +1,9 @@
 # The magic behind Monotonics
 
+Monotonics can be thought of like timekeepers, they measure the amount of time elapsed since the start of the monotonic (usually the start of the application).
```
AfoHT: Good stuff! What do you think about moving this block into the `[Delay and Timeout using Monotonics]` chapter, which would introduce "monotonics" before using them.

Assuming people read the book following the numbered steps that is the least surprising imho.

Now this links back to the "how to use" it, and while there's a link how "how to implement the monotonics" it's not clear it's also containing the "what is a monotonic".

Curious what you think would have been most helpful when you needed it :)

book/en/src/monotonic_impl.md
```diff
@@ -8,12 +12,14 @@ The [`rtic-time`] framework is flexible because it can use any timer which has c
 
 For RTIC 2.0, we assume that the user has a time library, e.g. [`fugit`], as the basis for all time-based operations when implementing [`Monotonic`]. These libraries make it much easier to correctly implement the [`Monotonic`] trait, allowing the use of almost any timer in the system for scheduling.
 
-The trait documents the requirements for each method. There are reference implementations available in [`rtic-monotonics`] that can be used for inspriation.
+The trait documents the requirements for each method. There are reference implementations available in [`rtic-monotonics`] that can be used for inspiration.
 
-- [`Systick based`], runs at a fixed interrupt (tick) rate - with some overhead but simple and provides support for large time spans
+- [`Systick based`], runs at a fixed interrupt (tick) rate - with some overhead but simple and provides support for large timespans
 - [`RP2040 Timer`], a "proper" implementation with support for waiting for long periods without interrupts. Clearly demonstrates how to use the [`TimerQueue`] to handle scheduling.
 - [`nRF52 timers`] implements monotonic & Timer Queue for the RTC and normal timers in nRF52's
```
Finomnis, line 15: Bunch of timers missing, for example stm32

book/en/src/monotonic_impl.md
```diff
@@ -8,12 +12,14 @@ The [`rtic-time`] framework is flexible because it can use any timer which has c
...
 
+Often, hardware counters only have a width of only 16 or 32 bit, meaning that they would overflow after some minutes to years, depending on the frequency of the counter peripheral. To overcome this issue, monotonic implementations for such counters must use an atomic overflow counter (e.g. [`portable_atomic::AtomicU32`]) to correctly calculate the current time instant. Please use and refer to [`rtic_time::half_period_counter`], which provider helper methods and examples for implementing this logic.
```
Finomnis: Might want to mention that solving the race condition between overflow counter and time register isn't trivial, hence the need for the half period counter.

book/en/src/monotonic_impl.md
```diff
@@ -8,12 +8,14 @@ The [`rtic-time`] framework is flexible because it can use any timer which has c
 
 For RTIC 2.0, we assume that the user has a time library, e.g. [`fugit`], as the basis for all time-based operations when implementing [`Monotonic`]. These libraries make it much easier to correctly implement the [`Monotonic`] trait, allowing the use of almost any timer in the system for scheduling.
 
-The trait documents the requirements for each method. There are reference implementations available in [`rtic-monotonics`] that can be used for inspriation.
+The trait documents the requirements for each method. There are reference implementations available in [`rtic-monotonics`] that can be used for inspiration.
 
-- [`Systick based`], runs at a fixed interrupt (tick) rate - with some overhead but simple and provides support for large time spans
+- [`Systick based`], runs at a fixed interrupt (tick) rate - with some overhead but simple and provides support for large timespans
```
Finomnis: Every monotonic supports large timespans, I think this comment is misleading

book/en/src/monotonic_impl.md
Bnyro: > What do you think about moving this block into the [Delay and Timeout using Monotonics] chapter, which would introduce "monotonics" before using them.
>
That's a good idea, thanks! I've moved the monotonics explanation to the beginning of the `[Delay and Timeout using Monotonics] ` chapter now.

> Now this links back to the "how to use" it, and while there's a link how "how to implement the monotonics" it's not clear it's also containing the "what is a monotonic".
>
True, cyclic references are usually not that great. In this case it also makes a lot of sense because new developers that start using RTIC will probably not start by implementing monotonics themselves, so they likely wouldn't get the idea to read the `Magic behind Monotonics` chapter anyways, so they might never figure out what monotonics are exactly.

book/en/src/monotonic_impl.md
AfoHT: Nice!

I think it can be worthwhile mentioning that HW tend to implement prescalers too, to help control the timer rate. But as application creator one must balance against the resolution you want/need for your particular timing constraints.

Or even use multiple monotonics! RTC clock ticking away for the years and days, something speedier for the radio/high speed peripherals after waking up... but I'm getting ahead of myself :) Big topic this, but one small step at a time! 🚀

Bnyro:
Thanks for your reviews @Finomnis and @AfoHT. I'll look into them tomorrow, gotta write some other small embedded programs now :)

book/en/src/monotonic_impl.md
Bnyro: True - I only fixed that the typo here, but this should definitely be cleaned up

book/en/src/monotonic_impl.md
Finomnis: What's the disadvantage of just having the timer run at max speed? Power draw?

Bnyro: I already expected that you would say that because I didn't write anything about the race condition and the awesomeness of the `half_periodic_counter` yet :)

Bnyro: The oscillator ticks at the same rate anyways no matter the prescaler, so in theory the additional power draw should be very very small, right? Okay right, we have more overflow interrupts that wake up the core, that might be an aspect.

book/en/src/monotonic_impl.md
Bnyro: I think the idea here is to mention some timers that are good reference implementations if somebody wants to implement a new monotonic, not to list all of them. The full list can be found in the `rtic_monotonics` crate documentation on `docs.rs` anyways.

book/en/src/monotonic_impl.md
Bnyro: I've added a note about prescalers now, but no pro contra list of why you should use a maximum / minimum prescaler setting

AfoHT (approved):
Thank you!

⏲️  🚀

PR #1200, opened 2026-07-18
Replace shady use of `rustc-cfg=features` with actual CFGs

datdenkikniet:
`rtic` and `rtic-macros` do some shenanigans with activating "features" (that are not activatable by library consumers), and `rtic` has a bunch of optional deps that can be enabled by library consumers.

Additionally, both crates use `compile_error` to convey problems with activating too many/too few features.

Both of these things can (and should) be handled by the build script instead. It prints out a much easier to parse error message, and doesn't even attempt to compile anything. 

This PR performs the suggested replacements, removes a funky `allow` along the way, and is technically breaking: some previously available features (intended only for internal use) have been removed from `rtic` and `rtic-macros`. Compilation will stop working for users of `rtic` that activated one of those features explicitly alongside their backend (i.e. `riscv` + `riscv-mecall-backend`).

To clarify, `rtic`'s feature set sees the following diff (used word-diff in an attempt to make it look a bit more obvious):
image

while `rtic-macros` feature set sees the following diff:
image

all of the removed features have always been "only activate these dependencies if other features are selected" and "must use in combination with some other feature" types of features: functionality-wise, nothing changes. However, others may have selected these features before. If we want to avoid this kind of breaking change, I can revert the non-`dep:`-to-`dep:` change.

### DIFF
--- a/rtic-macros/CHANGELOG.md
+++ b/rtic-macros/CHANGELOG.md
@@ -7,6 +7,11 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 
 ## [Unreleased]
 
+### Changed
+
+- Removed explicit `riscv-slic` feature (this was only used to activate functionality shared between `riscv-clint` and `riscv-mecall`).
+- Selecting no or more than one implementation now fails in the build script instead of during compilation.
+
 ## [v2.3.0] - 2026-07-08
 
 ### Fixed
--- a/rtic-macros/Cargo.toml
+++ b/rtic-macros/Cargo.toml
@@ -40,9 +40,8 @@ riscv-esp32c3 = []
 riscv-esp32c6 = []
 # riscv-clic = []
 # riscv-ch32 = []
-riscv-slic = []
-riscv-clint = ["riscv-slic"]
-riscv-mecall= ["riscv-slic"]
+riscv-clint = []
+riscv-mecall = []
 
 # backend API test
 test-template = []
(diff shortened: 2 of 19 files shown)


datdenkikniet:
Huh: it also seems that this change unbricks `rustfmt` for (some parts) of `rtic-macros`, so I've added a final commit to format that package.

datdenkikniet:
Feedback from claude (or other LLM) through @korken89:

> Nondeterministic and duplicated check-cfg output in rtic/build.rs

Fixed with the suggested list of tuples.

> Because the PR also drops cargo:rerun-if-changed=build.rs

Re-added

> rtic-macros/build.rs treats every feature as a backend ... An explicit backend list (as in rtic/build.rs)

Won't fix. `rtic/build.rs` is performing a N-to-M mapping, `rtic-macros` just activates a shared implementation for some `riscv` targets. If we ever expand this we can rework it then.

> rtic/build.rs lists the offending backends in "More than one backend selected: {backends:?}"; the rtic-macros version doesn't, and which feature ends up in feature after pop() depends on env::vars() order.

Won't fix. `rtic-macros` is not intended for use by end-users, only through `rtic`, and the error message already clearly indicates a path to resolution. We don't print the `pop`'ed feature in either build script if more than one feature is selected, so I don't see how it is relevant. 

> rtic-macros continues executing after the "more than one" error instead of returning

Fixed

> Pick cargo:: everywhere.

Fixed

AfoHT (approved):
Really neat!

I had _forgotten_ about those 

```
    feature = "cortex-m-source-masking",
    feature = "cortex-m-basepri",
    feature = "test-template",
    feature = "riscv-esp32c3",
    feature = "riscv-esp32c6",
    feature = "riscv-slic",
)))]
compile_error!("No backend selected");
```
eye-sores xD

Looks good to me, had a play-around locally as well 👍 

@korken89 did you have something to add?

korken89:
LGTM

PR #1201, opened 2026-07-20
rtic-macros: add `#[pre_rtic_hook]` function hook

glaeqen:
`#[pre_rtic_hook]` hook allows injecting code that runs directly after the jump from reset handler to main, that is right before RTIC starts its self-initialization. Useful for cases when using cortex-m-rt pre-init would be too early (.bss/.data not initialized) while running code in RTIC init would be too late (NVIC preconfigured, executor memory allocated by RTIC-generated `main`).

### DIFF
--- a/rtic-macros/src/codegen/main.rs
+++ b/rtic-macros/src/codegen/main.rs
@@ -48,6 +48,14 @@ pub fn codegen(app: &App, analysis: &Analysis) -> TokenStream2 {
 
     let main = util::suffixed("main");
     let init_name = &app.init.name;
+    let call_pre_rtic_hook = if let Some(pre_rtic_hook) = &app.pre_rtic_hook {
+        let name = &pre_rtic_hook.name;
+        quote!(
+            let _: () = #name();
+        )
+    } else {
+        quote!()
+    };
 
     let init_args = if app.args.core {
         quote!(core.into(), executors_size)
@@ -63,6 +71,8 @@ pub fn codegen(app: &App, analysis: &Analysis) -> TokenStream2 {
         #[doc(hidden)]
         #[no_mangle]
         unsafe extern "C" fn #main() -> ! {
+            #call_pre_rtic_hook
+
             #(#assertion_stmts)*
 
             #(#pre_init_stmts)*
--- a/rtic-macros/src/codegen/pre_rtic_hook.rs
+++ b/rtic-macros/src/codegen/pre_rtic_hook.rs
@@ -1,0 +1,31 @@
+use crate::syntax::ast::App;
+use crate::{
+    analyze::Analysis,
+};
+use proc_macro2::TokenStream as TokenStream2;
+use quote::quote;
+
+/// Generates the `#[pre_rtic_hook]` function if needed
+pub fn codegen(app: &App, _: &Analysis) -> TokenStream2 {
+    if let Some(pre_rtic_hook) = &app.pre_rtic_hook {
+        let attrs = &pre_rtic_hook.attrs;
+        let name = &pre_rtic_hook.name;
+        let stmts = &pre_rtic_hook.stmts;
+        let user_idle = if !pre_rtic_hook.is_extern {
+            Some(quote!(
+                #(#attrs)*
+                fn #name() {
+                    #(#stmts)*
+                }
+            ))
+        } else {
+            None
+        };
+
+        quote!(
+            #user_idle
+        )
+    } else {
+        quote!()
+    }
+}
--- a/rtic-macros/src/syntax/ast.rs
+++ b/rtic-macros/src/syntax/ast.rs
@@ -44,6 +44,9 @@ pub struct App {
 
     /// Async software tasks: `#[task]`
     pub software_tasks: Map<SoftwareTask>,
+
+    /// The `#[pre_rtic_hook]` function
+    pub pre_rtic_hook: Option<PreRticHook>,
 }
 
 /// Interrupts used to dispatch software tasks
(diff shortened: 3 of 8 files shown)


glaeqen:
@datdenkikniet

rtic-macros/src/codegen/main.rs
```diff
@@ -48,6 +48,14 @@ pub fn codegen(app: &App, analysis: &Analysis) -> TokenStream2 {
 
     let main = util::suffixed("main");
     let init_name = &app.init.name;
+    let call_pre_rtic_hook = if let Some(pre_rtic_hook) = &app.pre_rtic_hook {
```
datdenkikniet, line 51: I think `Option<TokenStream2>` is also `IntoTokens`, so this can just be a `map()`

rtic-macros/src/syntax/ast.rs
```diff
@@ -369,3 +372,21 @@ pub type SharedResources = Map<Access>;
...
+    pub attrs: Vec<Attribute>,
+
+    /// The name of the `#[pre_rtic_hook]` function
+    pub name: Ident,
+
+    /// The statements that make up this `#[pre_rtic_hook]` function
+    pub stmts: Vec<Stmt>,
+
+    /// The `#[pre_rtic_hook]` function is declared externally
+    pub is_extern: bool,
```
datdenkikniet, line 390: This is probably not up to this PR to fix, but (if I understand it correctly), surely `is_extern ^ stmts.is_empty()`. These two fields should really be

```rust
enum Definition {
    External,
    InPlace(Vec<Stmt>),
}
```

datdenkikniet (approved):
Only nits, and of course you should make `clippy` happy, but this change looks like a good one to me.

Use-case is a bit niche, but that should not stop this from being accepted IMO. Let's see if others have opinions :)

rtic-macros/src/codegen/pre_rtic_hook.rs
```diff
@@ -0,0 +1,31 @@
...
+use proc_macro2::TokenStream as TokenStream2;
+use quote::quote;
+
+/// Generates the `#[pre_rtic_hook]` function if needed
+pub fn codegen(app: &App, _: &Analysis) -> TokenStream2 {
+    if let Some(pre_rtic_hook) = &app.pre_rtic_hook {
+        let attrs = &pre_rtic_hook.attrs;
+        let name = &pre_rtic_hook.name;
+        let stmts = &pre_rtic_hook.stmts;
+        let user_idle = if !pre_rtic_hook.is_extern {
```
AfoHT, line 14: Copy paste remnants? `user_idle`

rtic-macros/src/syntax/parse/app.rs
```diff
@@ -480,11 +500,29 @@ impl App {
...
                                     }
                                 }
+                            } else if let Some(pos) = item
+                                .attrs
+                                .iter()
+                                .position(|attr| util::attr_eq(attr, "pre_rtic_hook"))
+                            {
+                                let _ = item.attrs.remove(pos);
+
+                                // If an idle function already exists, error
```
AfoHT, line 510: Same here with "idle"

AfoHT (changes requested):
Seems like a good feature to have, but why not using the existing `pre_init` modules and functions?

I did this: https://github.com/glaeqen/rtic/pull/1 to articulate what I had in mind, would be too tedious in the review boxes :D

Now there's just Send/Sync asserts before it, resources comes afterwards.

Some tests would be nice!

Also I'm really curious about the use-case details, I can only speculate; some particular HW needs setup/tweaking before RTIC can make sense of what it provides?

rtic-macros/src/codegen/pre_rtic_hook.rs
AfoHT: Stale if going with the PR linked

glaeqen: oops

glaeqen:
@AfoHT 

> Seems like a good feature to have, but why not using the existing pre_init modules and functions?

cortex_m_rt's pre_init runs before .bss and .data is initialized which makes running a code touching 0-initialized or initialized statics into UB. PS: The alternative solution here is to avoid running such code there in the first place which is something I considered but, again, this is somewhat non-trivial code that would have to be duplicated (extracted from HAL) and maybe narrowed down to what we need (eg. small silly things like defmt print-outs become somewhat dangerous if .bss/.data is effectively .uninit)

> I did this: https://github.com/glaeqen/rtic/pull/1 to articulate what I had in mind, would be too tedious in the review boxes

> Now there's just Send/Sync asserts before it, resources comes afterwards.

This looks like a great improvement, I didn't realize which pre_init statements you meant before but it is nicer.

> Some tests would be nice!

I didn't want to commit too much effort at first before I get anything that could be green-lit.

> Also I'm really curious about the use-case details, I can only speculate; some particular HW needs setup/tweaking before RTIC can make sense of what it provides?

We work on a bootloader that needs to call non-as-usually-trivial code before doing the trampoline jump but we don't want to initialize HW too much to give the jumped-into app as much of clean slate situation as possible. However the use-case have become much weaker the more time I spent with the problem and I started to think that calling our code as a first thing in the regular init is probably good enough. NVIC::unmask does nothing unless explicitly NVIC::pend (which the booted-into app won't do unless it decides to re-use the same vector and then provides a correct handle anyway) or some peripheral enabled interrupts (which after the reset won't be the case). Executor allocation does not seem to do anything meaningful either, it just writes a pointer to a local variable on main's stack to a static variable (unless I missed sth). So it seems like the worst thing that can happen as far as the leftover state goes is a bunch of unmasked NVICs that are not used. But maybe I missed something.

If that's the case, we could drop this feature and continue living our lives. @datdenkikniet  is also leaning in that direction right now but he can pitch in 😄

AfoHT:
> > Seems like a good feature to have, but why not using the existing pre_init modules and functions?
> 
> cortex_m_rt's pre_init runs before .bss and .data is initialized, <...>

As you indicate later on in the comment, but just to make myself clearer: I'm talking about RTIC's [`pre_init`](https://github.com/rtic-rs/rtic/blob/master/rtic-macros/src/codegen/pre_init.rs), [`post_init`](https://github.com/rtic-rs/rtic/blob/master/rtic-macros/src/codegen/post_init.rs) modules here :) This namespace seems really packed :D  

> This looks like a great improvement, I didn't realize which pre_init statements you meant before but it is nicer. 
> I didn't want to commit too much effort at first before I get anything that could be green-lit.

Fair!

> But maybe I missed something.
> 
> If that's the case, we could drop this feature and continue living our lives. @datdenkikniet is also leaning in that direction right now but he can pitch in 😄

Sometimes doing the experimental thing and actually testing it on HW is the only way to be really really sure!

perlindgren:
Closed, please re-open in case needed.

PR #1205, opened 2026-07-22
rtic-monotonics: move chip selection to the metapac dependencies

korken89:


korken89:
@Finomnis feel free to give this a look when/if you have time :)

Finomnis:
Uuf, that change is a mouthful :D uncertain when I'll find the time

rtic-monotonics/src/silabs/letimer.rs
```diff
@@ -129,23 +108,19 @@ impl TimerBackend {
...
-        &LETIMER0
-    }
 }
 
 static TIMER_QUEUE: TimerQueue<TimerBackend> = TimerQueue::new();
 
 // used to widen letimer from 24 to 64 bits
 // maximum value is (64 - 24) bit * 2 (due to half period) = 41bit
-const LETIMER_HALF_PERIOD_COUNTER: AtomicU32 = AtomicU32::new(0);
+static LETIMER_HALF_PERIOD_COUNTER: AtomicU32 = AtomicU32::new(0);
```
Bnyro, line 118: Good catch! Not sure how I failed that...

Bnyro:
The EFR part looks good to me, can't say anything about the other monotonics because I'm not familiar with them.

rtic-monotonics/src/silabs/letimer.rs
korken89: It flew past my eyes so I fixed it in passing, it should have been its own PR though

korken89:
Thanks for checking the EFR part @Bnyro  !

A small idea for a future improvement on the silabs side while you're here.

The reason the LETIMER `Mono::start()` lost its typed parameters is that the types are only nameable via the versioned paths (`letimer_v0`/`letimer_v1`, `cmu_v1`/`v3`/`v4`/`v7`), and those differ per chip.

Basically, `silabs-metapac` already generates version-neutral consts like `pub const LETIMER0: letimer_v1::Letimer`, so if the generator in `silabs-data-generated` also emitted aliases next to them, e.g. `pub use letimer_v1 as letimer;` and `pub use cmu_v3 as cmu;`, then `silabs_metapac::letimer::Letimer` would be nameable on every chip.

With that in a `0.5` we can restore `Mono::start(timer: Letimer, cmu: &Cmu)` and move the RTCC over to the neutral paths as well - exact types, no per-family features. The version requirement is open-ended now, so it's quite a small follow-up here once released.

What do you think? I can prepare the generator PR if you want, or feel free to steal the idea :)

rtic-monotonics/src/silabs/letimer.rs
Finomnis: Especially if we want to release a final v2 Bugfix release

rtic-monotonics/src/silabs/rtcc.rs
```diff
@@ -134,35 +126,34 @@ impl TimerQueueBackend for TimerBackend {
...
         if interrupt_flag.cc1() {
-            Self::rtcc().if_clr().write(|w| w.set_cc1(true));
+            RTCC.if_clr().write(|w| w.set_cc1(true));
 
             let prev = RTCC_HALF_PERIOD_COUNTER.fetch_add(1, Ordering::Relaxed);
-            ::core::assert!(prev % 2 == 0, "Monotonic must have skipped an interrupt!");
+            ::core::assert!(
+                prev.is_multiple_of(2),
+                "Monotonic must have skipped an interrupt!"
+            );
```
Finomnis, line 152: Either also change `prev % 2 == 1` on line 159, or leave this line as-is. I think those two lines should use the same pattern, and the pattern used matches the recommended pattern in https://docs.rs/rtic-time/latest/rtic_time/half_period_counter/#example.

rtic-monotonics/src/stm32/meta.rs
```diff
@@ -0,0 +1,180 @@
+//! Compile-time lookups into the `stm32-metapac` metadata.
```
Finomnis, line 1: I assume this file is mostly copy-paste from build.rs?

I don't really have the time and required background knowledge to review this file from scratch ...

rtic-monotonics/CHANGELOG.md
```diff
@@ -7,7 +7,24 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
...
-- Added Silabs `TIMER0`-`TIMER9` monotonics
+### Changed
+
+- [breaking] STM32 and Silabs chips are now selected via the chip feature on your own
+  `stm32-metapac`/`silabs-metapac` dependency; the per-chip features are removed (see README)
+- [breaking] Silabs monotonics are selected with `silabs_letimer0`, `silabs_rtcc` and
+  `silabs_timer0`-`silabs_timer9`; the per-family features are removed
+- [breaking] Silabs LETIMER `Mono::start()` no longer takes the timer and CMU objects
+- STM32 monotonics no longer use a build script
+- `stm32-metapac` (`>=16`), `imxrt-ral` (`>=0.5.4`) and `silabs-metapac` (`>=0.4`) accept any
```
Finomnis, line 18: Little bit scared about breaking changes in stm32-metapac that might break our code, but we will see. Let's go with it for now.

rtic-monotonics/CHANGELOG.md
```diff
@@ -7,7 +7,24 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
...
+  newer release, so the dependency can unify with your HAL's copy
+
+### Fixed
+
+- Silabs LETIMER half-period counter did not persist, wrapping time every timer period
```
Finomnis: @AfoHT Backport LETIMER fix to rtic_monotonics/release-2.2

Bnyro:
> The reason the LETIMER `Mono::start()` lost its typed parameters is that the types are only nameable via the versioned paths (`letimer_v0`/`letimer_v1`, `cmu_v1`/`v3`/`v4`/`v7`), and those differ per chip.
> 
> Basically, `silabs-metapac` already generates version-neutral consts like `pub const LETIMER0: letimer_v1::Letimer`, so if the generator in `silabs-data-generated` also emitted aliases next to them, e.g. `pub use letimer_v1 as letimer;` and `pub use cmu_v3 as cmu;`, then `silabs_metapac::letimer::Letimer` would be nameable on every chip.
> 
> With that in a `0.5` we can restore `Mono::start(timer: Letimer, cmu: &Cmu)` and move the RTCC over to the neutral paths as well - exact types, no per-family features. The version requirement is open-ended now, so it's quite a small follow-up here once released.
> 
> What do you think? I can prepare the generator PR if you want, or feel free to steal the idea :)
>
Yes, sounds like a very good idea in my ears! @andresv is the maintainer of `silabs-metapac`, so it might be better to discuss it with him first. But from my side I think it would be great to do that if you're motivated :)

korken89:
I nerd-sniped myself and fixed a PR :)

Finomnis (changes requested):

rtic-monotonics/src/stm32/meta.rs
korken89: Correct, plus a bit of refactoring to make it work outside `build.rs`.

rtic-monotonics/CHANGELOG.md
korken89: Ah yeah, this is the pragmatic fix.

rtic-monotonics/CHANGELOG.md
korken89: Same, but I expect that we will get issues reported then. Not sure how to get it perfectly future-proof right now.

rtic-monotonics/CHANGELOG.md
korken89: I'm pushing a backport to the branch - I see that it's not released yet.

rtic-monotonics/CHANGELOG.md
Finomnis: Not really possible, might always have to release a new version once a breaking upstream change occurs. But that's kind of normal for all crates tbh, I personally wouldn't mind.

The actual improvement is that we no longer need to release new versions when the pac **minor** version changes, like when they add new chips.

Finomnis: So with other words, I personally would not mind a slightly more narrow dependency, like `16` instead of `>=16`, so we still need to release new versions when the PAC releases a new major version. But in the end I'm fine with both :)

rtic-monotonics/src/silabs/rtcc.rs
korken89: Fixed

rtic-monotonics/CHANGELOG.md
Finomnis: The real issue I see is that if we release a `>=16`, and it is broken in the future, we might have to release a major version any way but we are then stuck with an old version with broken dependencies. Which we might have to yank and replace with a more narrow dependency definition. Not sure which one is more hastle.

rtic-monotonics/src/silabs/rtcc.rs
Finomnis: Why did you change it in the first place? Is that a new clippy lint? If so, I personally do think the previous version is kind of easier to read ... I'm not a big fan of clippy's decisions lately to be honest

Finomnis: But if you insist this is better we should change it **everywhere**, in all monotonics and in the documentation of rtic-time's half period monotonic docs

rtic-monotonics/CHANGELOG.md
korken89: I did a test by going through all major versions 16..=21 and this change works on all - so it seems like a loose requirement is OK. Not sure if I missed some incompatibility though, but at least it builds.

rtic-monotonics/CHANGELOG.md
korken89: `rtic-monotonics` v2.2.0 is now released and it includes the backport

rtic-monotonics/CHANGELOG.md
Finomnis: Then it's fine for me

rtic-monotonics/src/silabs/rtcc.rs
Finomnis: @korken89 Unsure if you saw this comment, we might want to reopen until reaching a conclusion

Finomnis: Never mind, just saw your changes.

Finomnis (approved):

Finomnis:
LGTM

rtic-monotonics/CHANGELOG.md
```diff
@@ -20,14 +29,15 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
...
+- Silabs `TIMER0`-`TIMER9` monotonics
 - Silabs (EFR32, EFM32) support using the inbuilt RTCC / LETIMER peripherals
 - nRF9151 and nRF9161 support (RTC and TIMER monotonics), backed by `nrf9120-pac`
 - Secure and non-secure variants for all nRF91 chips: `nrf9160-s`/`nrf9160-ns`, `nrf9151-s`/`nrf9151-ns`, `nrf9161-s`/`nrf9161-ns`
-- A lot of STM32 chip feature flags
 - Cortex-M `systick` can be configured with its external clock source
 
 ### Fixed
 
+- Silabs LETIMER half-period counter did not persist, wrapping time every timer period
```
Finomnis: FYI this changelog is pointless, Silabs wasn't released in v2.1, so there was nothing to fix

korken89: I moved it to the 2.2.0 release I made, so it should be correct?

Finomnis: It's under the 2.2.0 release, but 2.1 does not contain the Silabs code, so in the same release you both implement and fix the driver. Usually a fix refers to something that was already released, had a bug, and then got fixed in the current release. It's very unusual to list of a fix that refers to code added in the same release

Finomnis: So yes, it's of course not part of 3.0, but it shouldn't be part of the 2.2 release logs

korken89: Ahh now I understand, I've removed the "fixed" line.

Finomnis:
Looks still good to me :)

Finomnis:
Anything still blocking this?

korken89:
I'd like @AfoHT or anyone more to give it a look before merge, but I think it's fine to merge soon if no-one has time

rtic-monotonics/src/silabs/letimer.rs
```diff
@@ -88,39 +64,42 @@ impl TimerBackend {
...
-        timer.ctrl().write(|w| match tick_rate_hz {
-            32_768 => w.set_cntpresc(Cntpresc::Div1),
-            16_384 => w.set_cntpresc(Cntpresc::Div2),
-            8_192 => w.set_cntpresc(Cntpresc::Div4),
-            4_096 => w.set_cntpresc(Cntpresc::Div8),
-            2_048 => w.set_cntpresc(Cntpresc::Div16),
-            1_024 => w.set_cntpresc(Cntpresc::Div32),
-            512 => w.set_cntpresc(Cntpresc::Div64),
-            256 => w.set_cntpresc(Cntpresc::Div128),
+        CMU.clken0().modify(|w| w.set_letimer0(true));
```
datdenkikniet: This is a R-M-W of a register that can also enable other timers (i.e. enabled by a HAL). The likelihood of this interfering with anything else is close to zero, but not quite zero. Should we `critical-section` it?

Finomnis: You 100% sure? If so, yes, we should definitely. Timers must be 100% correct and should only fail on a hardware error.

datdenkikniet: image

In `cmu_v3`, at least, `clken0` appears to enable other bus clocks as well (sadly the docs don't clarify a lot, but it sure seems that way)

ETA: and the reference manual appears to support this conclusion: a single 32-bit register with ~31 bits, each enabling a specific (bus) clock.

Finomnis: In that case sure, please create a merge request :)

korken89: Fix added

datdenkikniet (approved):

andresv:
New version of https://crates.io/crates/silabs-metapac is released with @korken89 changes.

PR #1206, opened 2026-07-23
Fix docs.rs for rtic-monotonics

datdenkikniet:
Fix the docs.rs by replacing the features used for that build.

Also added v2.2.1 release in a separate commit (assuming that we want a release here)

Validated that this works by reproducing the failed run from `docs.rs`, and making the same substitution as I've made in this PR on the command line.


korken89 (commented):
Sorry, didn't see this one so I've done the change an already released `v2.2.1` - thanks anyways!

PR #1207, opened 2026-07-24
rtic-macros: Add missing docs to LocalSpawner struct

markhakansson:
Clippy will complain if it's configured to warn/deny on missing docs and one uses local tasks (`local_task=true`).

Also add `[deny(missing_docs)]` on local task example.


korken89 (approved):
Thanks!

PR #1208, opened 2026-07-25
rtic-sync: fix a concurrency bug with `SignalWriter::write_inner` + `WatchReader::try_get`

datdenkikniet:
The flow for this bug is:

1. `SignalWriter::write_inner` writes a value to the `Store`
2. `SignalWriter::write_inner` gets preempted by `WatchReader::try_get`
3. `WatchReader::try_get` reads the value from the `Store`
4. `WatchReader::mark_seen` marks the value as seen
5. `SignalWriter::write_inner` resumes, and marks the value as unseen

<del>I am fairly certain there is another bug lurking in the `changed` function: I feel like there is an inconsistency w.r.t. waking a second time, as the waker access and `seen` accesses are not synchronized in any way. That will land in another PR, though :)</del> Ah, okay: the `changed` function was affected by the exact same bug, but with this change it is also fixed. I've added a test for that as well (though it is largely the same).

This change is important as, without it, one cannot trust that `changed` will _only_ return if a change occurs: due to this bug, it can return twice for a single change, which is counter-intuitive and not what the API promises ("If current value is already seen, it will wait for a new value to be written")

### DIFF
--- a/rtic-sync/CHANGELOG.md
+++ b/rtic-sync/CHANGELOG.md
@@ -12,6 +12,7 @@ For each category, _Added_, _Changed_, _Fixed_ add new entries at the top!
 - A `Channel<T>` is safely `Send` and `Sync` when `T: Send`.
 - An `Arbiter<T>` is safely `Send` and `Sync` when `T: Send`.
 - `Signal<T>` and `Watch<T>` are safely `Send` and `Sync` when `T: Send`.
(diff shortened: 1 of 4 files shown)


datdenkikniet:
Sorry, poking around with github stacked PRs and thought it might not be working due to this PR existing, but that is not the case. This PR is still relevant :)

korken89 (approved):
Thanks!

PR #1210, opened 2026-07-25
rtic-sync: watch: add `reader` and `writer` functions

datdenkikniet:
In some instances, it can be very useful to have (`const`) access to the reader and writer halves of a `Watch` without being able to call `split`. Add functions to get them.

It is hard to use these functions truly correctly, but the alternative (having to create & track an `AtomicWaker` & associated state correctly instead) isn't any better.

As an example, using a `Watch` channel to detect whether an interrupt handler has been called, and doing some work on a lower priority as result of that interrupt.

```rust
static WATCH: Watch<()> = Watch::new();

fn InterruptHandler() {
    WATCH.writer().write(());
} 

// NB: only one instance of this function can be alive
// at any one time.
async fn on_interrupt() {
    let reader = WATCH.reader();
    loop {
        reader.changed().await;
    }
}
```

### DIFF
--- a/rtic-sync/CHANGELOG.md
+++ b/rtic-sync/CHANGELOG.md
@@ -14,6 +14,10 @@ For each category, _Added_, _Changed_, _Fixed_ add new entries at the top!
 - `Signal<T>` and `Watch<T>` are safely `Send` and `Sync` when `T: Send`.
 - Fixed a bug where, in an unlikely edge-case, `WatchReader::changed` could return twice for a single `WatchWriter::write`.
 
+### Added
+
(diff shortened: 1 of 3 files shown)


korken89 (commented):
Looks good, can you add a CHANGELOG entry as well?

datdenkikniet:
During my run I realized that we can do a little better here: instead of `unsafe fn reader()`, we can/should use the `already_split` bool as a `reader_alive` instead. That way, we can have `fn reader()` that copies the current `split()` behaviour. Additionally, we could allow `reader()` or `split()` to be called again if the `reader` is dropped.

I'm not sure if that will be a good change but want to experiment a little more before we merge it :D

And yes, I will add a changelog entry

datdenkikniet:
There! A bit bigger, but IMO a lot better

korken89 (approved):
Nice!

PR #1211, opened 2026-07-25
Remove testing features

datdenkikniet:
There were a bunch of testing features that really don't add anything: migrate them to `dev-dependencies`, or just turn them on by default.

Once more a technically breaking change: removing a feature from `rtic` (even though it was never intended to be used)

### DIFF
--- a/book/en/src/by-example.md
+++ b/book/en/src/by-example.md
@@ -38,13 +38,13 @@ Yields this output:
 ```console
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/xtask qemu --example locals`
-INFO  xtask > Testing for platform: Lm3s6965, backend: Thumbv7
-INFO  xtask::run > 👟 Build example locals (thumbv7m-none-eabi, release, "test-critical-section,thumbv7-backend", in examples/lm3s6965)
+INFO  xtask::run > QEMU run for platform: Lm3s6965, backend: Thumbv7
+INFO  xtask::run > 👟 Build example locals (thumbv7m-none-eabi, release, "thumbv7-backend", in examples/lm3s6965)
 INFO  xtask::run > ✅ Success.
-INFO  xtask::run > 👟 Run example locals in QEMU (thumbv7m-none-eabi, release, "test-critical-section,thumbv7-backend", in examples/lm3s6965)
+INFO  xtask::run > 👟 Run example locals in QEMU (thumbv7m-none-eabi, release, "thumbv7-backend", in examples/lm3s6965)
 INFO  xtask::run > ✅ Success.
-INFO  xtask::results > ✅ Success: Build example locals (thumbv7m-none-eabi, release, "test-critical-section,thumbv7-backend", in examples/lm3s6965)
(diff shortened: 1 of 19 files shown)


korken89 (changes requested):
This LGTM, I ran automated review on it as well that flagged the following:

1. One thing that's now broken (and was not broken by luck before) is the RISC-V tests, we should add it to CI as well.

`cargo xtask test --backend riscv32-imc-clint / riscv32-imc-mecall / riscv-esp32-c3` for testing.

2. `examples/esp32c6` lockfile is stale.

3. The PR's stated philosophy is "move it to dev-dependencies or turn it on by default," and lm3s6965's cortex-m/critical-section-single-core does exactly that. But portable-atomic/critical-section is still injected from xtask, which has two costs:

- `examples/lm3s6965/Cargo.lock` drops the `portable-atomic -> critical-section` edge, yet a thumbv6 example build re-adds it - the committed lock never matches post-build state.
- cd `examples/lm3s6965 && cargo build --features thumbv6-backend` no longer reproduces what CI builds.

Both go away if the example expresses it itself:

```
# examples/lm3s6965/Cargo.toml
thumbv6-backend = ["rtic/thumbv6-backend", "portable-atomic/critical-section"]
```

Same shape for `riscv-esp32c3-backend` in `esp32c3`. `to_rtic_features()` then collapses back to a plain backend name, and the host-test bug above disappears as a side effect.

datdenkikniet:
> One thing that's now broken (and was not broken by luck before) is the RISC-V tests, we should add it to CI as well.
>
> `cargo xtask test --backend riscv32-imc-clint / riscv32-imc-mecall / riscv-esp32-c3 for testing.`

I can't get this command to work on `master` either: `xtask` complains that lm3s6965 doesn't support this platform. Running with a supported platform (`hifive1`) also doesn't pass on either branch (albeit for different reasons). Won't fix here, let's fix complaints about non-working commands in #1213.

> examples/esp32c6 lockfile is stale.

Fixed. I'll introduce `--locked` to the CI at some point, this seems like a  thing that should never happen.

> The PR's stated philosophy  .. Both go away if the example expresses it itself:
>
> \# examples/lm3s6965/Cargo.toml
> thumbv6-backend = ["rtic/thumbv6-backend", "portable-atomic/critical-section"]

<del>Fixed.</del>

> Same shape for riscv-esp32c3-backend in esp32c3. to_rtic_features()

<del>Fixed.</del> OK nevermind, it blew up. The `to_rtic_features()` thing cannot be simplified in the current state. Won't fix here.

datdenkikniet:
> The PR's stated philosophy .. Both go away if the example expresses it itself:

OK, re-fixed this now: adding the features in the respective `Cargo.toml`s makes the examples compile on their own, but it cannot be removed from `to_rtic_features()` (yet) because that is used both to activate the features for the `rtic` package and to activate the features for examples that use a specific backend.

Fixing that is out of scope for this PR, as it doesn't require that we add/remove features to/from `rtic`, so won't fix. Will resolve in #1213

datdenkikniet:
(and to clarify: we're getting rid of all of the feature shenanigans going on in the examples in #1213 :) )

korken89 (approved):

PR #1212, opened 2026-07-25
xtask: don't try to set rust flags for examples

datdenkikniet:
I don't know how this became such a mess, but important details are splling left right and center. I'm going to publish a bunch of changes (hopefully a stack!) that tries to fix `xtask` so that it becomes less of a "must know it all" builder.

There is now (and there has been, for a while) a better way to set linker stuff. `xtask` should not have to care about it, so remove it all and move it into build scripts.

The only thing that actually changes it that some examples are now no longer linked with `flip-link`, which is a non-problem IMO.


korken89 (dismissed):
LGTM

korken89 (approved):

PR #1213, opened 2026-07-25
xtask: try to untangle the mess

datdenkikniet:
I don't know how this happened (though I definitely contriuted to it :P), but `xtasks` has become a terrible mess and is practically impossible to maintain: everything is everywhere, which _really_ shouldn't be the case.

Main improvements:

* Commands & arguments are now fully separate for packages and examples
* Commands run "all sub-parts" by default. Arguments can only reduce the amount of things that are ran, not increase
* `cargo xtask ci` is not as much of a behemoth
* Reworked `xtask` to have next to no global arguments
* Reworked all `xtask` code so that example-commands take a `Platform` and derive their target + backend + example files from that.
* Remove `rayon` support from `xtask`: it's cool, but kind of useless (due to cargo blocking on builds). We can always add it back later
* Stop running both `cargo check` and `cargo clippy`, just run `cargo clippy`
* Make it simple to rework `xtask fmt` just format all `Cargo.toml`s instead of only specific packages. Not executed in this PR to keep the file noise low.
* All examples now build by default, and only have extra features/arguments if there is no sane default to pick

There's still a lot to be done, but now all parts of `xtask` are sufficiently modularized that one can actually make refactoring changes that _don't_ affect the whole thing.

This is one big commit because it required many, small changes that affected lots of separate places, and create lots of conflicts between those commits (i.e. update a place two times in a row, only for those updates to be removed entirely in a later commit).


datdenkikniet:
Damn, `cargo xtask test rtic -b thumbv7` broke with `cortex-m v0.7,8` because they're now gating the `basepri` and `basepri_max` registers behind the `--target` (which seems correct, but is very unpleasant for our `xtasks` to deal with). This means that we cannot `export` those registers while building for the host tuple, but I _think_ we rely on them for the UI tests (that _have_ to be built for the host tuple? Don't think we can tell `trybuild` which target to build for)... Welp, this is postponed another week (but I also think this breaks _all_ CI runs, not just the ones in this PR)

PR #1218, opened 2026-07-30
book: document local tasks

markhakansson:
The concept of `local_task`s was not documented in the book. Add a short section at the end of the software task chapter, describing this feature.

### DIFF
--- a/book/en/src/by-example/software_tasks.md
+++ b/book/en/src/by-example/software_tasks.md
@@ -113,3 +113,15 @@ $ cargo xtask qemu --verbose --example zero-prio-task
 ---
 
 Application side safety: Technically, the RTIC framework ensures that `poll` is never executed on any _software_ task with _completed_ future, thus adhering to the soundness rules of async Rust.
+
+## Local tasks
+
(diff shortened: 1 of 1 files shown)


book/en/src/by-example/software_tasks.md
```diff
@@ -113,3 +113,17 @@ $ cargo xtask qemu --verbose --example zero-prio-task
 ---
 
 Application side safety: Technically, the RTIC framework ensures that `poll` is never executed on any _software_ task with _completed_ future, thus adhering to the soundness rules of async Rust.
+
+## Local tasks
+
+For situations where running tasks at the lowest priority is not an option, you may use local tasks.
```
datdenkikniet: I think this narrows the use-case too much, i.e. you can use this whenever, not just if you cannot run at the lowest priority. I think we could rephrase (and hoist the `!Send` information):

```
Another method of spawning tasks that require non-`Send` arguments are local tasks.
```

markhakansson: Updated with suggestion!

AfoHT (approved):
Nice :)

PR #1223, opened 2026-08-03
Fix bugs in local tasks

### LINKED ISSUE #1222: mut argument syntax in local tasks not supported
Is syntax like this supported?

```rust
    #[task(local = [msg_channel_sender_acm, acm_pipe_reader], priority = 1, local_task = true)]
    async fn usb_acm_task(mut ctx: usb_acm_task::Context, mut usb_acm: CdcAcmClass) {
```

the compiler complains about the `mut` keyword before the argument name.

I can solve the issue with

```rust
    #[task(local = [msg_channel_sender_acm, acm_pipe_reader], priority = 1, local_task = true)]
    async fn usb_acm_task(mut ctx: usb_acm_task::Context, mut usb_acm: CdcAcmClass) {
        let mut usb_acm = usb_acm;
```
but nore sure if that is 0 cost. It would be nice if the syntax above is supported.

datdenkikniet:
Fix two bugs in local tasks:

1. `local_task` was forced to be the last argument to the `#[task]`
2. `local_task` arguments cannot "be fancy" (i.e. be anything other than a plain `ident`)

For #2 I have no clue if I'm doing it right: it's a bit hard to figure out what `regroup_inputs` is supposed to do, but I've basically copied what the non-`local_task` spawn does and it appears to work.

If merged, fixes #1222

### DIFF
--- a/rtic-macros/src/syntax/parse.rs
+++ b/rtic-macros/src/syntax/parse.rs
@@ -208,10 +208,6 @@ fn task_args(tokens: TokenStream2) -> parse::Result<Either<HardwareTaskArgs, Sof
             let ident: Ident = input.parse()?;
             let ident_s = ident.to_string();
 
-            // Handle equal sign
-            let eq = input.parse::<Token![=]>();
-
-            // Only local_task supports omitting the value
             if &*ident_s == "local_task" {
                 if local_task.is_some() {
                     return Err(parse::Error::new(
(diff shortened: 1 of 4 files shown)


rtic-macros/src/syntax/parse.rs
```diff
@@ -208,10 +208,6 @@ fn task_args(tokens: TokenStream2) -> parse::Result<Either<HardwareTaskArgs, Sof
             let ident: Ident = input.parse()?;
             let ident_s = ident.to_string();
 
-            // Handle equal sign
-            let eq = input.parse::<Token![=]>();
-
-            // Only local_task supports omitting the value
             if &*ident_s == "local_task" {
                 if local_task.is_some() {
```
datdenkikniet, line 212: This looks like a big change, but I just put what was previously the fall-through case in the `else` block so that both cases use the same comma parsing logic

korken89 (approved):
Thanks for fixing!

PR #1226, opened 2026-08-06
Fix #1225

### LINKED ISSUE #1225: `extern "Rust"` tasks can take `Context<'static>` despite being non-diverging, allowing for aliasing mutable access
In this example:

```rust
#![no_main]
#![no_std]

use core::task::Poll;
use panic_semihosting as _;

use cortex_m_semihosting::{debug, hprintln};

#[rtic::app(device = lm3s6965, dispatchers = [SSI0])]
mod app {
    use super::*;

    #[shared]
    struct Shared {}

    #[local]
    struct Local {}

    #[init]
    fn init(_: init::Context) -> (Shared, Local) {
        ping::spawn().ok();

        (Shared {}, Local {})
    }

    extern "Rust" {
        #[task(local =  [ value: NotCopy = NotCopy { value: 2 }, counter: u32 = 0] )]
        async fn ping(c: ping::Context<'a>);
    }

    #[task]
    async fn pong(_c: pong::Context) {
        hprintln!("Pong");
        ping::spawn().ok();
    }

    #[task]
    async fn holding_forever_mut(
        _c: holding_forever_mut::Context<'_>,
        value: &'static mut NotCopy,
    ) {
        hprintln!("Holding forever mut, {:?}", value);
        core::future::poll_fn::<(), _>(|_| Poll::Pending).await;
        hprintln!("Exiting holding_forever_mut");
    }

    #[task]
    async fn holding_forever_mut2(
        _c: holding_forever_mut2::Context<'_>,
        value: &'static mut NotCopy,
    ) {
        hprintln!("Holding forever mut2. {:?}", value);
        core::future::poll_fn::<(), _>(|_| Poll::Pending).await;
        hprintln!("Exiting holding_forever_mut2");
    }
}

use app::*;

#[derive(Debug)]
struct NotCopy {
    #[allow(dead_code)]
    pub value: u32,
}

async fn ping(c: ping::Context<'static>) {
    hprintln!("Ping");

    match holding_forever_mut::spawn(c.local.value) {
        Ok(()) => {}
        Err(v) => {
            holding_forever_mut2::spawn(v).ok();
        }
    }

    *c.local.counter += 1;

    if *c.local.counter == 3 {
        debug::exit(debug::EXIT_SUCCESS); // Exit QEMU simulator
    }

    pong::spawn().ok();
}
```

both `holding_forever_mut` and `holding_forever_mut2` hold a mutable reference to `c.local.value`. This is UB.

Without the `extern "Rust"`, the code is transformed in such a way that the context for `ping` must be non-`'static`. However, with the `extern "Rust"` task, this doesn't happen, and we internally call `ping(unsafe { Context::new })`, which is unsound if `ping` does not return `!` (as it is the same reference, twice).

Using a non-`'static` lifetime for the `Context` is OK, but we never restrict it, which is a problem.

datdenkikniet:
For this to function correctly, we must 

Assert that `extern` software tasks are one of two things:

* An `async fn` with a `Context<'a>` where `'a != 'static`
* An `async fn` that returns `!`

and

Assert that `extern` hardware tasks are an `fn` with a `Context<'a>` where `'a != 'static`.

All other tasks (not `extern` or non-divergent) are left along.

This PR adds two layers of validation: in `extern "Rust"` blocks, checks that the signature is correct (i.e. non-divergent tasks must not have `Context<'static>`), and for spawning code of `extern` tasks it validates that the lifetime of the to-be-spawned-future is non-`static`, or that the returned future is divergent.

Checking that a future is divergent is done using an `Into<Infallible>` hack. The DX isn't great, but there doesn't seem to be any other way to name the `!` requirement.

Sadly it doesn't seem like we can test much of this fully: the testing infra (= UI tests) doesn't support `extern "Rust"` tasks.

Fixes #1225 

A test example that showcases all the errors:

<details>
<summary>An example for lm3s6965 showing the caught cases</summary>

```rust
#![no_main]
#![no_std]

use panic_semihosting as _;

#[rtic::app(device = lm3s6965)]
mod app {
    use super::*;

    #[shared]
    struct Shared {}

    #[local]
    struct Local {}

    #[init]
    fn init(_: init::Context) -> (Shared, Local) {
        todo!()
    }

    extern "Rust" {
        // Error cases
        // These two we can check in the macro
        #[task]
        async fn non_diverging_static(_: non_diverging_static::Context<'static>);

        #[task(binds = GPIOA)]
        fn hw_task_static(_: hw_task::Context<'static>);

        // These three can only be checked by the compiler
        #[task]
        async fn extern_doesnt_diverge(_: extern_doesnt_diverge::Context<'_>) -> !;

        #[task]
        async fn extern_requires_static(_: extern_requires_static::Context<'_>);

        #[task(binds = GPIOA)]
        fn hw_task_extern_static(_: hw_task_extern_static::Context);

        // Accepted cases
        #[task]
        async fn non_diverging_non_static(_: non_diverging_non_static::Context<'_>);

        #[task]
        async fn non_diverging_no_lt(_: non_diverging_no_lt::Context);

        #[task]
        async fn diverging_static(_: diverging_static::Context<'static>) -> !;

        // NOTE: we don't (and don't need to) validate whether the actual function
        // definition uses <'static>.
        #[task]
        async fn diverging_non_static(_: diverging_non_static::Context<'_>) -> !;

        // NOTE: we don't (and don't need to) validate whether the actual function
        // definition uses <'static>.
        #[task]
        async fn diverging_no_lt(_: diverging_no_lt::Context) -> !;
    }
}

use app::*;

fn hw_task_extern_static(_: hw_task_extern_static::Context<'static>) {}

async fn extern_doesnt_diverge(_: extern_doesnt_diverge::Context<'_>) {
    loop {}
}

async fn extern_requires_static(_: extern_requires_static::Context<'static>) {}

async fn non_diverging_non_static(_: non_diverging_non_static::Context<'_>) {}

async fn non_diverging_no_lt(_: non_diverging_no_lt::Context<'_>) {}

async fn diverging_static(_: diverging_static::Context<'static>) -> ! {
    loop {}
}

async fn diverging_non_static(_: diverging_non_static::Context<'_>) -> ! {
    loop {}
}

async fn diverging_no_lt(_: diverging_no_lt::Context<'_>) -> ! {
    loop {}
}

```

</details>

### DIFF
--- a/rtic-macros/CHANGELOG.md
+++ b/rtic-macros/CHANGELOG.md
@@ -15,6 +15,10 @@ For each category, *Added*, *Changed*, *Fixed* add new entries at the top!
 - `local_task` arguments can be patterns instead of just an ident.
 - The `local_task = true` argument to `#[task]` must no longer be the last argument.
 
+### Fixed
+
+- Previously, non-diverging `extern "Rust"` tasks could have a `Context` with a `<'static>` lifetime. This is no longer allowed.
+
 ## [v2.3.0] - 2026-07-08
 
 ### Fixed
--- a/rtic-macros/src/codegen/hardware_tasks.rs
+++ b/rtic-macros/src/codegen/hardware_tasks.rs
@@ -8,6 +8,7 @@ use crate::{
 };
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
+use syn::Lifetime;
 
 /// Generate support code for hardware tasks (`#[exception]`s and `#[interrupt]`s)
 pub fn codegen(app: &App, analysis: &Analysis) -> TokenStream2 {
@@ -24,6 +25,8 @@ pub fn codegen(app: &App, analysis: &Analysis) -> TokenStream2 {
         let exit_stmts = interrupt_exit(app, analysis);
         let config = handler_config(app, analysis, symbol.clone());
 
+        let lifetime = Lifetime::new("'non_static", name.span());
+
         mod_app.push(quote!(
             #[allow(non_snake_case)]
             #[no_mangle]
@@ -35,11 +38,12 @@ pub fn codegen(app: &App, analysis: &Analysis) -> TokenStream2 {
 
                 const PRIORITY: u8 = #priority;
 
-                rtic::export::run(PRIORITY, || {
-                    #name(
-                        #name::Context::new()
-                    )
-                });
+                fn exec<#lifetime>() {
+                    let ctx = unsafe { #name::Context::<#lifetime>::new() };
+                    #name(ctx)
+                }
+
+                rtic::export::run(PRIORITY, exec);
 
                 #(#exit_stmts)*
             }
--- a/rtic-macros/src/codegen/module.rs
+++ b/rtic-macros/src/codegen/module.rs
@@ -145,8 +145,18 @@ pub fn codegen(ctxt: Context, app: &App, analysis: &Analysis) -> TokenStream2 {
 
         let lifetime = Lifetime::new("'non_static", name.span());
 
-        let spawn = if app.software_tasks[t].is_bottom || !app.software_tasks[t].is_extern {
-            quote! { exec.spawn(#name(unsafe { #name::Context::new() } #(,#input_untupled)*)); }
+        let task = &app.software_tasks[t];
+        let spawn = if !task.is_extern {
+            quote! {
+                let future = #name(unsafe { #name::Context::new() } #(,#input_untupled)*);
+                exec.spawn(future);
+            }
+        } else if task.is_bottom {
+            quote! {
+                let future = #name(unsafe { #name::Context::new() } #(,#input_untupled)*);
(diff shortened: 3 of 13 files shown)


datdenkikniet:
Ah, crap, there are probably more use-cases that I am missing: `idle` also needs it's share of "ah whatever" allowance... What a pain

Edit: no, nevermind! `idle` must be diverging, so it doesn't suffer from this problem and is not affected by the code change.

datdenkikniet:
No, this is still a draft. I need to double check that non-foreign tasks that do specify a static lifetime still compile (this was always allowed. The internal tasks' call signature is transformed to be non-static. This can't be done for externally defined tasks, which is the whole reason behind #1225 existing)

datdenkikniet:
Okay, now then. Only add checks to paths that are `extern` and not `!`. This should not break any legitimate usage that worked before.

datdenkikniet:
Have also included #1224 to show that the CI is clean

korken89 (approved):

PR #1235, opened 2026-08-20
Store each async task's executor in a `static`

### LINKED ISSUE #1234: Async task executors live in `main`'s stack frame
Every async software task gets an `AsyncTaskExecutor<F>`, and `#[app]` puts it in `main`'s frame as
a `ManuallyDrop` local. Each one is published to a `static AsyncTaskExecutorPtr` so the dispatchers
and the spawn path can reach it.

The executors are not in `.bss`, so `size` and `nm` do not account for them. They count against the
stack instead. `init` covers that at run time: it reads the MSP after allocating, and panics if the
stack has crossed `__ebss`.

That check costs more than the few instructions it compiles to. It makes `panic!` reachable, so an
application that never panics elsewhere links the panic handler and `core::fmt` for it.

Every poll and every wake goes through an `AtomicPtr`. Nothing can fold or alias-analyse across that
load.

### Proposal

Give each task a `static` of its own. The future's type still cannot be named, so declare the
storage as bytes. Two `const fn`s take the size and the alignment from the task function itself.
They can do that because a trait binds that function's future as an associated type. The image is
all zero, so it lands in `.bss` and costs no initialiser.

`embassy-executor` uses this mechanism for its task pools on stable: a `TaskFn` trait, a
`task_pool_size`/`task_pool_align` pair, and a byte holder.

Two things then have nothing left to do. The stack-overflow check has no stack allocation to compare
against, and `init::Context::executors_size` has no size to report. The 17
`new_N_args`/`from_ptr_N_args` pairs also collapse into 17 impls of one trait, at the same
16-parameter cap.

A `static` also stops v2.1.3 from recurring. That fix added `+ 'static` to the executor's future. A
`static` cannot hold a non-`'static` type, so the bound comes for free.

CordlessCoder:
### What changes

Each async software task's executor moves out of `main`'s frame and into a `static` of its own.

The future's type still cannot be written down, so the storage is declared as bytes: an `ExecutorHolder<SIZE, ALIGN>`.
Two `const fn`s get the size and the alignment from the task function itself. That works because `ExecFn` binds the function's future as an associated type, which `size_of` and `align_of` then name. The holder's image is all zero, so it lands in `.bss` and costs no initialiser.

Three things go with the allocation:
- `check_stack_overflow_before_init`, which checked it. Backends implement this function, so removing it changes the backend API.
- `init::Context::executors_size`, which reported its size.
- `AsyncTaskExecutor::new_N_args` and `from_ptr_N_args`. 17 pairs, replaced by 17 impls of `ExecFn` at the same 16-parameter cap.

### Why

The executors were not in `.bss`, so `size` and `nm` did not account for them. They counted against
the stack instead. That is what the MSP check in `init` was for: read the stack pointer after
allocating, and panic if it has crossed `__ebss`.

Every poll and every wake also loaded an `AtomicPtr` and dereferenced it, inhibiting optimizations.

A `static` also stops v2.1.3 from recurring. That fix added `+ 'static` to the executor's future.
A `static` cannot hold a non-`'static` type, so the bound comes for free.

### Measured

lm3s6965 examples, `--release`, `thumbv7-backend`. 55 each side, 54 comparable: `executor-size` only exists before, `static-exec` only after.

RAM moves out of the frame and into `.bss`, and comes out slightly smaller. `main`'s stack change against the `.bss` delta:

| example | `main` frame | `.bss` | net |
|---|---|---|---|
| `async-channel` | 336 → 12 | +312 | −12 |
| `async-timeout` | 176 → 0 | +168 | −8 |
| `async-task-multiple-prios` | 32 → 0 | +12 | −20 |
| `common` | 32 → 8 | +20 | −4 |

The saving is small. What matters is that those 312 bytes are now in `.bss`, where `size` and a linker assert can see them.

37 examples land between −200 and +80 bytes of `.text`, median −100.
One grows: `async-channel`, by 80. All 80 are in the zero-priority dispatcher, and `main` loses 102.

17 examples lose more than 400 bytes, and 11 of those lose about 3500. It
removes their only reachable `panic!`, so dropping it drops the panic handler and `core::fmt` too.
`init` goes from 13 formatting and panic symbols to none.

`examples/lm3s6965/Cargo.lock` also carries the `2.3.0` → `2.3.1` bump the release commit missed.
Building any example at master dirties it the same way.

### Breaking

- `init::Context::executors_size` is gone.
- `check_stack_overflow_before_init` is gone from the backend API. An out-of-tree backend that implements it needs the function removed.
- `AsyncTaskExecutorPtr`, `new_N_args` and `from_ptr_N_args` are gone from `rtic::export::executor` (`#[doc(hidden)]`).

### Testing

`cargo xtask` fmt, check, clippy, test, qemu and example-check all pass.

`examples/executor-size` measured the stack allocation, so it goes with it. `examples/static-exec` replaces it, and covers:
- a task with no arguments, and one with several
- two priorities, so more than one dispatcher polls a static
- the three ways the macro builds a future: inline, `extern` and diverging, and `extern` borrowing its context.

Closes #1234


korken89 (changes requested):
Thanks for the PR, I did a quick scan and found the following issues:

- Dead msp/read_sp exports (I think these should be removed now)
- The MSP-check we did would catch custom `_stack_start`, now I think this is lost
- Try to not use unicode in the tests, em-dash etc
- Changelog has wrong format (new entries under old releases???)
- `exec_from_holder`'s const generics are unconstrained, upheld by construction now which is correct but is easy to break. I would add something like to make sure we don't break this in the future:
```
const {
    assert!(SIZE == size_of::<AsyncTaskExecutor<Fut>>());
    assert!(ALIGN == align_of::<AsyncTaskExecutor<Fut>>());
}
```

This looks like it may have been generated by AI (sorry if not), please do give it an extra pair of eyes if so. :D

Great thing to fix, thanks for putting the effort!

CordlessCoder:
Thanks for the super detailed response, a rebase moved the changelog entries down 🤦🏻‍♂️ 

I'll try to implement your suggested changes today.

CordlessCoder:
The clippy warning seems to be unrelated to my changes - probably because of the 1.98 update.

I removed msp/read_sp, the em-dash from that test, and added the assertion you suggested.

I'm not sure if the `_stack_start` check is worth it, cortex-m-rt 0.7.5 already adds `_stack_end` and `ASSERT(_stack_start >= _stack_end)` in link.x, and an always-reachable panic! can cost quite a bit of flash.
I can add it back if you think it's worth it.

korken89:
Ah that's nice, then skip the `_stack_start`. Not sure on the `clippy` issue ... but we've not upgraded the compiler so I'm not sure.

CordlessCoder:
New lint fixed in #1237

korken89:
Great find! I've added it to the merge queue.

korken89 (approved):
Thanks for improving!

PR #1236, opened 2026-08-20
Fugit 0.5.0

Finomnis:
This is an experimental integration of fugit 0.5.0.

Be aware that fugit is on its way to 1.0.0 and this is one of the last pre-release versions. 

This means we can merge, but we should not release until fugit 1.0.0.

### DIFF
--- a/ci/expected/lm3s6965/async-timeout.run
+++ b/ci/expected/lm3s6965/async-timeout.run
@@ -5,12 +5,12 @@ the hal takes a duration of Duration { ticks: 45 }
 hal returned 5
 the hal takes a duration of Duration { ticks: 45 }
 hal returned 5
-now is Instant { ticks: 213 }, timeout at Instant { ticks: 263 }
+now is MonotonicInstant { ticks: 213 }, timeout at MonotonicInstant { ticks: 263 }
 the hal takes a duration of Duration { ticks: 35 }
-hal returned 5 at time Instant { ticks: 249 }
-now is Instant { ticks: 313 }, timeout at Instant { ticks: 363 }
+hal returned 5 at time MonotonicInstant { ticks: 249 }
+now is MonotonicInstant { ticks: 313 }, timeout at MonotonicInstant { ticks: 363 }
 the hal takes a duration of Duration { ticks: 45 }
-hal returned 5 at time Instant { ticks: 359 }
-now is Instant { ticks: 413 }, timeout at Instant { ticks: 463 }
+hal returned 5 at time MonotonicInstant { ticks: 359 }
+now is MonotonicInstant { ticks: 413 }, timeout at MonotonicInstant { ticks: 463 }
 the hal takes a duration of Duration { ticks: 55 }
 timeout
--- a/examples/nrf52840_blinky/Cargo.toml
+++ b/examples/nrf52840_blinky/Cargo.toml
@@ -17,7 +17,7 @@ defmt-rtt = "0.4"
 panic-probe = { version = "0.3", features = ["print-defmt"] }
 cortex-m-semihosting = "0.5.0"
 nrf52840-hal = "0.16.0"
-fugit = { version = "0.3.7", features = ["defmt"] }
+fugit = { version = "0.4.0", features = ["defmt"] }
 
 [dependencies.rtic]
 path = "../../rtic"
--- a/examples/nrf52840_blinky/Cargo.toml
+++ b/examples/nrf52840_blinky/Cargo.toml
@@ -17,7 +17,7 @@ defmt-rtt = "0.4"
 panic-probe = { version = "0.3", features = ["print-defmt"] }
 cortex-m-semihosting = "0.5.0"
 nrf52840-hal = "0.16.0"
-fugit = { version = "0.4.0", features = ["defmt"] }
+fugit = { version = "0.5.0", features = ["defmt"] }
 
 [dependencies.rtic]
 path = "../../rtic"
--- a/examples/rp2040_local_i2c_init/Cargo.toml
+++ b/examples/rp2040_local_i2c_init/Cargo.toml
@@ -21,7 +21,7 @@ features = ["rp2040"]
 [dependencies]
 cortex-m = "0.7"
 embedded-hal = { version = "0.2.7", features = ["unproven"] }
-fugit = "0.3"
+fugit = "0.4.0"
 rp-pico = "0.9.0"
 panic-probe = "0.3"
 portable-atomic = { version = "1", features = ["critical-section"] }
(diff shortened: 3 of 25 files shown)


Finomnis:
I'd like @korken89's review :)

korken89:
Nice, overall I think it looks good!

One thing I'm wondering is if we want to relax `type Instant` in `rtic-time/src/lib.rs` from `Ord` to `PartialOrd` as you want to be able to do scheduling based of wrapping timers as well.

What's your take?

Finomnis:
> One thing I'm wondering is if we want to relax `type Instant` in `rtic-time/src/lib.rs` from `Ord` to `PartialOrd` as you want to be able to do scheduling based of wrapping timers as well.

@korken89 are you sure? In what scenario would you want that? Scheduling in a wrapping timeline sounds very confusing...

Finomnis:
@korken89 I understand that you intentionally break the PartialOrd contact with wrapping monotonics to get the comparison operators for convenience, which to be honest I accept but I'm not a fan of. If it were my library I would probably not impl PartialOrd for the wrapping instant.

We shouldn't actually use it in an algorithm that was originally written for non wrapping timelines... Unless I'm missing something here

Finomnis:
@korken89 actually, after reading through [`TimerQueueTicks`](https://docs.rs/rtic-time/latest/rtic_time/timer_queue/trait.TimerQueueTicks.html) I guess it is written with wrapping timelines in mind. I still don't like that we use `PartialOrd` for this, it's just simply the wrong trait for wrapping timers. Maybe we should create a new trait for this purpose.

Finomnis:
Actually, the situation is more complicated. There are [`TimerQueueBasedInstant`](https://docs.rs/rtic-time/latest/rtic_time/monotonic/trait.TimerQueueBasedInstant.html) for the instants that a time queue delivers, and then there is [`Monotonic::Instant`](https://docs.rs/rtic-time/latest/rtic_time/trait.Monotonic.html#associatedtype.Instant).

For the `Monotonic::Instant` I would strongly argue that we need to keep `Ord` here, otherwise libraries that use monotonics generically can no longer rely on `Ord`, which seems against the spirit of monotonics in general. This in turn dictates that `TimerQueueBasedInstant` is also `Ord`, otherwise it would not be compatible with `Monotonic`. The whole reason why we introduced `MonotonicInstant` in `fugit` was to make it `Ord`, so removing from `Monotonic` is counterproductive imo.

If you are talking about the [`TimerQueueTicks`](https://docs.rs/rtic-time/latest/rtic_time/timer_queue/trait.TimerQueueTicks.html), that one has no `Ord` requirement and is already compatible with wrapping timers. So no change is required there.

korken89:
This comes mostly from the history of RTIC. When we made the initial timer support (in the C days) we believed extending timers was too complicated for users, so we made sure scheduling worked with and without an extended timer - fugit was born to make this less error prone in Rust.

Today I do agree a wrapping timeline is mostly confusing, but one use-case from back then remains: systems that go to sleep without a fully alive timer (we have had products using this as well). There the interrupt driven extension to 64 bits stops working, and such a system can still do correct relative scheduling when it awake (which is what `WrappingInstant` models).

And I agree on `Monotonic::Instant`, generic code should be able to keep relying on `Ord` no doubt about but I think we can have both by splitting the trait:

```
/// All of today's methods move here.
pub trait WrappingMonotonic {
    type Instant: PartialOrd + Copy + ...;
    ...
}

/// What `Monotonic` means today.
pub trait Monotonic: WrappingMonotonic<Instant: Ord> {}
impl<T: WrappingMonotonic<Instant: Ord>> Monotonic for T {}
```

With this `M: Monotonic` bounds compile unchanged and keep `Ord`, while a sleepy system implements only the base trait with a `WrappingTimerInstant` for it's custom timer. So when you make your monotonic impl you choose if it's a wrapping or monotonic version.

What do you think about this? Then we don't close the door on "sleepy systems".
This does not need to block this PR, I can take it as a follow-up if you feel like it's too much work. :) And sorry for the wall of text. xD

Finomnis:
I agree that if there is a valid usecase we should support it.

I think `WrappingMonotonic` is a misnomer, either something is a monotonic or wrapping.

And I think using `PartialOrd` for this is dangerous. You said you want `PartialOrd` for the comparison operators, but using it as the actual trait whilst breaking its contract is a very dangerous path imo.

We should introduce an `InstantOrd` trait and use that one instead. I personally would also drop `PartialOrd` from `WrappingInstant`, if I'm completely honest, and replace it with `InstantOrd`. I would have `InstantOrd` provide a similar api as `TimerQueueTicks`.

Yes, we would lose the comparison operators for `WrappingInstant`, but I think that's fine.

korken89:
That's a very neat idea! And yeah `WrappingMonotonic` is not great, an issue that came from us just starting to call schedulable timers "monotonics".

Do you want to take a stab at it or should I?

Finomnis:
> Do you want to take a stab at it or should I?

Feel free, I don't have much time right now.

The open question is, if we introduce an `InstantOrd` trait, where should it live? In `rtic-time` or in `fugit`? I guess in`fugit`?

korken89:
I think that's correct, `fugit` should own its comparisons. I'll take a stab at it :D

korken89:
Alright, I've done a stab at `fugit` and `rtic-time`/`rtic-monotonics` - feel free to give feedback!

https://github.com/korken89/fugit/tree/instant-ord
https://github.com/rtic-rs/rtic/tree/timebase-split

Finomnis:
> Alright, I've done a stab at `fugit` and `rtic-time`/`rtic-monotonics` - feel free to give feedback!
> 
> https://github.com/korken89/fugit/tree/instant-ord https://github.com/rtic-rs/rtic/tree/timebase-split

Gotta be honest, I like it :)

Did the few comments I made reach you? Unsure how comments to commits outside of PRs get delivered.

Finomnis:
@korken89 Will you PR it to fugit? Or talk to the rtic group/@AfoHT first? I think it's a good change, I'm for it.

korken89:
Ah sorry, I've been away on a business trip and am back today, I'll give them a look today/tomorrow!

Finomnis:
> Ah sorry, I've been away on a business trip and am back today, I'll give them a look today/tomorrow!

Np, we're all just freelancers :)

korken89:
I've opened them as PRs so we can collect comment:

https://github.com/rtic-rs/rtic/pull/1242
https://github.com/korken89/fugit/pull/89

I can see in emails that you have made comments, but I can't for the life of me find the comments :(

Finomnis:
> I've opened them as PRs so we can collect comment:
> 
> #1242 [korken89/fugit#89](https://github.com/korken89/fugit/pull/89)
> 
> I can see in emails that you have made comments, but I can't for the life of me find the comments :(

You need to watch the commit in side-by-side diff then they are visible. At least on my machine

Finomnis:
@korken89 it was only 2, I copied both of them over, but realized one is wrong anyway

Finomnis:
Superseded by #1242

PR #1237, opened 2026-08-21
Fix needless_late_init Clippy lint

CordlessCoder:
This lint seemed to start triggering with the Rust 1.98.0 update.


korken89 (approved):
Thanks!

PR #1247, opened 2026-09-16
Fix async priority limit to stay above all software tasks

### LINKED ISSUE #1246: Monotonics blocking delay on STM32G4 can hang
Calling a blocking delay on a monotonic timer generated by rtic-monotonics seems to deadlock in some specific cases.
I noticed it happens when there's a task one priority level higher than the task that gets deadlocked.
In the example below setting SPI's prio to 3 and draw's to 1 fixes the issue.

I don't have any guesses as to why, this is the minimal code required to reproduce the issue.

```rust
#![no_main]
#![no_std]

use crate_name as _;

#[rtic::app(
    device = stm32g4xx_hal::pac,
    dispatchers = [USART1]
)]
mod app {
    use rtic_monotonics::stm32::prelude::*;
    use stm32g4xx_hal::{
        prelude::*,
        rcc
    };

    rtic_monotonics::stm32_tim4_monotonic!(Mono, 1_000_000);

    #[shared]
    struct Shared { }
    #[local]
    struct Local { }

    #[init]
    fn init(cx: init::Context) -> (Shared, Local) {
        let pwr = cx.device.PWR
            .constrain()
            .freeze();
        let rcc_conf = rcc::Config::hsi();

        let rcc = cx.device.RCC
            .constrain()
            .freeze(rcc_conf, pwr);

        defmt::info!("Clock init done: {:?}", rcc.clocks);

        defmt::info!("Mono::start({=u32})", rcc.clocks.apb1_tim_clk.raw());
        Mono::start(rcc.clocks.apb1_tim_clk.raw());

        draw::spawn().ok();

        (Shared { }, Local { })
    }

    #[task(priority = 1)]
    async fn draw(_cx: draw::Context) {
        {
            defmt::info!("pre blocking delay");
            Mono.delay_ms(100);
            defmt::info!("pre async delay");
            Mono::delay(1000_u64.micros()).await;
        };

        loop {
            Mono::delay(500u64.millis()).await
        }
    }

    #[task(priority = 2, binds = SPI1)]
    fn spi1(_cx: spi1::Context) {
    }
}
```

This is the stack trace that gets printed when I interrupt probe-rs:
```
Core 0
    Frame 0: Instant<u64, 1, 1000000>::const_cmp<1, 1000000> @ 0x800053e inline
        /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fugit-0.3.9/src/instant.rs:73:20
    Frame 1: partial_cmp<1, 1000000> @ 0x800053e inline
        /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fugit-0.3.9/src/instant.rs:206:27
    Frame 2: lt<fugit::instant::Instant<u64, 1, 1000000>, fugit::instant::Instant<u64, 1, 1000000>> @ 0x800053e inline
        /home/user/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cmp.rs:1411:14
    Frame 3: delay_ms @ 0x800053e inline
        /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rtic-time-2.0.1/src/monotonic/embedded_hal_macros.rs:41:23
    Frame 4: {async_fn#0} @ 0x800052a inline
        /home/user/redacted/src/bin/deadlock.rs:56:18
    Frame 5: AsyncTaskExecutor<deadlock::app::draw::{async_fn_env#0}>::poll @ 0x8000512 inline
        /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rtic-2.3.1/src/export/executor.rs:209:26
    Frame 6: {closure#0} @ 0x80004de inline
        /home/user/redacted/src/bin/deadlock.rs:6:1
    Frame 7: run<deadlock::app::USART1::{closure_env#0}> @ 0x80004d4 inline
        /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rtic-2.3.1/src/export/cortex_basepri.rs:17:9
    Frame 8: USART1 @ 0x80004d4
        /home/user/redacted/src/bin/deadlock.rs:6:1
    Frame 9: External interrupt #37 @ 0x8000903
    Frame 10: __rtic_init_resources<deadlock::app::main::{closure_env#0}> @ 0x8000903
        /home/user/redacted/src/bin/deadlock.rs:9:3
    Frame 11: main @ 0x8000b10
        /home/user/redacted/src/bin/deadlock.rs:6:1
    Frame 12: Reset @ 0x8000212
```

korken89:
RTIC_ASYNC_MAX_LOGICAL_PRIO was one below the lowest hardware task, so a software task right below a hardware task shared the priority of the monotonic interrupt and a blocking delay in it hung (#1246). The limit is now also at least one above the highest software task.


AfoHT (dismissed):
Looks good, hoping that CI will agree (and work!) 🚀 :)

Finomnis:
Technically the issue still exists when hardware tasks run blocking delays...

The example in the bug report uses blocking delays in software tasks, which shouldn't be done usually. But in hardware tasks this is a legit usecase that will break by default.

Or do I misunderstand something?

korken89:
It's an OK usecase, but not supported by monotonics that rely on interrupts to do progress. It's not something that can be won with smarter prios sadly as someone can always make a blocking loop at maximum priority.

techmccat:
I don't quite understand why blocking delays need interrupts to fire (the macro DelayNs impl seems to busy wait on Backend::now() < done), but couldn't the interrupt service routine be handled in the busy-wait loop of the blocking delay instead if the priority can't be set higher?

korken89:
Yeah, we can always do:

```rust
fn now() -> Self::Ticks {
    calculate_now(
        || {
            Self::on_interrupt();
            $overflow.load(Ordering::Relaxed)
        },
        || $timer.cnt().read(),
    )
}
```

To do interrupt handling inside now as a fallback. It will add that to hot paths everywhere though, but that might be worth it?

korken89:
The only issue I see is that it's not a complete fix. If a higher prio context blocks for too long even doing this in `now()` is not a fix.

korken89:
I've added a commit for this to show how it would look

Finomnis:
> Yeah, we can always do:
> 
> ```rust
> fn now() -> Self::Ticks {
>     calculate_now(
>         || {
>             Self::on_interrupt();
>             $overflow.load(Ordering::Relaxed)
>         },
>         || $timer.cnt().read(),
>     )
> }
> ```
> 
> To do interrupt handling inside now as a fallback. It will add that to hot paths everywhere though, but that might be worth it?

Is on_interrupt threadsafe??

EDIT: I just checked stm32's on_interrupt and it is most certainly **not** threadsafe, so please don't do this :)

rtic-monotonics/src/silabs/timer.rs
```diff
@@ -378,7 +378,11 @@ macro_rules! make_silabs_timer {
 
             fn now() -> Self::Ticks {
                 calculate_now(
-                    || $overflow.load(Ordering::Relaxed),
+                    || {
+                        // Catch up on overflow flags so a spin loop keeps time without the interrupt.
+                        Self::on_interrupt();
```
Finomnis, line 383: on_interrupt is not thread safe!!

Finomnis: I get why you did it this way, there's a critical section in there, but that does not make it thread safe!

Nothing prevents the situation where the timer interrupt fired and the interrupt gets processed, and while being processed a blocking wait calls on_interrupt again. Critical sections do not wait until an ongoing interrupt is finished, unless that interrupt uses a critical section itself!

Finomnis: Don't get me wrong, I think this is a good idea in general, but we need to introduce a critical section where on_interrupt was called originally.

korken89: Ops, yeah that went too fast. I'll look more into it during the weekend

Finomnis: Also I'm not sure if we really want to put this inside of calculate_now. Also, you guys keep saying rtic_monotonics are reference implementations, so if we change this, we should instead rework the monotonics backend api to solve this use for all monotonics, ours and external ones.

Finomnis: An alternative would be to not fix it at all, but instead detect it and raise a compilation error. Like, if the priority of the interrupts is automatically computed, raise a compile time error if it overlaps with the hardware tasks or software tasks. Then, force the user to specify the interrupt priority manually and accept whatever the user defines.

Finomnis: @korken89 Also, be aware that rtic might not know about all tasks. For example, one major rtic usecase is embassy-stm32, which registers its own interrupt handlers.

So my suggestion:
- compute the software task max priority
- set the timer interrupt priority to one above the software task priority
- if that's not possible (because software task is at highest possible priority), compiler error
- if timer interrupt priority is higher or equal to the lowest hardware task priority, compiler error telling the user that he should set the interrupt priority manually
- if timer priority is set manually, do not error if it collides with hardware tasks. But still error if it collides with software tasks.

Alternative, which I personally would even prefer: force the timer interrupt priority to be set manually. In my opinion this interrupt is so critical for system stability that it should be an explicit value that the user must understand, it should not be auto generated at all. The user should actively understand that there is a tradeoff between timer stability and hardware task timing reliability.

rtic/CHANGELOG.md
```diff
@@ -27,6 +27,10 @@ Example:
 - Removed `init::Context::executors_size`, which reported a stack allocation that no longer happens.
 - Removed the stack-overflow check `init` ran after allocating executors.
 
+### Fixed
+
+- `RTIC_ASYNC_MAX_LOGICAL_PRIO` is always above the highest software task, so monotonics can preempt a blocking software task (#1246).
+
```
Finomnis, line 33: This is also documented incorrectly in the rtic-monotonics docs, it there says one below the highest hardware task. Even before this change this is wrong.

AfoHT:
> So my suggestion:
> - compute the software task max priority
> - set the timer interrupt priority to one above the software task priority
> - if that's not possible (because software task is at highest possible priority), compiler error
> - if timer interrupt priority is higher or equal to the lowest hardware task priority, compiler error telling the user that he should set the interrupt priority manually
> - if timer priority is set manually, do not error if it collides with hardware tasks. But still error if it collides with software tasks.
>
> Alternative, which I personally would even prefer: force the timer interrupt priority to be set manually. In my opinion this interrupt is so critical for system stability that it should be an explicit value that the user must understand, it should not be auto generated at all. 

That is imho a really good idea, the timer interrupt priority should be visible and controllable.
Without a breaking change though, RTICv2 should default to the best possible "default"; the steps you suggest look solid.

I think it is reasonable we should also enforce "correctness" with compiler errors, even though previously one could compile some creative "not-great for actual use"...

rtic/CHANGELOG.md
korken89: Should be fixed now

rtic-monotonics/src/silabs/timer.rs
korken89: There is now a new API so one is forced to select an explicit value or the RTIC generated default

rtic-macros/src/analyze.rs
```diff
@@ -57,12 +57,22 @@ pub fn app(analysis: analyze::Analysis, app: &App) -> Analysis {
...
-        .iter()
-        .map(|(_, task)| task.args.priority)
+        .values()
+        .map(|task| task.args.priority)
         .min()
-        .map(|v| v - 1); // One less than the smallest HW task
+        .map(|min_hw_prio| {
+            let above_sw = max_sw_prio.map_or(0, |prio| prio.saturating_add(1));
+            (min_hw_prio - 1).max(above_sw)
+        });
```
Finomnis: Why is the hw prio important here? Can't it just be `above_sw`? The only situation where it is actually at `min_hw_prio` is if there is a gap between hw and sw tasks, and imo it's sane to assume that some none-rtic interrupts might fill that gap.

rtic/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ Example:
 
 ### Fixed
 
+- `RTIC_ASYNC_MAX_LOGICAL_PRIO` is above the highest software task when the device allows it, so monotonics can preempt a blocking software task (#1246).
```
Finomnis: Not correct as of now, as it is directly below the hardware task unless the software task pushes it up. Although I like the behavior more as described here and think we should implement it accordingly.

rtic-monotonics/src/lib.rs
```diff
@@ -48,10 +48,14 @@
...
-//! hardware tasks), or, if no hardware task is available, is set to the maximum priority in the
-//! system.
+//! The priority of the timer interrupt is given to `start` as a [`Priority`]. A task running at
+//! or above it must not use blocking delays, as time stops advancing while the interrupt cannot
+//! run.
+//!
+//! [`Priority::rtic_default`] is the priority RTIC reserves for async HAL drivers. It is 1 less
+//! than the lowest hardware task priority, but never lower than 1 above the highest software task
+//! priority, capped to the maximum priority. If there are no hardware tasks, it is the maximum
+//! priority in the system.
```
Finomnis: Reminder to adjust this to whatever behavior we decide in the other comments

